PR #52, opened 2020-10-22
Try installing rustup if needed in setup.sh [SAME VERSION]

bfjelds:


build/setup.sh
```diff
@@ -23,4 +15,36 @@ else
     apt update
     apt install -y $apt_dependencies
 fi
+if [ $? -ne 0 ]; then
```
kate-goldenring, line 18: should we keep quoting around _$?_ consistent? Sometimes you use `if [ $? -ne 0 ];` then and others `if [ "$?" -ne "0" ];`

kate-goldenring (approved):
Adding these explicit catches in the setup is helpful. Thanks!

jiria:
All for adding explicit catches, though wonder if something else goes south, we have the same problem. How about just disable set -e around the which invocations?

Issue #50, opened 2020-10-22
How to deploy akri to Google Compute Engine and DigitalOcean


DazWilkin:
Wrote a [post](https://pretired.dazwilkin.com/posts/201022/) summarizing how to deploy the akri end-to-end to both Google Compute Engine and DigitalOcean.

Hope it helps foster adoption of akri!

squillace:
hey! that's great -- any chance you could throw them into a PR? I love running things on DO.....

DazWilkin:
Sure!

Where (in the repo) would you like the files added?

I propose one document for each cloud then folks can add equivalent instructions for other platforms?

squillace:
I like it. @bfjelds @kate-goldenring what would you think?

bfjelds:
@DazWilkin, this is incredible!  i like the idea of docs/end-to-end-demo-*.md (something like: gce | digital-ocean) files for each environment.

DazWilkin:
Sounds good.

Will do it tomorrow.

PR #56, opened 2020-10-23
End-to-End Demo "all up" script [SAME VERSION]

DazWilkin:
Addresses: #50 

Bash script that combines the instructions for installing MicroK8s, akri, v4l2loopback module and the Demo to a cloud instance.

> **NOTE** Hopefully (!) shared by all cloud deployments (GCE, DO etc.) and so creating in its own branch.

> **NOTE** We'll need to ensure that this remains current with the existing [End-to-End Demo](./docs/end-to-end-demo.md) instructions


scripts/end_to_end_demo.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+sudo snap install microk8s --classic --channel=1.18/stable
+
+sudo usermod -a -G microk8s dazwilkin
```
bfjelds, line 5: maybe `dazwilkin` should be $USER?

scripts/end_to_end_demo.sh
```diff
@@ -0,0 +1,51 @@
...
+
+sudo microk8s status --wait-ready
+
+sudo microk8s enable dns helm3 rbac
+
+echo "--allow-privileged=true" >> /var/snap/microk8s/current/args/kube-apiserver
```
kate-goldenring, line 11: Maybe change to the following to avoid permissions error:
```
echo "--allow-privileged=true" | sudo tee -a /var/snap/microk8s/current/args/kube-apiserver
```

kate-goldenring:
I dont think we need a version change for this. Can you add [SAME VERSION] to the end of your PR title to pass this check?

scripts/end_to_end_demo.sh
```diff
@@ -0,0 +1,51 @@
...
+
+sudo gst-launch-1.0 -v videotestsrc pattern=smpte horizontal-speed=1 ! "video/x-raw,width=640,height=480,framerate=10/1" ! avenc_mjpeg ! v4l2sink device=/dev/video2 &
+
+sudo microk8s.helm3 repo add akri-helm-charts https://deislabs.github.io/akri/
+sudo microk8s.helm3 install akri akri-helm-charts/akri \
+  $AKRI_HELM_CRICTL_CONFIGURATION \
+  --set useLatestContainers=true \
+  --set udevVideo.enabled=true \
+  --set udevVideo.udevRules[0]='KERNEL==\"video[0-9]*\"'
+
```
bfjelds, line 49: we might create a wait between the helm install and the apply streaming app to let the brokers get up and running.

something like:

```
until kubectl get pods -l akri.sh/configuration=akri-udev-video -o jsonpath='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status}{"\n"}' | grep -v ContainersReady | grep "Ready=True" | wc -l | grep 2; do echo "waiting for k3s to become ready"; sleep 10; done
```

bfjelds: with maybe a better waiting message: `waiting for udev video brokers to become ready`

bfjelds: i think if we wait for the brokers to come up, then the instance and configuration services will be up and ready for the streaming app to start querying its pod env vars.

bfjelds:
this is incredible.  thank you so much for this contribution!!

scripts/end_to_end_demo.sh
kate-goldenring: I like the idea of a wait. When I tested it out, the streaming app errored and then started running when the brokers came up, so worst case it errors briefly. But, avoiding any erroring is definitely ideal.

scripts/end_to_end_demo.sh
kate-goldenring: For some context: When I ran it with the original line, the agent pods wouldn't be deployed because allow privileged was never set due to permissions errors.

kate-goldenring:
I think a good follow up to this PR would be to add an option in the normal End-to-End demo to just use this script as well and then jump to clean up. My only concern is consistency. Since theyll need to use sudo for all commands. Maybe we should leave sudo in for all commands instead of switching to sudo context.

scripts/end_to_end_demo.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
```
bfjelds, line 1: i think we should reference microk8s in the name of the script.  maybe end_to_end_microk8s_demo.sh

scripts/end_to_end_demo.sh
DazWilkin: Apologies! Yes, certainly not `dazwilkin`.

For DO this can be dropped as user interactions are as root too.

For GCP, it's non-trivial as the startup script runs as root but the user only be known when that user ssh's in.

DazWilkin: I think, because the subsequent commands all use `sudo microk8s.kubectl`, this line can be dropped.

Will redo both deployments to confirm the change, then update.

kate-goldenring:
> this is incredible. thank you so much for this contribution!!

+1 Thank you so much for adding this!

kate-goldenring:
Should [version check workflow](https://github.com/deislabs/akri/blob/170c790a4b96265b4defb9b6652875ca0b782d9c/.github/workflows/check-versioning.yml#L6) be updated to ignore this new scripts folder?

scripts/end_to_end_demo.sh
bfjelds: maybe 

> Should [version check workflow](https://github.com/deislabs/akri/blob/170c790a4b96265b4defb9b6652875ca0b782d9c/.github/workflows/check-versioning.yml#L6) be updated to ignore this new scripts folder?

yes, i think so.

bfjelds:
> Should [version check workflow](https://github.com/deislabs/akri/blob/170c790a4b96265b4defb9b6652875ca0b782d9c/.github/workflows/check-versioning.yml#L6) be updated to ignore this new scripts folder?

yes i think so

kate-goldenring:
Should [version check workflow](https://github.com/deislabs/akri/blob/170c790a4b96265b4defb9b6652875ca0b782d9c/.github/workflows/check-versioning.yml#L6) be updated to ignore this new scripts folder? @bfjelds what do you think?

scripts/end_to_end_demo.sh
bfjelds: can the startup script reference arguments for GCP?  it seems like you can almost, kinda, sort of do it: https://cloud.google.com/compute/docs/startupscript#custom

bfjelds: assuming $USER during SSH would be a known id tied to your GCP account?

scripts/end_to_end_demo.sh
DazWilkin: After I commit this change, I'll revise the GCE|DO docs to reflect it.

scripts/end_to_end_demo.sh
DazWilkin: Fixed: https://github.com/deislabs/akri/pull/56/commits/688c98a412a4078055ae42d57934b3f8bac5d742

scripts/end_to_end_demo.sh
DazWilkin: Fixed: https://github.com/deislabs/akri/pull/56/commits/7c04f32209c84f59b45fe35fec49a0adab6563d5

I hope you don't mind my using `kubectl wait ...` as this seems cleaner

scripts/end_to_end_demo.sh
DazWilkin: Fixed: https://github.com/deislabs/akri/pull/56/commits/a4eaae5aa06c920c8033c226bccdf739b81c0cc9

DazWilkin:
@kate-goldenring should I still add "[SAME VERSION]" to the PR title?

DazWilkin:
> @kate-goldenring should I still add "[SAME VERSION]" to the PR title?

I guess so....  "Check versioning ..."

DazWilkin:
Thank you both for the insightful feedback!

scripts/end_to_end_demo.sh
DazWilkin: Revised script filename:
+ Compute Engine PR #57 (Commit: https://github.com/deislabs/akri/pull/57/commits/c41ace92f688ac5c5c3d81ff8860efd7812f192f)
+ DigitalOcean PR #58 : (Commit: https://github.com/deislabs/akri/pull/58/commits/424f6e64779bacafd0b6bc87659413c45688f673)

bfjelds (approved):

kate-goldenring (approved):

kate-goldenring:
> @kate-goldenring should I still add "[SAME VERSION]" to the PR title?

Sorry for the delay. The checks use the pr name at the time of the latest commit, which is why the version check is still failing despite the title change. We can still go ahead and merge it. Thanks again for putting this together!

PR #73, opened 2020-10-27
Controller check capacity

### LINKED ISSUE #37: Controller not bringing down pending pods in case of `deviceUsage` slot collision
As expected, the kubelet will not schedule a pod if the pod requests an Instance's `deviceUsage` slot that has already been taken, and the pod will be left in Pending state. At this point, the controller should come and take down that pod (so it can possibly be rescheduled to a different slot if capacity has not been met). However, **this is not currently happening. The pod is staying in a Pending state.**

 
**Output of `kubectl get pods -o wide`**
```
akri-agent-daemonset-98p49                         1/1     Running   0          21h
akri-agent-daemonset-x9k9r                         1/1     Running   0          21h
akri-controller-deployment-75d655d869-nfmvz        1/1     Running   0          21h
worker1-akri-debug-echo-foo-8120fe-pod             0/1     Pending   0          21h
worker1-akri-debug-echo-foo-a19705-pod             0/1     Pending   0          21h
worker2-akri-debug-echo-foo-8120fe-pod             1/1     Running   0          21h
worker2-akri-debug-echo-foo-a19705-pod             1/1     Running   0          21h
```

**Kubernetes Version: [e.g. Native Kubernetes 1.19, MicroK8s 1.19, Minikube 1.19, K3s]**
MicroK8s 1.19

**To reproduce, run on cluster with 2 workers:**
```sh
helm install akri akri-helm-charts/akri --set debugEcho.enabled=true --set debugEcho.capacity=1 --set debugEcho.shared=true --set agent.allowDebugEcho=true --debug --set agent.host.dockerShimSock=/var/snap/microk8s/common/run/containerd.sock --set agent.host.crictl=/usr/local/bin/crictl --set debugEcho.capacity=1
```

**Expected behavior**
Controller should bring down the pending pods, and in this case where capacity is 1, pods should not be rescheduled

**Expected problem**
The controller has a grace period it allows to pass before bringing down pods. That grace period is being calculated from a timestamp that only occurs after a pod is assigned to a node, which never occurs if allocate fails. Consequently, we see the pod continuing to run in a Pending state as the Controller doesn't remove it if no start time was found: 
```
[2020-10-21T16:05:28Z TRACE controller::util::pod_action] time_choice_for_non_running_pods - no start time found ... give it more time? ("worker1-akri-debug-echo-foo-a19705-pod")
[2020-10-21T16:05:28Z TRACE controller::util::pod_action] time_choice_for_non_running_pods - give_it_more_time: (true)
```

kate-goldenring:
Currently, the controller creates a broker pod for every node listed in an Instance. This means if 5 nodes can see an instance but `capacity` is 2, the controller will schedule pods to all 5 nodes. Then, when the kubelet on each of the nodes tries to schedule the Pod with a call to `allocate`, only 2 of 5 will succeed. The other 3 nodes will be left in a pending state. 
Now, the controller checks how many nodes already have broker pods and compares that to capacity in order to decide how many other nodes should get brokers.

Test with DebugEcho Configuration by building the controller and agent containers and running:
```sh
    $AKRI_HELM_CRICTL_CONFIGURATION \
    --set imagePullSecrets[0].name="<your secret>" \
    --set agent.image.repository="ghcr.io/.../agent" \
    --set agent.image.tag="v0.0.37-amd64" \
    --set controller.image.repository="ghcr.io/...controller" \
    --set controller.image.tag="v0.0.37-amd64" \
    --set debugEcho.enabled=true \
    --set debugEcho.shared=true \
    --set debugEcho.capacity=1 \
    --set agent.allowDebugEcho=true 
```

closes #37 

**If applicable**:


controller/src/util/instance_action.rs
```diff
@@ -472,17 +474,41 @@ pub async fn handle_instance_change(
...
             }
         })
-        .collect::<Vec<String>>();
+        .sum();
+
+    // Assert that more nodes don't receive pods than specified by capacity
+    let num_nodes_to_receive_pod = capacity as i32 - num_nodes_already_with_pods;
+    trace!("handle_instance_change - capacity [{}] - number of nodes already with pods [{}] = [{}] node[s] to receive pods", capacity, num_nodes_already_with_pods, num_nodes_to_receive_pod);
+    let nodes_to_add: Vec<String> = if num_nodes_to_receive_pod < 0 {
+        error!(
```
jiria, line 493: Would this not happen on capacity change? Or would that trigger different handler and it would not come here?

kate-goldenring: The only way to change capacity currently is to modify the configuration. Currently, when the agent sees a configuration modification, it deletes the associated instance[s] and all device plugins and recreates it. All pods will be brought down by the controller (once the instance is deleted) and this code will be run through again for an Instance added event.

kate-goldenring:
@bfjelds Should this have triggered a Codecov report?

controller/src/util/instance_action.rs
```diff
@@ -472,17 +474,41 @@ pub async fn handle_instance_change(
...
-    let nodes_to_add = nodes_to_act_on
+    // Check how many nodes already have pods
+    let num_nodes_already_with_pods: i32 = nodes_to_act_on
         .iter()
-        .filter_map(|(node, context)| {
-            if ((context.action) == PodAction::Add) | ((context.action) == PodAction::RemoveAndAdd)
-            {
-                Some(node.to_string())
+        .filter_map(|(_, context)| {
+            if (context.action) == PodAction::NoAction {
```
bfjelds, line 481: i don't remember ... does NoAction equate to a pod is there?

bfjelds: are we trying to get the used capacity here?  or are we trying to get the number of brokers that have been deployed?

bfjelds: Instance.deviceUsage will include broker pods AND pods that people have tried to schedule.

If we query pods for the labels our controller adds, we will only get brokers.

kate-goldenring: Hmm thats an interesting point. So you're saying that this scenario could happen
1. Operator creates a Configuration with capacity of 5 in a two worker node cluster
2. Controller deploys 2 brokers -> Agents reserve 2 of the 5 `deviceUsage` slots
3. Operator deploys 2 pods requesting that Instance as a resource limit
4. Agent reserves 2 slots for those pods
Now 4 of 5 slots have been taken but the Akri Controller only deployed 2 brokers. Is this the behavior we want? It seems that the controller should check how many pods are using the instance, whether Akri deployed broker or not. 

But to answer your question: I was using the number of nodes with `PodAction:NoAction` as the number of brokers already deployed because [when determining an action for a pod](https://github.com/deislabs/akri/blob/c80e73d049fde9b57c8aa8308a248e7afe7002c2/controller/src/util/instance_action.rs#L455), running ones keep this status. Pods that should be removed also get the [NoAction](https://github.com/deislabs/akri/blob/c80e73d049fde9b57c8aa8308a248e7afe7002c2/controller/src/util/instance_action.rs#L412) label which may not be modified if there is an early return from [`determine_action_for_pod`](https://github.com/deislabs/akri/blob/c80e73d049fde9b57c8aa8308a248e7afe7002c2/controller/src/util/instance_action.rs#L163). 

TLDR: this need to change and like you said doesnt take into account the other pods that have been scheduled to a resource that arent brokers. So should we use the number of taken `deviceUsage` slots as the capacity?

bfjelds: That is the scenario I was thinking about.  I do think that Instance.deciceUsage should be the source of truth.

Part of me thinks that the controller shouldn't worry about capacity...just create a pod for each node in Instance.nodes and let the scheduler deal with the rest.   As a result, there will be some Pending pods, but that just ensures that we continually have HA.

But I'm not sure I have thought that completely through.

bfjelds: Alternatively, we could remove the node affinity and just create Instance.caoacity pods....what is scheduled and what is pending will depend on allocate (which will control operator deployed pods as well)

bfjelds: The second s probably more efficient if there are lots and lots of nodes.

And probably better for us to get out of the node choosing game.

bfjelds: But is a bigger change in behavior.

kate-goldenring: In this case, would the controller even need to monitor instance changes? Would the scheduler also be in charge of bringing down pods that are `Pending`. I agree that it makes sense to move away from the controller choosing the node.

kate-goldenring: I wonder if we should have controller check capacity (based on deviceUsage used slots) for now, especially if we want to set default capacity to 1, since there is a higher chance of the pending pods case. Or we can make the controller better at bringing down pending pods.

bfjelds: i think we should stop worrying about Pending.

i like this for strategies:

* **CapacityDaemonSet** - this is what we have now.  if capacity=5 and there are 10 nodes that can see an instance, we create 10 pods with node-affinity, 5 end up Pending.
* **CapacityDeployment** - this would be new (and should be the eventual default) ... if capacity=5 and there are 10 nodes that can see an instance, a deployment with 5 replicas is created using resources.limit and PLACEHOLDER for targeting (no node-affinity, k8s scheduler owns scheduling)

kate-goldenring:
The pending pods allow for HA when a node goes down. The controller should continue to have a "CapacityDaemonSet" deployment strategy of creating a broker pod for every node, even if some end up pending. This PR would have ended that strategy.

PR #79, opened 2020-10-27
Enable controller running on worker node

### LINKED ISSUE #67: Allow running Akri controller on non-dedicated control plane nodes
Akri Helm chart allows to specify whether to run the controller `onlyOnMaster`. But for that, Akri expects a specific label to be present on the control plane nodes. Not all K8s distributions have such tag. Also, with AKS, you cannot deploy onto control plane nodes. 

Akri should be able to run on clusters without the specific label that Akri is looking for and also on clusters where users cannot deploy onto CP nodes.

Consider moving the toleration outside of the `onlyOnMaster` condition in the Helm chart. This way, Akri could run on a control plane, but not only. With `onlyOnMaster`, it would only run on control plane.

This would not allow disabling the option to run on CP, but I think that can be solved later based on feedback.

kate-goldenring:
In the Akri Helm chart, you can specify an `onlyOnMaster` value. By default, it is set to `true`, requiring a specific label to be present on the control plane nodes. Not all K8s distributions have such tag. Also, on some cloud Kubernetes providers like AKS, you cannot deploy onto control plane nodes.

This PR moves the toleration to allow the controller to run on a node with the Kubernetes default "master" label outside of the `onlyOnMaster` condition in the Helm chart. This way, Akri could run on a control plane but it isn't a requirement. With `onlyOnMaster` set to true, it will only run on control plane.

closes #67 

**If applicable**:


bfjelds (approved):

docs/user-guide.md
```diff
@@ -46,7 +46,7 @@ To get started using Akri, you must first decide what you want to discover and w
         ```
     1. If using **Kubernetes**, Helm and crictl do not require additional configuration.
 
-1. Install Akri Helm chart and enable the desired Configuration (in this case, ONVIF is enabled). See the [ONVIF Configuration documentation](./onvif-sample.md) to learn how to customize the Configuration. Instructions on deploying the udev Configuration can be found in [this document](./udev-sample.md).
+1. Install Akri Helm chart and enable the desired Configuration (in this case, ONVIF is enabled). See the [ONVIF Configuration documentation](./onvif-sample.md) to learn how to customize the Configuration. Instructions on deploying the udev Configuration can be found in [this document](./udev-sample.md). To allow the controller to run on a node other than the control plane, add `--set controller.onlyOnMaster=false` to your install command below.
```
jiria, line 49: Should onlyOnMaster be false by default?

bfjelds: is there a reason to ever have it be only on master?  i don't know why i set it at first ... if we never had, would we adding a way to enforce that now?

jiria: You could argue that controllers might make sense to be collocated and closer to etcd/apiservers. Re adding the enforcement now, it probably would not be on my high priority list.

jiria: Also, if you have dedicated control plane nodes, it makes sense to have controller there instead of using resources and be possibly exposed to bad players on worker nodes.

kate-goldenring: seems like consensus is to leave it to true by default? I see both sides here

jiria: I believe both me and Brian are advocating for switching to false. But Brian, please correct me if I misrepresented you.

bfjelds: i agree with it being false.  i'm not clear, at this point, why it should ever be true.

deployment/helm/values.yaml
```diff
@@ -31,8 +31,9 @@ controller:
     tag:
     # pullPolicy is the Akri Controller pull policy
     pullPolicy: Always
-  # onlyOnMaster dictates whether the Akri Controller will only run on a master node
-  onlyOnMaster: true
+  # onlyOnMaster dictates whether the Akri Controller will only run on node with 
+  # the master node label 'node-role.kubernetes.io/master'
```
jiria, line 35: Should we clarify here the value of the label too, since the actual filter is not just on the label key?

docs/user-guide.md
```diff
@@ -44,9 +40,17 @@ To get started using Akri, you must first decide what you want to discover and w
...
+1. Install Akri Helm chart and enable the desired Configuration (in this case, ONVIF is enabled). See the [ONVIF Configuration documentation](./onvif-sample.md) to learn how to customize the Configuration. Instructions on deploying the udev Configuration can be found in [this document](./udev-sample.md). 
+
+    To ensure that the Akri Controller only runs on the control plane, add `--set controller.onlyOnMaster=true` to your install command below. This will only run the controller on a node with the label `node-role.kubernetes.io/master`, which is the default label for the control plane node for Kubernetes. However, MicroK8s and K3s do not have this label by defaul, so you can add it by running `kubectl label node ${HOSTNAME,,} node-role.kubernetes.io/master= --overwrite=true`.
```
jiria, line 52: would say "on the control plane nodes"

jiria: Also same comment about the label value.

jiria, line 52: nit: default (spelling typo)

jiria (approved):

deployment/helm/values.yaml
kate-goldenring: Can you clarify "since the actual filter is not just on the label key"? Isnt this the case? OnlyOnMaster is specifying that the controller will only run on a node with that label, so it seems to be a filter on that label key. Are you suggesting we add more commenting here about what the label means?

bfjelds: I think removing the node selector means it can run anywhere...maybe the master, maybe not?

jiria: My point was that the filter is this:
```
nodeSelector:
        node-role.kubernetes.io/master: ""
```

So it is not just label key, it is both key and value (empty string). As such, my argument was that the document should fully state what is the nodeSelector (key+value). Or do you not agree?

deployment/helm/templates/controller.yaml
```diff
@@ -40,12 +40,12 @@ spec:
       nodeSelector:
         "kubernetes.io/os": linux
       {{- end }}
-      {{- if .Values.controller.onlyOnMaster }}
-      # Only run this pod on the master.
       tolerations:
```
bfjelds, line 43: should we add another flag controller.allowOnMaster (default to true) that toggles the tolerations?

deployment/helm/values.yaml
bfjelds: we could create a more generalized approach to onlyOnMaster.  something like we do for imagePullSecrets.  that would allow someone to specify whatever they wanted (and it could default to nothing):

```
# nodeSelector is the array of NodeSelectors that help target nodes with specific labels.
# This can be set from the helm command line using `--set nodeSelectors[0].name-role.kubernetes.io/master=""`
nodeSelectors: []
```

bfjelds: one thing i just noticed is that it looks like nodeSelector is being specified twice in controller.yaml (once for onlyLinux and once for onlyOnMaster).  i just tested it and that doesn't work ... only the last thing set ends up in the kubernetes object.

deployment/helm/templates/controller.yaml
kate-goldenring: Why wouldn't someone want the controller on master? I see it being another setting for customizing but could also see reason to wait until someone requests it

bfjelds: maybe they want the master to be dedicated to Kubernetes?  maybe their master is more locked down somehow?

deployment/helm/values.yaml
bfjelds: to fix the duplicate nodeSelector issue (and add more configurations for someone), we could adjust:

values.yaml:
```yaml
  # onlyOnMaster dictates whether the Akri Controller will only run on a node labeled with `node-role.kubernetes.io/master: ""`
  onlyOnMaster: false
  # linuxOnly dictates whether the Akri Controller will only run on a linux node
  linuxOnly: true
  # nodeSelectors is the array of nodeSelectors used to target nodes for the Akri Controller to run on
  # This can be set from the helm command line using `--set nodeSelectors.label="value"`
  nodeSelectors: {}
```
controller.yaml (remove both existing nodeSelector lines and replace with) :
```yaml
      {{- if or .Values.controller.linuxOnly .Values.controller.onlyOnMaster }}
      nodeSelector:
        {{- if .Values.controller.nodeSelectors }}
          {{- toYaml .Values.controller.nodeSelectors | nindent 8 }}
        {{- end }}
        {{- if .Values.controller.linuxOnly }}
        "kubernetes.io/os": linux
        {{- end }}
        {{- if .Values.controller.onlyOnMaster }}
        node-role.kubernetes.io/master: ""
        {{- end }}
      {{- else }}
        {{- if .Values.controller.nodeSelectors }}
      nodeSelector:
          {{- toYaml .Values.controller.nodeSelectors | nindent 8 }}
        {{- end }}
      {{- end }}
```

then these commands would all work:
```sh
# keeps linux only, sets only on master, lets user add any nodeSelector they want
helm template ./deployment/helm --set controller.linuxOnly=true,controller.nodeSelectors.foo=bard,controller.onlyOnMaster=true
# defaults to only on linux and whatever the default is for onlyOnMaster
helm template ./deployment/helm
```

deployment/helm/templates/controller.yaml
kate-goldenring: That makes sense.

deployment/helm/values.yaml
```diff
@@ -31,10 +31,17 @@ controller:
...
+  # onlyOnControlPlane dictates whether the Akri Controller will only run on nodes with 
+  # the label with (key, value) of ("node-role.kubernetes.io/master", "")
+  onlyOnControlPlane: false
+  # allowOnControlPlane dictates whether a toleration will be added to allow to Akri Controller 
+  # to run on the control plane node
+  allowOnControlPlane: true
   # linuxOnly dictates whether the Akri Controller will only run on a linux node
   linuxOnly: true
+  # nodeSelectors is the array of nodeSelectors used to target nodes for the Akri Controller to run on
+  # This can be set from the helm command line using `--set nodeSelectors.label="value"`
```
bfjelds, line 43: `--set nodeSelectors` should be `--set controller.nodeSelectors`

deployment/helm/templates/controller.yaml
```diff
@@ -36,17 +36,27 @@ spec:
...
-      {{- end }}
-      {{- if .Values.controller.onlyOnMaster }}
-      # Only run this pod on the master.
+      {{- if .Values.controller.allowOnControlPlane }}
       tolerations:
         # Allow this pod to run on the master.
         - key: node-role.kubernetes.io/master
           effect: NoSchedule
+      {{- end }}
+      {{- if or .Values.controller.linuxOnly .Values.controller.onlyOnControlPlane }}
```
jiria, line 45: Should this if not include .Values.controller.nodeSelectors?

jiria: Or should nodeSelectors not be here given that they are also below?

deployment/helm/templates/controller.yaml
```diff
@@ -36,17 +36,27 @@ spec:
...
+        {{- end }}
+        {{- if .Values.controller.linuxOnly }}
+        "kubernetes.io/os": linux
+        {{- end }}
+        {{- if .Values.controller.onlyOnControlPlane }}
         node-role.kubernetes.io/master: ""
+        {{- end }}
+      {{- else }}
+        {{- if .Values.controller.nodeSelectors }}
+      nodeSelector:
```
jiria, line 58: Can the nodeSelector be specified multiple times?

deployment/helm/templates/controller.yaml
bfjelds: there are 2 clauses here.  one for when either linuxOnly or onlyOnControlPlane ... and one when neither of those.

in both clauses, nodeSelectors is added if it is set.

this allows us to only include the yaml `nodeSelector` iff there is something to add to nodeSelector.

deployment/helm/templates/controller.yaml
bfjelds: it is not.  that is what this change is all about fixing.

bfjelds: if is added once for `{{- if or .Values.controller.linuxOnly .Values.controller.onlyOnControlPlane }}`

and once for `{{- else }}{{- if .Values.controller.nodeSelectors }}`

evrardjp:
Curious: Is it always necessary to bump the version inside a PR?

deployment/helm/templates/controller.yaml
bfjelds: it can be simplified (the `or` can accept 3 args):

```yaml
      {{- if or .Values.controller.linuxOnly .Values.controller.onlyOnMaster .Values.controller.nodeSelectors }}
      nodeSelector:
        {{- if .Values.controller.nodeSelectors }}
          {{- toYaml .Values.controller.nodeSelectors | nindent 8 }}
        {{- end }}
        {{- if .Values.controller.linuxOnly }}
        "kubernetes.io/os": linux
        {{- end }}
        {{- if .Values.controller.onlyOnMaster }}
        node-role.kubernetes.io/master: ""
        {{- end }}
      {{- end }}
```

deployment/helm/templates/controller.yaml
bfjelds: a simpler form would be:

```yaml
      {{- if or .Values.controller.linuxOnly .Values.controller.onlyOnMaster .Values.controller.nodeSelectors }}
      nodeSelector:
        {{- if .Values.controller.nodeSelectors }}
          {{- toYaml .Values.controller.nodeSelectors | nindent 8 }}
        {{- end }}
        {{- if .Values.controller.linuxOnly }}
        "kubernetes.io/os": linux
        {{- end }}
        {{- if .Values.controller.onlyOnMaster }}
        node-role.kubernetes.io/master: ""
        {{- end }}
      {{- end }}
```

bfjelds (approved):

deployment/helm/templates/controller.yaml
jiria: Thanks, missed the else part. But agreed, we should use this simplified version.

docs/user-guide.md
```diff
@@ -59,9 +61,26 @@ helm install akri akri-helm-charts/akri
...
+   Configuration documentation](./onvif-sample.md) to learn how to customize the Configuration. Instructions on
+   deploying the udev Configuration can be found in [this document](./udev-sample.md). 
+
+    By default the Controller can be deployed to any control plane or worker node. These settings can be changed by
+    adding extra settings when installing Akri below. If you don't want the Controller to ever be scheduled to control
+    plane nodes, add `--set controller.allowOnControlPlane=false` to your install command below. Conversely, if you only
+    want the Controller to run on control plane nodes, add `--set controller.onlyOnControlPlane=true`. This will
+    guarantee the Controller only runs on nodes with the label (key, value) of (`node-role.kubernetes.io/master`, ""),
+    which is the default label for the control plane node for Kubernetes. However, control plane nodes on MicroK8s and
+    K3s do not have this label by default, so you can add it by running `kubectl label node ${HOSTNAME,,}
```
jiria, line 82: Should we mention here that they can also use the nodeSelectors value to add any existing labels (as K3s has one)?

PR #83, opened 2020-10-28
slack badge

kate-goldenring:
No where on our GitHub do we point to a channel for questions and discussion. This adds a badge that will lead to our slack channel.

**If applicable**:


bfjelds:
the badges section is getting pretty full.

i wonder if we shouldn't just limit our main README to just:

LINE1: slack, kubernetes version, codecov
LINE2: tests on K8s, k3s, mk8s

jiria:
I am split here, but wonder if internal dev labels could go onto some dev sub doc. So basically what Brian said on the front page, the rest offload on page with instructions to rebuilding/extend.

kate-goldenring:
I was worried it was too crowded at first, too, but then I looked at [Thanos](https://github.com/thanos-io/thanos/blob/master/README.md) and figured we are fine. I worry about moving some of the badges to development.md because it'll probably take us longer to notice when they are failing, but then again its the e2e tests that are most likely to fail since we dont run them during PRs. Also, development.md seems like a weird home for them. We could also add a "Checks" section to the README at the bottom

jiria:
My feeling was also that I have seen much more on other projects, hence split.

bfjelds (approved):

bfjelds:
@evrardjp makes a great point about badges that need to be manually updated.  

at least for Kubernetes version, i think the manual maintenance side of it is balanced by the "in your face" highlighting that the badge provides (versus just text ... even **emphasized** text).

i say we roll with what you've created and we can adjust if we get feedback.

jiria (approved):

Issue #69, opened 2020-10-26
How do I create a configuration without a broker?


jiria:
How can I create a configuration that will detect leaf devices and advertise them as extended resources, but not deploy any brokers automatically?

kate-goldenring:
Our documentation was not clear on how to do this and we were using Akri's sample brokers as defaults. #87 fixed this. Now, broker specs will not be added to the generic protocol Configurations unless a broker image is specified. This is explained in our [user guide](https://github.com/deislabs/akri/blob/main/docs/user-guide.md) and elaborated upon in our [documentation](https://github.com/deislabs/akri/blob/main/docs/requesting-akri-resources.md) on scheduling pods to resources advertised by Akri.

Issue #33, opened 2020-10-20
Implementing On-Device Audio


cwoolum:
I am currently running a k3s cluster on some Raspberry Pi 4s. Would it be possible for Akri to utilize an audio card on one of these Pi's as an Akri Plugin?

I would assume a device plugin would need to be created. Does this make sense as an Akri use-case?

jiria:
Hello Chris,

If the card gets exposed through udev, you can certainly expose it to pods using Akri, no code change needed. This use case should be covered by Akri as is.

Let us know!

kate-goldenring:
Hi @cwoolum! As @jiria mentioned, our udev protocol supports discovering anything on the Linux device file system. You simply have to specify a [udev rule](https://wiki.archlinux.org/index.php/Udev) that can be used to locate the device[s] in your Configuration. As to making your Akri Configuration, you can either modify our [sample udevVideo Configuration](https://github.com/deislabs/akri/blob/main/docs/udev-sample.md#modifying-the-udev-rule) or [create your own](https://github.com/deislabs/akri/blob/main/docs/modifying-akri-installation.md#generating-modifying-and-applying-a-configuration-yaml). I'll walk you through both. 

We will create a Configuration that will discover all devices in the sound subsytem and deploy an nginx broker pod. Your udev rule will probably be more specific. This rule will discover all devices listed when you run `udevadm trigger --verbose --dry-run --type=devices --subsystem-match=sound`
1. Option 1: Modifying the udevVideo Configuration
Simply run the following on your cluster. Be sure to initialize your cluster for Akri. Steps for this can be found at the beginning of the [rpi4 demo](https://github.com/deislabs/akri/blob/main/docs/rpi4-demo.md).
```
helm repo add akri-helm-charts https://deislabs.github.io/akri/
helm install akri akri-helm-charts/akri \
    --set useLatestContainers=true \
    --set udevVideo.enabled=true \
    --set udevVideo.udevRules[0]='SUBSYSTEM=="sound"' \
    --set udevVideo.brokerPod.image.repository='nginx'
```
You will see nginx brokers deployed to every discovered sound device.

2. Option 2: Create a Configuration
It's easier to work off an existing Configuration, so download the udevVideo one by running:
```
helm template akri akri-helm-charts/akri \
    --set useLatestContainers=true \
    --set udevVideo.enabled=true \
    --set controller.enabled=false \
    --set agent.enabled=false > udevAudio.yaml
```
[...]

PR #87, opened 2020-10-29
Generic udev Configuration

### LINKED ISSUE #68: Add documentation on how to use udev outside of udevvideo
Currently, two use cases for Akri are spelled out in the docs. onvifvideo and udevvideo. But udev can be used for other devices as well.

Please add a top level entry on how to use udev for non-video related leaf devices. Also consider making that a top level and reference the video use case from that as an example. Added bonus would be a template for Helm, but not necessary, if a sample Akri Configuration is provided and explained how it should be used.

kate-goldenring:
We currently only have documentation and a sample Configuration for how to use the udev discovery handler to discover usb cameras. This adds a generic sample udev Configuration and documentation on how to use it to find any device discoverable by udev. It also modifies the name of the ONVIF configuration from onvif-video to onvif. This pr creates a standard of providing a generic configuration for each protocol that can be applied to a cluster with helm via `--set <protocol>-enabled=true`. Additionally, it sets the standard that a broker image must be specified on installation, else Akri will advertise the resources but not deploy brokers. Finally, it creates a standard of creating a `<protocol>-configuration.md` document for each generic configuration provided.

We've received questions on how to use udev to discover audio devices (ref #33). Modifying the udevVideo configuration seems illogical and starting from an empty yaml is a big barrier to entry. 
closes #68 
ref #69 

**If applicable**:


deployment/helm/values.yaml
```diff
@@ -164,6 +164,45 @@ onvifVideo:
...
+  # udev configuration
+  name: akri-udev
+  # properties is a map of properties that will be passed to any instances
+  # created as a result of applying this udev configuration
+  properties:
+  # udevRules is the list of udev rules used to find instances created as a result of
+  # applying this udev configuration
+  udevRules:
+  # capacity is the capacity for any instances created as a result of
+  # applying this udevVideo configuration
```
kate-goldenring, line 180: just "udev"

kate-goldenring:
I think it would make sense at this point to remove the udevVideo configuration. Instead in the [udev video sample documentation](https://github.com/deislabs/akri/blob/main/docs/udev-sample.md#usage), we can walk through the steps of setting the environment variables (udev values would need to modified to be able to configure this), broker image, and Configuration name. Thoughts?

deployment/helm/templates/udev.yaml
```diff
@@ -0,0 +1,52 @@
...
+    imagePullSecrets:
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
+  {{- end }}
+  {{- if .Values.udev.brokerPod.image.repository }}
+  # only add service specs if a broker image was specified
+  instanceServiceSpec:
+    type: {{ .Values.udev.instanceService.type }}
+    ports:
+    - name: grpc
```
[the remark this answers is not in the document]
jiria: Also wonder if for more complex scenarios, we should provide a sample config in the doc below to allow users to customize whether to have service specs or what the service specs should look like. Same for broker spec. Users might want to customize even the broker spec more than what we allow.

docs/udev-configuration.md
```diff
@@ -0,0 +1,148 @@
+# Using the udev Discovery Protocol in a Configuration
```
jiria, line 1: Should this doc be pointed to from the main README as part of the supported protocols?

docs/udev-configuration.md
```diff
@@ -0,0 +1,148 @@
...
+All the instances will be named in the format `<configuration-name>-<hash>`. You could change the name of the
+Configuration and resultant Instances to be `sound-device` by adding `--set udev.name=sound-devices` to your
+installation command. Now, you can schedule pods that request these Instances as resources. Assuming the Configuration
+name has been set to `sound-devices`, you can request the `sound-device-ha5h00` Instance as a resource by adding the
+following to the PodSpec of your Deployment or Job:
+```yaml
+  resources:
+    limits:
+      akri.sh/sound-device-ha5h00: "1"
+    requests:
```
jiria, line 75: Love this doc. Though wonder (perhaps after this PR), if we should have this documented generically for all protocols.

kate-goldenring:
> I think it would make sense at this point to remove the udevVideo configuration. Instead in the [udev video sample documentation](https://github.com/deislabs/akri/blob/main/docs/udev-sample.md#usage), we can walk through the steps of setting the environment variables (udev values would need to modified to be able to configure this), broker image, and Configuration name. Thoughts?

I wonder if we should phase out the udevVideo configuration, since it is part of our e2e demo. Some people may be currently running it. I could change the e2e to use just udev and then maybe we remove udevVideo entirely for our next release. That bodes the question: do be make onvifVideo onvif?

deployment/helm/templates/udev.yaml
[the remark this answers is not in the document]
jiria: Aaa, thanks. Perhaps just a link to that then. Is it possible for folks to drop instanceServiceSpec and  configurationServiceSpec if desired?

docs/udev-configuration.md
jiria: The user-guide looks to me as ONVIF specific, with a link to udev (and agreed that generic udev should be preferred, perhaps with a reference to how to adjust to udevvideo, as you mentioned earlier). Should we perhaps modify the user-guide to have the common start and then have a split per protocol, where we can add more as we support more?

docs/udev-configuration.md
jiria: Agreed on the generic and sample, onvif is an exception I think.

Here I was pointing out the fact that you are documenting how to directly take advantage of the generated extended resources (that folks could leverage directly through their pod specs if they want, not through configuration broker specs), rather than just depending on automatic broker deployment. My point was that this would be applicable for other protocols as well, and as such, should be in a protocol agnostic document. Please tell me if you feel otherwise.

kate-goldenring: oh definitely! I can move this out to a different doc

deployment/helm/templates/udev.yaml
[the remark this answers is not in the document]
kate-goldenring: Not currently. I can add that option.

deployment/helm/values.yaml
```diff
@@ -164,6 +164,45 @@ onvifVideo:
...
+  enabled: false
+  # name is the Kubernetes resource name that will be created for this
+  # udev configuration
+  name: akri-udev
+  # properties is a map of properties that will be passed to any instances
+  # created as a result of applying this udev configuration
+  properties:
+  # udevRules is the list of udev rules used to find instances created as a result of
+  # applying this udev configuration
+  udevRules:
```
bfjelds, line 178: not sure if it is needed, but I've initialized with empty array in other places: `udevRules: []`

docs/udev-configuration.md
```diff
@@ -0,0 +1,148 @@
...
+Instead of having to assemble your own udev Configuration yaml, we have provided a [udev Helm
+template](../deployment/helm/templates/udev.yaml). To add the udev Configuration to your cluster, simple set
+`udev.enabled=true`. Be sure to also **specify one or more udev rules** for the Configuration. Helm will automatically
+apply the udev Configuration yaml for you, and the Akri Agent will advertize discovered leaf devices as resources. By
+default, the udev Configuration does not specify a broker pod or services, so upon discovery, broker pods will not be
+deployed nor will services be created. Later, we will discuss [how to add a custom broker to the
+Configuration](./#adding-a-custom-broker-to-the-configuration).
+```bash
+helm repo add akri-helm-charts https://deislabs.github.io/akri/
+helm install akri akri-helm-charts/akri \
```
bfjelds, line 50: use akri-dev:  `helm install akri akri-helm-charts/akri-dev`

deployment/helm/templates/udev.yaml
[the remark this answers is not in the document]
bfjelds: i like the idea of parameterizing the name ... i didn't for onvifVideo because our sample app was depending on the name.  but for this, i think it would be great .. .the default could be 'grpc`

deployment/helm/values.yaml
kate-goldenring: I wanted it to throw an error if empty so it would prompt people to add a rule when trying to install helm, but I can see why setting an empty array would be good.

bfjelds: I like your idea more than initializing with []

deployment/helm/templates/onvif.yaml
```diff
@@ -0,0 +1,93 @@
...
+      # #   brokerPod:
+      # #     env:
+      # #       ENV_KEY1: VALUE_1
+      env:
+        {{- range $key, $val := .Values.onvif.brokerPod.env }}
+        - name: {{ $key }}
+          value: {{ $val | quote }}
+        {{- end }}
+      {{- end }}
+      resources:
```
jiria, line 56: Its fine if this is out of scope, but will ask. Should we support specification of other requires/limits? Possibly this could be relegated to folks using custom configuration, instead of the precreated template.

docs/onvif-configuration.md
```diff
@@ -1,33 +1,34 @@
...
 
 ## Usage
-To use the default ONVIF Configuration in your Akri-enabled cluster, you can simply set `onvifVideo.enabled=true` when installing the Akri Helm chart.  More information about the Akri Helm charts can be found in the [user guide](./user-guide.md#understanding-akri-helm-charts).
+To use the default ONVIF Configuration in your Akri-enabled cluster, you simply set `onvif.enabled=true` when installing the Akri Helm chart. If you would like broker pods to be deployed automatically to discovered cameras, set `udev.brokerPod.image.repository` to point to your broker image. Alternatively, if it meets your scenario, you could use the Akri frame server broker as done below. If you would rather manual deploy pods to utilize the cameras advertized by Akri, don't specify a broker pod and see our documentation on [requesting resources advertized by Akri](./requesting-akri-resources.md). More information about the Akri Helm charts can be found in the [user guide](./user-guide.md#understanding-akri-helm-charts).
```
jiria, line 14: manual -> manually

docs/requesting-akri-resources.md
```diff
@@ -0,0 +1,59 @@
...
+helm repo add akri-helm-charts https://deislabs.github.io/akri/
+helm install akri akri-helm-charts/akri-dev \
+    --set useLatestContainers=true \
+    --set protocolA.enabled=true
+```
+After installing Akri and your Configuration, list all discovered instances by running `kubectl get akrii`. Note `akrii` is a short name for Akri Instance.
+All the instances will be named in the format `<configuration-name>-<hash>`. You could change the name of the
+Configuration and resultant Instances to be `protocolA-device` by adding `--set protocolA.name=protocolA-device` to your
+installation command. Now, you can schedule pods that request these Instances as resources. Assuming the Configuration
+name has been set to `protocolA-device`, you can request the `protocolA-device-ha5h00` Instance as a resource by adding the
```
jiria, line 15: The Instance name is dependent on more than just the Configuration name, right? Should we say: you can request for example `...` Instance as a resource to show that you should not expect ha5h00 to exist?

jiria (approved):

deployment/helm/templates/onvif.yaml
kate-goldenring: We have [documentation on requesting other limits](https://github.com/deislabs/akri/blob/main/docs/modifying-akri-installation.md#modifying-the-brokerpodspec). I think if we find people creating issues for this/ asking for it to be configurable via helm values, we could add it.

docs/requesting-akri-resources.md
kate-goldenring: The instance name is always <configuration-name>-<id> where id is hash(property defining device) for shared and hash(property to define device + node name) for unshared. an example "property to define device" for udev would be devnode. IE: /dev/video0. Sounds like I should explain where ha5h00 came from. How about this:
All the instances will be named in the format `<configuration-name>-<id>`, where `<id>` is a hash of a descriptor of the device and the node that can see the device in the case where the device is not sharable, such as a usb camera. For example, the id of a usb camera at devnode `/dev/video0` on a node named workerA would be hash(/dev/video0workerA). On the other hand, since IP cameras are sharable, the `id` for an IP camera would be hash(rstp url).

jiria: Not sure if the detail is needed, but I think it would be cool to add it. +1

deployment/helm/templates/onvif.yaml
jiria: Sounds good.

docs/modifying-akri-installation.md
```diff
@@ -45,11 +48,12 @@ command could be run:
 ```bash 
 helm upgrade akri akri-helm-charts/akri-dev \
     --set useLatestContainers=true \
-    --set onvifVideo.enabled=true \
-    --set onvifVideo.ipAddresses.action=Exclude \
-    --set onvifVideo.ipAddresses.items[0]=10.0.0.1
+    --set onvif.enabled=true \
+    --set onvif.brokerPod.image.repository="ghcr.io/../onvif-broker" \
```
bfjelds, line 52: why ghcr.io/../onvif-broker?  should '..' be explained somewhere?

docs/onvif-configuration.md
```diff
@@ -77,10 +81,14 @@ pod, you can update the Configuration like this:
...
     --set useLatestContainers=true \
-    --set onvifVideo.enabled=true \
-    --set onvifVideo.capacity=2
+    --set onvif.enabled=true \
+    --set onvif.brokerPod.image.repository="ghcr.io/deislabs/akri/onvif-video-broker:latest-dev" \
+    --set onvif.capacity=2
 ```
 
+## Disabling automatic service creation
+By default, the generic ONVIF Configuration will create services for (all the brokers of) a specific Akri Instance and (all the brokers of) an Akri Configuration. Disable the create of Instance level services and Configuration level services by setting `--set onvif.createInstanceServices=false` and `--set onvif.createConfigurationService=false`, respectively.
```
bfjelds, line 90: why is 'all of the brokers of' parenthetical?

docs/modifying-akri-installation.md
kate-goldenring: Good point. Do you think it is better to include the image -- maybe just adding \<your broker image> -- and explain how a broker image is needed for broker pods to be deployed or to not include it at all? and just focus on discovery?

docs/onvif-configuration.md
kate-goldenring: I guess I wanted to emphasize "Instance" level service vs "Configuration" level services. Will happily remove the parentheses.

bfjelds (approved):

docs/modifying-akri-installation.md
kate-goldenring: Made some changes to explain the broker images

PR #92, opened 2020-11-03
Update end-to-end-demo.md

didier-durand:
Hi there,

/var/snap/etc.. no longer exists in recent version : it is now /snap/ect...   Also args/kube-apiserver not present on Ubuntu 20.04 : is default-args/kube-apiserver the proper replacement?

Didier

### DIFF
--- a/docs/end-to-end-demo.md
+++ b/docs/end-to-end-demo.md
@@ -61,7 +61,7 @@ carry out one or the other, then continue on with the rest of the steps.
     ```
 1. Enable privileged pods and restart microk8s.
     ```sh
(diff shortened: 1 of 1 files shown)


bfjelds:
it seems it should align with the MicroK8s docs (https://microk8s.io/docs/configuring-services): `The configuration of these services is read from files stored in the $SNAP_DATA directory, which normally points to /var/snap/microk8s/current ... The apiserver daemon is started using the arguments in ${SNAP_DATA}/args/kube-apiserver`

@didier-durand , is SNAP_DATA defined in you env?

didier-durand:
@bfjelds : my bad on this one. I messed up on the paths. Sorry for the noise! Closing PR. Didier

Issue #114, opened 2020-11-08
Akri udev ATTRS ignored


jiria:
**Describe the bug**
Based on the Akri [udev grammar](https://github.com/deislabs/akri/blob/main/agent/src/protocols/udev/udev_rule_grammar.pest) it seems that ATTRS might be supported, but when I pass ATTRS as part of my udev rule, Agent is ignoring it:
```
[2020-11-08T09:04:14Z TRACE agent::protocols::udev::discovery_impl] parse_udev_rule - unsupported field ATTRS{manufacturer}
```
Is v2 not yet supported?

**Output of `kubectl get pods,akrii,akric -o wide`**
agent and controller are running, my configuration is specified, but too many instances are found, because ATTRS of the rule are ignored

**Kubernetes Version: K3s**

**To Reproduce**
Supplying rule such as `SUBSYSTEM=="tty"\, ATTRS{manufacturer}=="Silicon Labs"\, ATTRS{idProduct}=="ea60"'` matches all tty devices.

**Expected behavior**
Supplying rule such as `SUBSYSTEM=="tty"\, ATTRS{manufacturer}=="Silicon Labs"\, ATTRS{idProduct}=="ea60"'` should only match tty devices that have manufacturer and idProduct in the device or its parents.

**Logs (please share snips of applicable logs)**
Snippet from Agent log:
```
[2020-11-08T09:04:14Z TRACE agent::util::config_action] do_periodic_discovery - start for config akri-udev
[2020-11-08T09:04:14Z TRACE agent::util::config_action] do_periodic_discovery - loop iteration for config akri-udev
[2020-11-08T09:04:14Z TRACE agent::protocols::udev::discovery_handler] discover - for udev rules ["SUBSYSTEM==\"tty\", ATTRS{manufacturer}==\"Silicon Labs\", ATTRS{idProduct}
==\"ea60\""]
[2020-11-08T09:04:14Z INFO  agent::protocols::udev::discovery_impl] parse_udev_rule - enter for udev rule string SUBSYSTEM=="tty", ATTRS{manufacturer}=="Silicon Labs", ATTRS{
idProduct}=="ea60"
[2020-11-08T09:04:14Z TRACE agent::protocols::udev::discovery_impl] parse_udev_rule - parsing udev_rule "SUBSYSTEM==\"tty\", ATTRS{manufacturer}==\"Silicon Labs\", ATTRS{idPr
oduct}==\"ea60\""
[...]

PR #115, opened 2020-11-10
Error out Agent for unsupported udev rules

kate-goldenring:
Currently, the udev discovery handler ignores unsupported fields and operations in udev rules, only filtering the devices by the supported fields in the rule. This results in more devices being discovered than intended (as brought up in issue ref #114) and a mismatch between the Configuration and what is actually happening. Instead, the agent will error out when an unsupported field or operation is used, making it more apparent. 
This PR also makes documentation more explicit, emphasizing what udev fields are currently supported. It also makes a few minor spelling fixes in code and inline comments. 

The following errors out the Agent pod as expected (`ATTRS` is currently unsupported):
```sh
helm install akri akri-helm-charts/akri-dev \
    --set agent.image.repository="ghcr.io/.." \
    --set agent.image.tag="v0.0.39-amd64" \
    --set imagePullSecrets[0].name=regcred \
    --set udev.enabled=true \
    --set udev.udevRules[0]='KERNEL=="video[0-9]*"\, ATTRS{manufacturer}=="Some Manufacturer"'
```
**If applicable**:


jiria:
The example in the PR comment should work, it does not contain ATTRs.

docs/udev-configuration.md
```diff
@@ -8,13 +8,9 @@ Akri's udev discovery handler parses udev rules listed in a Configuration, searc
...
-Consequently, the discovery handler will ignore all action operations ("=" , "+=" , "-=" , ":=") and action fields such
-as `TEST` in the udev rules. You should only use match operations ("==",  "!=") and fields such as `KERNEL` and
-`DEVPATH` in your udev rules. Additionally, there are some match fields that look up the device hierarchy, such as
-`SUBSYSTEMS`, that are yet to be supported and are also ignored. See the
-[grammar](../agent/src/protocols/udev/udev_rule_grammar.pest) for the full list of supported fields.
+Consequently, the discovery handler will throw an error if any of the rules contain an action operation ("=" , "+=" , "-=" , ":=") or action fields such as `IMPORT` in the udev rules. You should only use match operations ("==",  "!=") and fields such as `KERNEL` and `DEVPATH` in your udev rules. Additionally, there are some match fields that look up the device hierarchy, such as `SUBSYSTEMS`, that are yet to be supported and will throw an error if used. Support for these will be added soon. To summarize, only the following udev fields are currently supported: `DEVPATH`, `KERNEL`, `TAG`, `DRIVER`, `SUBSYSTEM`, `ATTRIBUTE`, `PROPERTY`.
```
jiria, line 13: Consider pointing to the example*.rules files from here as well.

jiria: Also, can we add a line that would point to some docs on how to validate the rules? My worry is the following:
1) User identifies a device that is to be matched
2) User uses udevadm to list its fields and constructs the UDEV filter
3) User does not know if the filter is too wide, such that it would include other unintended devices
4) User deploys Akri Config and is surprised that it hit more than what was expected

udevadm is great for figuring out the attributes specific to a device, but does not help to ensure that you are not casting a wide net (unless there is some other mode that I missed).

kate-goldenring:
> The example in the PR comment should work, it does not contain ATTRS.

Sorry ATTR does work. I accidentally pasted the example of what works. Thanks for catching this!

docs/udev-configuration.md
kate-goldenring: This is a good point. While `udevadm trigger` supports looking up devices based on the currently supported fields, it does not support looking up the device hierarchy, so you cannot use it to test `ATTRS` when support comes for that, and it does not support regex like udev. Adding in more details as to how to use `udevadm trigger` would only be a temporary and incomplete solution to helping people craft udev rules. As an example, to test the udev rule `--set udev.udevRules[0]='KERNEL=="video1"\, ATTRS{manufacturer}=="Some Manufacturer"'`, you could run
`udevadm trigger --verbose --dry-run --type=devices --name-match=video1 --attr-match=manufacturer="Some Manufacturer"`

A different solution is the following:
Create a new rules file in the `/etc/udev/rules.d/` directory. There should already be the `70-snap.core.rules` file there that is fun to explore. Add the rule[s] you are planning to add to your configuration to it, adding an TAG action to the rule. Then you can search for all devices with that tag. 
1. navigate to `/etc/udev/rules.d`
1. Create a file called `90-akri.rules` and add your rules to it. Note how 90 is the prefix to the file name. This makes sure these rules are run before the others in `70-snap.core.rules`, preventing them from being overwritten.
```sh
# Rule to tag video and sound devices
KERNEL=="video[0-9]*", TAG+="akri_tag"
SUBSYSTEM=="sound", KERNEL=="card[0-9]*", TAG+="akri_tag"
```
1. Reload the udev rules and trigger them
```sh
udevadm control --reload
udevadm trigger
```
1. List the devices that have been tagged, which Akri will discover
```sh
udevadm trigger --verbose --dry-run --type=devices --tag-match=akri_tag
```
1. Modify rule as needed, being sure to reload and trigger the rules each time
1. Remove the tag from devices
```sh
# Rule to tag video and sound devices. Note how  `+=` turns to `-=`.
KERNEL=="video[0-9]*", TAG-="akri_tag"
SUBSYSTEM=="sound", KERNEL=="card[0-9]*", TAG-="akri_tag"
```
1. Reload the udev rules and trigger them
```sh
udevadm control --reload
udevadm trigger
```
1. Confirm the tag has been removed and no devices are listed
```sh
udevadm trigger --verbose --dry-run --type=devices --tag-match=akri_tag
```
1. Create an Akri configuration with your udev rule!

jiria: I presume the first example that you mentioned should not have been `ATTRS{manufacturer}`, but instead `ATTR{manufacturer}`, right? 

I think this is a great guide, we should include it in the docs.

Also, could the tagging approach be used to at least workaround the missing support for parent features? It would mean that somebody has to deploy the udev rule file and refresh udev rules, but then would we be able to just use Akri to detect the specific tag? 

In my case, this is all I get for the leaf device, which is not super useful. Only parent devices have useful attributes.
```
looking at device '/devices/platform/soc/3f980000.usb/usb1/1-1/1-1.5/1-1.5:1.0/ttyUSB0/tty/ttyUSB0':
    KERNEL=="ttyUSB0"
    SUBSYSTEM=="tty"
    DRIVER==""
```

kate-goldenring: Double yes: accidentally added the S again and yes tagging could be a work around

kate-goldenring: Added documentation to `udev-configuration.md`. Let me know what you think.

bfjelds: i might simplify this a bit ... from:

You should only use match operations ("==",  "!=") and fields such as `KERNEL` and `DEVPATH` in your udev rules. Additionally, there are some match fields that look up the device hierarchy, such as `SUBSYSTEMS`, that are yet to be supported and will throw an error if used. Support for these will be added soon. To summarize, only the following udev fields are currently supported: `DEVPATH`, `KERNEL`, `TAG`, `DRIVER`, `SUBSYSTEM`, `ATTRIBUTE`, `PROPERTY`.

to:

You should only use match operations ("==",  "!=") and the following udev fields: `DEVPATH`, `KERNEL`, `TAG`, `DRIVER`, `SUBSYSTEM`, `ATTRIBUTE`, `PROPERTY`.  There are some match fields that look up the device hierarchy, such as `SUBSYSTEMS`, that are yet to be supported and will throw an error if used. Support for these will be added soon.

bfjelds (approved):

bfjelds:
Intresting ... looks like our Helm workflow just went from 3.4.0 to 3.4.1.  And our RBAC ClusterRole bits need changing.  I suspect we can ignore this failure for now or update the Helm workflow (like these: https://github.com/search?q=extension%3Ayaml+azure+setup-helm+v3.4.0) from:

```yaml
- uses: azure/setup-helm@v1
```

to 

```yaml

- uses: azure/setup-helm@v1
  with:
    version: "v3.4.0"
```

if we move this to a branch PR rather than a fork PR, we can test the workflow change too.

kate-goldenring:
> Intresting ... looks like our Helm workflow just went from 3.4.0 to 3.4.1. And our RBAC ClusterRole (namespace should be `rbac.authorization.k8s.io/v1` not `rbac.authorization.k8s.io/v1beta1`) bits need changing.
> 
> There are probably 2 changes that need to be made for this:
> 
> 1. Establish what version of Helm to use in the workflow
> 2. Fix the namespace issue in our RBAC
> 
> We can probably ignore this failure for this PR ...
> 
> Or update the rbac.yaml and the Helm workflow (like these: https://github.com/search?q=extension%3Ayaml+azure+setup-helm+v3.4.0) from:
> 
> ```yaml
> - uses: azure/setup-helm@v1
> ```
> 
> to
> 
> ```yaml
> - uses: azure/setup-helm@v1
>   with:
>     version: "v3.4.0"
> ```
> 
> if we move this to a branch PR rather than a fork PR, we can test the workflow change too.
> 
> **OR**
> 
> We could make the Helm change in separate PR and rebase this one. That might make the most sense. I'll start on the Helm PR.



> Intresting ... looks like our Helm workflow just went from 3.4.0 to 3.4.1. And our RBAC ClusterRole (namespace should be `rbac.authorization.k8s.io/v1` not `rbac.authorization.k8s.io/v1beta1`) bits need changing.
> 
> There are probably 2 changes that need to be made for this:
> 
> 1. Establish what version of Helm to use in the workflow
> 2. Fix the namespace issue in our RBAC
> 
> We can probably ignore this failure for this PR ...
> 
> Or update the rbac.yaml and the Helm workflow (like these: https://github.com/search?q=extension%3Ayaml+azure+setup-helm+v3.4.0) from:
> 
> ```yaml
> - uses: azure/setup-helm@v1
> ```
> 
> to
> 
> ```yaml
> - uses: azure/setup-helm@v1
>   with:
>     version: "v3.4.0"
> ```
> 
> if we move this to a branch PR rather than a fork PR, we can test the workflow change too.
> 
> **OR**
> 
> We could make the Helm change in separate PR and rebase this one. That might make the most sense. I'll start on the Helm PR.

Thanks for diagnosing this. I'll wait for #118 to merge and then rebase.

docs/udev-configuration.md
```diff
@@ -37,6 +33,34 @@ spec:
       -  'SUBSYSTEM=="sound", ATTR{vendor}=="Great Vendor"'
 ```
 
+### Testing a udev rule
+To test which devices Akri will discover with a udev rule, you can run the rule locally adding a tag action to it. Then you can search for all devices with that tag, which will be the ones discovered by Akri.
+1. Create a new rules file called `90-akri.rules` in the `/etc/udev/rules.d` directory, and add your udev rule[s] to it. For this example, we will be testing the rule `SUBSYSTEM=="sound", KERNEL=="card[0-9]*"`. Add `TAG+="akri_tag"` to the end of each rule. Note how 90 is the prefix to the file name. This makes sure these rules are run before the others in the default `70-snap.core.rules`, preventing them from being overwritten. Feel free to explore `70-snap.core.rules` to see numerous examples of udev rules.
```
jiria, line 38: I believe the numeric prefixes are not higher is earlier, they just force ascending ordering, at which point 70 will run before 90.

Also nit: I would vote for (s), not [s].

docs/udev-configuration.md
```diff
@@ -37,6 +33,34 @@ spec:
...
+    ```sh
+    udevadm control --reload
+    udevadm trigger
+    ```
+1. List the devices that have been tagged, which Akri will discover.
+    ```sh
+    udevadm trigger --verbose --dry-run --type=devices --tag-match=akri_tag
+    ```
+1. Modify the rule as needed, being sure to reload and trigger the rules each time.
+1. Remove the tag from the devices -- note how  `+=` turns to `-=` -- and reload and trigger the udev rules. Alternatively, if you are trying to discover devices with fields that Akri does not yet support, such as `ATTRS`, you could leave the tag and add it to the rule in your Configuration with `TAG=="akri_tag"`.
```
jiria, line 52: Should the alternatively be its own section that we could point to?

jiria (approved):

docs/udev-configuration.md
kate-goldenring: good point! meant to say that the higher number ensures they arent overwritten

docs/udev-configuration.md
kate-goldenring: I'd rather work to expanding the protocol to more fields

jiria: Sounds good.

jiria (approved):

PR #121, opened 2020-11-11
Helm workflow should lint using latest Helm version (as warning, not as failure) [SAME VERSION]

bfjelds:
When Helm releases a new version, the linting capabilities change.  We should leverage any new lints as a way of seeing new potential issues in our Helm chart.  At the same time, these shouldn't cause a break in our CI builds.

**If applicable**:

### DIFF
--- a/.github/workflows/run-helm.yml
+++ b/.github/workflows/run-helm.yml
@@ -51,7 +51,7 @@ jobs:
       - uses: azure/setup-helm@v1
 
       - name: Lint helm chart
-        run: helm lint deployment/helm
+        run: helm lint deployment/helme && echo "lint finished successfully" || echo "lint found issues"
 
 
   helm:
--- a/.github/workflows/run-helm.yml
+++ b/.github/workflows/run-helm.yml
@@ -51,7 +51,7 @@ jobs:
       - uses: azure/setup-helm@v1
 
       - name: Lint helm chart
-        run: helm lint deployment/helm && echo "lint finished successfully" || echo "lint found issues"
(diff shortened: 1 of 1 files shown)


kate-goldenring (approved):

.github/workflows/run-helm.yml
```diff
@@ -27,6 +27,33 @@ env:
...
+          ref: refs/pull/${{ github.event.pull_request.number }}/head
+
+      - name: Checkout the head commit of the branch
+        if: ${{ github.event_name != 'pull_request_target' }}
+        uses: actions/checkout@v2
+
+      - uses: azure/setup-helm@v1
+
+      - name: Lint helm chart
+        run: helm lint deployment/helm && echo "lint finished successfully" || echo ::warning::"lint found issues"
```
jiria, line 54: Would the echo hide the failure (overall exit code will be 0); or do you not want this step to cause failure of the pipeline? Wondering if it is not a failure, if it will not get noticed.

jiria: Thinking more, if we are hardcoding the helm version, should we not leave it as failure? Warning would make more sense if we were on the latest. (or should we stop hardcoding now that it would just be a warning?)

bfjelds: there are two parts to this:
1. the hard-coded version: the linting here will cause a failure
1. the non-specified, most recent version: the linting here will not cause a failure and we will have to look to find the warning (it will not show up anywhere)

my intent is to not allow random version changes to Helm cause a failure in our CI.

but to also allow us to see any potential issues with upcoming versions of Helm.  there is no way to bubble up a "warning" without causing a failure in the CI.  so this is my best approximation.

PR #128, opened 2020-11-13
Fix test path

jiria:
Test_list_and_watch assumes it has write permission to /tmp/device-plugins/. This is possibly dangerous and also does not work with multiple users running tests on a single host. Furthermore, the original implementation left old directories behind. The new implementation uses a temporary directory that is unique per test run and will get automatically deleted.

This PR has also minor production code change to accomodate of passing of specialized dp path.

**If applicable**:


agent/src/util/device_plugin_service.rs
```diff
@@ -751,16 +750,17 @@ pub async fn build_device_plugin(
...
     info!("build_device_plugin - entered for device {}", instance_name);
     let capability_id: String = format!("{}/{}", AKRI_PREFIX, instance_name);
     let unique_time = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)?;
     let device_endpoint: String = format!("{}-{}.sock", instance_name, unique_time.as_secs());
-    let socket_path: String = format!(
-        "{}{}",
-        DEVICE_PLUGIN_PATH.to_string(),
-        device_endpoint.clone()
-    );
+    let socket_path: String = Path::new(device_plugin_path)
```
jiria, line 759: Why does this need to be a String?

jiria: Specifically, wouldn't &str be more efficient than cloning String along?

kate-goldenring: Yes, it would, but i think I ran into some lifetime issues that seemed to bubble up to this closure in `serve`: `connect_with_connector(service_fn(move |_: Uri| UnixStream::connect(path.clone())))`. Feel free to give it a go! Would love to know how to make the `&str` lifetime live long enough without declaring it static.

jiria: Ok, will leave probably for another PR when I get a chance.

Please let me know if anything else looks suspicious, Rust newbie here.

agent/src/util/constants.rs
```diff
@@ -11,10 +11,10 @@ pub const K8S_DEVICE_PLUGIN_VERSION: &str = "v1beta1";
 
 /// DevicePluginPath is the folder the kubelet expects to find Device-Plugin sockets. Only privileged pods have access to this path.
 #[cfg(not(test))]
-pub const DEVICE_PLUGIN_PATH: &str = "/var/lib/kubelet/device-plugins/";
+pub const DEVICE_PLUGIN_PATH: &str = "/var/lib/kubelet/device-plugins";
 /// Path for testing `DevicePluginService`
 #[cfg(test)]
-pub const DEVICE_PLUGIN_PATH: &str = "/tmp/device-plugins/";
+pub const DEVICE_PLUGIN_PATH: &str = "dummy";
```
bfjelds, line 17: Is the dummy definition needed?

bfjelds (approved):

agent/src/util/constants.rs
jiria: I think so, the product code still builds when I run cargo test. I could conditionally compile the code that uses it away when building test, but that did not sound right. How do you propose to solve this?

bfjelds: Just one definition of the variable, no cfg attribute?

jiria: Good point. Was thinking this will not make folks assume that the value is used for test case. But I think the simplification is better option. Will take out the test value.

kate-goldenring (approved):

PR #147, opened 2020-11-24
Remove privileged security context from Akri Agent

kate-goldenring:
While [Kubernetes documentation](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) states that Device Plugins must run in a privileged security context, this does not seem to be the case. For example, by default, the [Nvidia Device Plugin](https://github.com/NVIDIA/k8s-device-plugin/tree/v0.7.1) runs unprivileged unless the CPU manager is enabled. All that is required it that the device plugin directory `/var/lib/kubelet/device-plugins` is mounted into the Agent. 

This PR reduces the Agent attack surface by removing the privileged security context. 

Tested with udev and ONVIF.

**If applicable**:


.github/workflows/run-test-cases.yml
```diff
@@ -179,9 +179,6 @@ jobs:
         sudo microk8s.enable helm3
         sudo microk8s.enable rbac
         sudo microk8s.enable dns
-        echo "--allow-privileged=true" | sudo tee -a /var/snap/microk8s/current/args/kube-apiserver
-        sudo microk8s.stop
-        sudo microk8s.start
```
bfjelds, line 184: is the restart needed for enabling rbac or dns or helm3?

deployment/helm/templates/agent.yaml
```diff
@@ -61,8 +61,6 @@ spec:
             mountPath: /host/var/run/dockershim.sock
           - name: devices
             mountPath: /run/udev
-        securityContext:
-          privileged: true
```
bfjelds, line 65: is there any reason to leave this in with configurable parameter in values.yaml (that defaults to false)?

.github/workflows/run-test-cases.yml
kate-goldenring: No. I dont think it is. MicroK8s docs says "enable add-ons" and then doesnt specify stop and restart: https://microk8s.io/docs

bfjelds: 👍

deployment/helm/templates/agent.yaml
kate-goldenring: I don't see a reason right now to have it privileged. But i could do the same thing i do in [this pr](https://github.com/deislabs/akri/pull/148) and make the security context configurable.

bfjelds: a simplified version of that pr is what i was suggesting.

jiria (approved):

PR #148, opened 2020-11-24
Configure broker security context

kate-goldenring:
Instead of making udev brokers run privileged, users may want to configure their broker pods to run with certain user/group privileges. The PR allows those configurations.

**If applicable**:


docs/udev-configuration.md
```diff
@@ -136,6 +136,21 @@ Installation](./customizing-akri-installation.md) to learn how to [modify the br
...
+Helm. For example, to instead run all processes in the Pod with user ID 1000 and group 3000, do the following: 
+```bash
+helm repo add akri-helm-charts https://deislabs.github.io/akri/
+helm install akri akri-helm-charts/akri-dev \
+    --set useLatestContainers=true \
+    --set udev.enabled=true \
+    --set udev.udevRules[0]='SUBSYSTEM=="sound"\, ATTR{vendor}=="Great Vendor"' \
+    --set udev.brokerPod.image.repository=nginx \
+    --set udev.brokerPod.securityContext.runAsUser=1000 \
+    --set udev.brokerPod.securityContext.runAsGroup=3000
```
jiria, line 151: Would set the group to 1000 as well, as that is the usual default group.

docs/end-to-end-demo.md
```diff
@@ -97,7 +97,10 @@ carry out one or the other (or adapt to your distribution), then continue on wit
     alias kubectl='microk8s kubectl'
     alias helm='microk8s helm3'
     ```
-1. Enable privileged pods and restart microk8s.
+1. For the sake of this demo, the udev video broker pods run privileged to easily grant them access to video devices, so
```
jiria, line 100: When you set the rw permission as part of DeviceSpec, I have noticed that Kubelet adds those to others as well, so the privileged should not be needed if you could pass the permissions as part of the Configuratoin.

kate-goldenring: Great. It seems like this should be left as is for now and then once the Agent starts using DeviceSpec instead of Mounts and permissions are made settable via the Configuration, we can remove this?

bfjelds (approved):

docs/end-to-end-demo.md
jiria: Right, makes sense.

jiria (approved):

docs/udev-configuration.md
```diff
@@ -136,6 +136,21 @@ Installation](./customizing-akri-installation.md) to learn how to [modify the br
 spec](./customizing-akri-installation.md#modifying-the-brokerpodspec) and [service
 specs](./customizing-akri-installation.md#modifying-instanceservicespec-or-configurationservicespec) in the Configuration.
 
+### Setting the broker Pod security context
+By default in the generic udev Configuration, the udev broker is run in privileged security context. This container
+[security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) can be customized via
+Helm. For example, to instead run all processes in the Pod with user ID 1000 and group 3000, do the following:
```
jiria, line 142: Consider updating the group here as well.

PR #157, opened 2020-12-03
Optimize build and test all supported versions

bfjelds:
For PRs, we are building containers 3x to test 3 flavors of Kubernetes.  Using artifacts, we should be able to build one time and test as many flavors as we want.

**If applicable**:

### DIFF
--- a/.github/workflows/run-test-cases.yml
+++ b/.github/workflows/run-test-cases.yml
@@ -134,7 +134,7 @@ jobs:
           - K3s-1.18
           - K3s-1.19
           - MicroK8s-1.18
-          # - MicroK8s-1.19
+          - MicroK8s-1.19
           - Kubernetes-1.16
           - Kubernetes-1.17
           - Kubernetes-1.18
@@ -144,8 +144,8 @@ jobs:
         include:
           - kube-runtime: MicroK8s-1.18
             kube-version: 1.18/stable
-          # - kube-runtime: MicroK8s-1.19
-          #   kube-version: 1.19/stable
+          - kube-runtime: MicroK8s-1.19
+            kube-version: 1.19/stable
           - kube-runtime: K3s-1.18
             kube-version: v1.18.9+k3s1
           - kube-runtime: K3s-1.19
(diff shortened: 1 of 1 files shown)


kate-goldenring:
Should we test supported rust version too?

bfjelds:
> Should we test supported rust version too?

i don't think that this PR is the place for that, i think that belongs in rust workflow maybe?

.github/workflows/run-test-cases.yml
```diff
@@ -58,25 +59,88 @@ on:
...
+        BUILD_ARM32: 0
+        BUILD_ARM64: 0 
+        PREFIX: ghcr.io/deislabs/akri
+        LABEL_PREFIX: pr
+      run: |
+        make akri-build
+        make controller-build-amd64
+        make agent-build-amd64
+        docker save ${PREFIX}/agent:${LABEL_PREFIX}-amd64 > agent.tar
+        docker save ${PREFIX}/controller:${LABEL_PREFIX}-amd64 > controller.tar
```
jiria, line 104: Should we add logout?

jiria (approved):

.github/workflows/run-test-cases.yml
bfjelds: i'm not clear what you have in mind?  do you mean something like:

echo "Saving ${PREFIX}/agent:${LABEL_PREFIX}-amd64 to agent.tar"
docker save ${PREFIX}/agent:${LABEL_PREFIX}-amd64 > agent.tar

bfjelds: this isn't intended to fix the credential thing.  is that what you are referring to?

jiria: Yes. If you think it would not help anything, fine.

bfjelds: ok cool ... i agree we need to fix that, but i think it would be better to address that more holistically than here.

.github/workflows/run-test-cases.yml
```diff
@@ -58,25 +59,102 @@ on:
...
+
+    steps:
+    - name: Checkout the merged commit from PR and base branch
+      uses: actions/checkout@v2
+      if: github.event_name == 'pull_request_target'
+      with:
+        # pull_request_target is run in the context of the base repository
+        # of the pull request, so the default ref is master branch and
+        # ref should be manually set to the head of the PR
+        ref: refs/pull/${{ github.event.pull_request.number }}/head
```
kate-goldenring, line 78: Whenever checking out the repo, if possible try to set ` persist-credentials: false` like here https://github.com/deislabs/akri/blob/63010918c04ba4042e86953277870e538f0d476e/.github/workflows/build-video-streaming-app-container.yml#L71 for added security. It is enabled by default.

bfjelds: 👍

PR #161, opened 2020-12-04
Remove pull_request_target triggers and minimize credential use [SAME VERSION] [IGNORE INTERMEDIATE BUILDS]

bfjelds:
Using pull_request_target with any secrets is risky at best.

The trigger was added because we needed to log into docker to avoid docker's new anonymous pull rate limits.  However, github has come to an agreement with docker to avoid this (https://github.com/actions/virtual-environments/issues/1445#issuecomment-713861495), so we no longer need to invoke `docker login` before any base image pull in our workflows.  That removes the requirement for most of our credential use.

The remainder of our credential use is for either:
1. accessing our intermediate build images
1. pushing new container images

We should make our intermediate build container images public.  This would simplify our users' development story, and reduce our need for `docker login` in our workflows.

Pushing new containers is limited to merges (push) and releases.  Neither of which involve pull_request_target.

We should remove pull_request_target triggers.

**If applicable**:

### DIFF
--- a/.github/actions/build-component-multi-arch/action.yml
+++ b/.github/actions/build-component-multi-arch/action.yml
@@ -20,7 +20,7 @@ inputs:
     description: Component prefix used by Makefile
     required: true
   github_event_name:
-    description: Specify the github event name (push, pull_request_target, release, etc)
+    description: Specify the github event name (push, pull_request, release, etc)
     required: true
 runs:
   using: 'node12'
--- a/.github/actions/build-component-multi-arch/action.yml
+++ b/.github/actions/build-component-multi-arch/action.yml
@@ -22,12 +22,6 @@ inputs:
   github_event_name:
     description: Specify the github event name (push, pull_request_target, release, etc)
     required: true
-  dockerhub_username:
-    description: Dockerhub name
-    required: true
-  dockerhub_password:
-    description: Dockerhub password
-    required: true
 runs:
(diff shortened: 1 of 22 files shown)


kate-goldenring:
This does not solve the issue of someone modifying main.js in a PR, correct? We are still passing our gh secrets to those files.

bfjelds:
> This does not solve the issue of someone modifying main.js in a PR, correct? We are still passing our gh secrets to those files.

It does address that.

`pull_request_target` has access to secrets, a branch `pull_request` has access to secrets (but can only be created by an internal contributor), a forked `pull_request' (an external PR) does not.

The PR eliminates `pull_request_target` triggers, so no forked PR will have access to secrets anymore.

When a fork `pull_request` is triggered, it will get an empty string anytime it requests a secret.  This was a problem for us when our intermediate containers were private because we needed the secret to get the containers (or when we needed to log into dockerhub to avoid their anonymous pull rate limits).  But now that we made the intermediate containers public (and github smoothed over the rate limit issue), the credentials are no longer needed for the build phase of our actions ... which are the only parts of the actions that execute during a PR.

So, if a forked PR made a change to an main.js, the PR would get an empty string as input.  If the change to main.js was intended to steal the secret, it would steal a blank string rather than our secret.

kate-goldenring (approved):

Issue #132, opened 2020-11-17
[Extensibility] ZeroConf


DazWilkin:
**Is your feature request related to a problem? Please describe.**g

I've begun prototyping (!) a ZeroConf protocol implementation for Akri

**Is your feature request related to a way you would like Akri extended? Please describe.**

[ZeroConf](https://en.m.wikipedia.org/wiki/Zero-configuration_networking) is probably (!) a useful protocol implementation particularly due to its reliable naming of (transient) devices.

**Describe the solution you'd like**

Akri protocol implementation with Broker samples.

kate-goldenring:
@DazWilkin, can you put in a PR with a proposal for what you envision? Maybe something like the one for [OPC UA](https://github.com/deislabs/akri/blob/main/docs/proposals/opcua.md) that provides background on zero conf, what filters/settings will be a part of the Configuration, and any security considerations?

DazWilkin:
Yes, I can draft something.

To set expectations: I'm mostly unfamiliar with ZeroConf and am writing more of a proof-of-concept using it to learn more about ZeroConf, Akri and Rust. I don't have any production uses of ZeroConf.

It would be useful to read a proposal written by folks that are familiar with ZeroConf and who use it in production.

I'll draft a proposal and include programming notes, concerns, omissions, feature requests etc.

https://github.com/DazWilkin/akri/blob/proposal-zeroconf/docs/proposals/zeroconf.md

DazWilkin:
@kate-goldenring @bfjelds 
Should I raise this as a separate issue? Or would you prefer I continue to keep ZeroConf-related questions here?

The [zeroconf](https://crates.io/crates/zeroconf) crate has a bunch of Linux dependencies, see https://github.com/deislabs/akri/issues/104#issuecomment-729264997 **and** it requires that the agent be started with several daemons.

`Dockerfile.agent` was:

```Dockerfile
ENTRYPOINT ["./agent"]
```

To:

```Dockerfile
ENTRYPOINT ["bash","-c","service dbus start && service avahi-daemon start && ./agent"]
```

This uncommon approach yields a shell running the 2 services and the agent.

Problems:

+ As mentioned in #104, more agents potentially bulks out the agent image
+ Two (more) services for ZeroConf but what happens as other protocols add more

kate-goldenring:
Hey Daz,

I think we can keep this issue as our thread for Zeroconf questions.

[...]

PR #163, opened 2020-12-08
[Extensibility] Zeroconf

DazWilkin:
Implements: https://github.com/deislabs/akri/issues/132

Provides a Linux-only Zeroconf protocol implementation for Akri

I've tried (!) to avoid repeating some of the commitment feedback from `http-extensibility`.

**If applicable**:
- [X] this PR contains documentation
- [X] inline documentation builds (`cargo doc`)


shared/src/akri/configuration.rs
```diff
@@ -45,6 +46,13 @@ fn default_action() -> FilterType {
     FilterType::Include
 }
 
+/// This defines the Zeroconf data stored in the Configuration
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct ZeroconfDiscoveryHandlerConfig {
+    pub filter: String,
```
bfjelds, line 53: it looks like the zeroconf crate supports several filters and this is leveraging that ... is the crate's filter syntax well known?  if not, we might consider using the filter constructs that onvif and udev use (which allow inclusion and exclusion).

samples/brokers/zeroconf/kubernetes/zeroconf.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: akri.sh/v0
```
bfjelds, line 1: we should move this to deployment/helm/templates/zeroconf.yaml

samples/brokers/zeroconf/Dockerfiles/standalone
```diff
@@ -0,0 +1,50 @@
+ARG PLATFORM=amd64
```
bfjelds, line 1: move this to build/containers/Dockerfile.zeroconf-generic

samples/brokers/zeroconf/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
bfjelds, line 1: is this needed?

.github/workflows/build-zeroconf-broker-container.yaml
```diff
@@ -0,0 +1,29 @@
+name: "Build Zeroconf Protocol Brokers"
```
bfjelds, line 1: i think for this protocol, we probably want to integrate it more than with our extensibility examples (http, nessie).

to that end, we should update `build/akri-containers.mk` to build the zeroconf broker and create a workflow that uses our actions like `.github/workflows/build-udev-video-broker-container.yml`

.dockerignore
```diff
@@ -2,6 +2,10 @@
 # Ignore everything
 **
 
+!samples/brokers/zeroconf
```
bfjelds, line 5: if we integrate zeroconf as an "inbox" protocol (update root Cargo.toml, use our build workflow actions), we shouldn't need to do this (the existing Rust code for agent, controller, & brokers is not built with a docker build command).

bfjelds:
This is great, great stuff ... thank you!

The biggest initial feedback I have is that I'd prefer to have zeroconf be an integrated part of Akri rather than an extensibility example ... so I'd ask that it be fit into the existing makefiles/workflows/actions.

shared/src/akri/configuration.rs
DazWilkin: Oh, I like this. I wish I'd thought of it. It's cleaner and clearer than what I have. I'll incorporate. Thank you.

DazWilkin: I think much better, thank you!

Fixed: https://github.com/deislabs/akri/pull/163/commits/ea19332ca2ba670dc79904d646929b8d5f4f82e7

samples/brokers/zeroconf/.gitignore
DazWilkin: It is not.

Fixed: https://github.com/deislabs/akri/pull/163/commits/91def2a5f886cd36ca208c63d9baa77df725e008

agent/src/protocols/zeroconf/discovery_handler.rs
```diff
@@ -27,6 +25,35 @@ const DEVICE_HOST: &str = "AKRI_ZEROCONF_DEVICE_HOST";
...
 
+fn filter(config: &ZeroconfDiscoveryHandlerConfig, service: &ServiceDiscovery) -> bool {
+    trace!("[zeroconf:filter] Service: {:?}", service);
+    let include = (if let Some(name) = &config.name {
+        let result = name == service.name();
+        trace!("[zeroconf:filter] Name ({}) [{}]", name, result);
+        result
+    } else {
+        true
+    }) && (if let Some(domain) = &config.domain {
```
bfjelds, line 36: i'm not familiar with zeroconf, is it possible to filter on name and domain?  if so, you might consider restructuring this.

deployment/helm/crds/akri-configuration-crd.yaml
```diff
@@ -18,6 +18,17 @@ spec:
                 protocol: # {{ProtocolHandler}}
                   type: object
                   properties:
+                    zeroconf: # {{ZeroconfDiscoveryHandler}}
+                      type: object
+                      properties:
+                        kind:
+                          type: string
+                        name:
+                          type: string
```
bfjelds, line 27: would making these string[] make sense?  would someone want to deploy one broker to handle zeroconf devices with several names/domains?

for onvif, i found having the ability to exclude to be helpful as well.  i can imagine wanting your broker to handle every device of kind=foo except ones in domain=blah.

bfjelds: maybe a vNext sort of thing?

DazWilkin: I was going for simple(st).

Assuming we have 6 services (on `local`):
```console
a _http._tcp 7777 local project=akri protocol=x location=foo # HTTP
b _rust._tcp 8888 local project=akri protocol=x location=bar
c _rust._tcp 9999 local project=akri protocol=x location=baz

d _rust._tcp 7777 local project=akri protocol=y location=foo
e _rust._tcp 8888 local project=akri protocol=y location=bar
f _http._tcp 9999 local project=akri protocol=y location=baz # HTTP
```

I propose (!?):

For `protocol=x` we'd deploy one configuration:

```YAML
spec:
  protocol:
    zeroconf:
      kind: "_rust._tcp"
      domain: "local"
      txtRecords:
        project: akri
        protocol: x
```

... and for `protocol=y` another

And, because, in the above, we're filtering by `kind=_rust._tcp`, we'd only see 2 services on each Instance (`_http._tcp` would be excluded).

And, because the Instance includes an environment variable for each TXT record, within any `protocol=[x|y]`, we'd be able to infer location `AKRI_ZEROCONF_DEVICE_LOCATION=[foo|bar|baz]`.

What do you think?

DazWilkin:
@bfjelds @kate-goldenring

One thing you're not seeing in the PR is the changes to `./build/containers/Dockerfile.agent` that are necessary to include the dependencies for the `zeroconf` crate (for Linux). There would be a different set of dependencies for e.g. `astro-dnssd` crate (for Linux, for Windows, for Mac OS).

```Dockerfile
ARG PLATFORM=amd64
ARG CROSS_BUILD_TARGET=x86_64-unknown-linux-gnu
FROM ${PLATFORM}/debian:buster-slim
ARG CROSS_BUILD_TARGET
RUN echo "Creating container based on ${PLATFORM}/debian:buster-slim"
RUN echo "Using Rust binaries from ${CROSS_BUILD_TARGET}"

# Link the container to the Akri repository
LABEL org.opencontainers.image.source https://github.com/deislabs/akri

# Copy over container legal notice
COPY ./build/container-images-legal-notice.md .

RUN apt-get update && \
    apt-get install -y --no-install-recommends \
    avahi-daemon \
    clang \
    libavahi-client-dev \
    libclang-3.9-dev \
    libssl-dev \
    libxcb-shape0-dev \
    libxcb-xfixes0-dev \
    xorg-dev \
    llvm-dev \
    openssl \
    && \
    apt-get clean

COPY ./target/${CROSS_BUILD_TARGET}/release/agent /agent

ENV RUST_LOG agent,akri_shared

ENTRYPOINT ["bash", "-c", "service dbus start && service avahi-daemon start && ./agent"]
```

> **NOTE** These additions include those [deps](https://crates.io/crates/zeroconf#prerequisites) specified by `zeronconf` and others that I needed.

As I mentioned previously, I'm concerned about this (unavoidable) bloat of Akri Agent due to its monolithic nature.

Q1: Are you willing to accept these deps as a consequence of adding new discovery handlers?
Q2: Are you willing to accept these deps for the `zeroconf` crate (Linux-only) when there may be different deps for `astro-dnssd`?

One option is that I don't pursue integrating `zeroconf` at this time into Akri's core build process to avoid these issues.

I'm unfamiliar with make and there's quite some complexity in the makefiles in this project. If you'd like me to proceed, I can have a stab at it but I may need some help, OK?

bfjelds:
@DazWilkin, that explanation is super helpful.

Can you quantify the bloat a little?  Maybe let us know what the agent footprint is before the change and after the change (maybe for both debug and release, but most importantly for release).

I think if the bloat isn't obscene, we are happy to take it on for the moment.

In the near future, we hope to implement a solution to our monolithic problem :)

DazWilkin:
Before the change is what you have now.

With `zeroconf`, the agent is running 829MB. It's probably that this could be improved but I'm unfamiliar with these deps.

I think perhaps, more importantly, agent users would get e.g. `zeroconf` regardless (as p/o the monolith) whether they want it.

And so they get any increased size but they also get its security profile and risk its CVEs.

To be clear, this is a consequence of the monolith not zeroconf but it raises the reasonable (enterprise) customer concern "Why do we need to patch the agent for a zeroconf exploit that we're not using?"

DazWilkin:
I have an old `v0.0.38-amd64` of the agent that is 126MB.

bfjelds:
I just built locally and see that without your changes:

debug (cargo build --bin=agent):                 352M
release (cargo build --bin=agent --release):    23M

can you build with `--release` and report how big it is?

DazWilkin:
I get 21M (!?) but isn't the issue the image size?

bfjelds:
> I get 21M (!?) but isn't the issue the image size?

You are totally right.  Sorry, the old app programmer in me hasn't fully given way to this new world yet.

DazWilkin (commented):
> > I get 21M (!?) but isn't the issue the image size?
> 
> You are totally right. Sorry, the old app programmer in me hasn't fully given way to this new world yet.

Nothing to apologize for :-)

bfjelds:
> > > I get 21M (!?) but isn't the issue the image size?
> > 
> > 
> > You are totally right. Sorry, the old app programmer in me hasn't fully given way to this new world yet.
> 
> Nothing to apologize for :-)

With a bit more time thinking, my gut says that we should hit the pause button on this until we get a different extensibility model that would allow this provider to be more on-demand.  But I'll pull more people into this discussion to see if we can get more heads involved.

kate-goldenring:
> > > > I get 21M (!?) but isn't the issue the image size?
> > > 
> > > 
> > > You are totally right. Sorry, the old app programmer in me hasn't fully given way to this new world yet.
> > 
> > 
> > Nothing to apologize for :-)
> 
> With a bit more time thinking, my gut says that we should hit the pause button on this until we get a different extensibility model that would allow this provider to be more on-demand. But I'll pull more people into this discussion to see if we can get more heads involved.

I agree that this might expand the size of the Agent too much. Maybe we can do the following:
1. Investigate which of the Zeroconf crate dependencies is blowing up the size
1. Determine whether the astro-dns dependencies are smaller. @DazWilkin did you resolve the issue you were having getting astro-dns to work
1. If neither of the first 2 are conclusive, wait until we have an extensibility model

DazWilkin:
Comment threading would be useful, nah?

The `zeroconf` crate stipulates dependencies ([link](https://crates.io/crates/zeroconf#prerequisites)) but, I have been able to tweak these:

+ Build: `clang`, `libavahi-client-dev`
+ Run: `avahi-daemon`, `libavahi-client-dev`

The `agent`'s Dockerfile does not appear to be multistage but adding:

+ Agent: `clang`, `libavahi-client-dev`, `avahi-daemon`

Reduces the image size to 476MB

For a multi-stage build test, the image size was 206MB but this may be unrealistic.

I've received no response to my issue on `astro-dnssd`.

The issue is likely to arise with any extensions that are added to the Agent while it's a monolith. Zeroconf is likely just the first.

deployment/helm/crds/akri-configuration-crd.yaml
bfjelds: Simple seems fine.  It isn't clear to me whether people would prefer to have 1 configuration handle a lot of instances or if they want a closer to 1:1 relationship between configurations and instances.  We can adjust if usage dictates a different approach. :)

bfjelds:
> The issue is likely to arise with any extensions that are added to the Agent while it's a monolith. Zeroconf is likely just the first

No doubt.  Getting a working gRPC interface into the Agent to allow out-of-proc discovery would be a valuable addition ASAP!!

kate-goldenring:
Hi @DazWilkin ! The new `DiscoveryHandler` model has been merged. Its a big change, so I put together [a document](https://hackmd.io/@akri/r1NjkIFQO) with the steps to get this PR into the new mold. Let me know how I can help --  I could put PR's into yours with the nitty stuff like a Helm template for the Configuration. Rebasing is going to be tricky, and I wonder if creating a new PR is better.

DazWilkin:
Awesome work!

As you know I've had a look at your HTTP example and it's **great**.

I think I/we should close this PR.

I'm working on another project at the moment and it's unclear when I'll get time to rewrite the Zeroconf protocol.

If I do get an opportunity, this will be top of my Akri list.

kate-goldenring:
> Awesome work!
> 
> As you know I've had a look at your HTTP example and the gRPC-based `DiscoveryHandler` is **great**.
> 
> I think I/we should close this PR.
> 
> I'm working on another project at the moment and it's unclear when I'll get time to rewrite the Zeroconf protocol.
> 
> If I do get an opportunity, this will be top of my Akri list.

Sounds good to me!

PR #170, opened 2020-12-10
Configuration label to instance services

kate-goldenring:
Currently, it is hard to get all services associated with a configuration. Say an application wanted to grab each instance service of a Configuration, it would need to regex scan the names of the services. This adds a label with the associated configuration to the instance level services.

**If applicable**:


shared/src/k8s/service.rs
```diff
@@ -136,6 +136,10 @@ pub fn create_new_service_from_spec(
             AKRI_INSTANCE_LABEL_NAME.to_string(),
             instance_name.to_string(),
         );
+        labels.insert(
+            AKRI_CONFIGURATION_LABEL_NAME.to_string(),
```
bfjelds, line 140: somewhere, we test what labels get added, that test should be updated to validate this

bfjelds:
this change worries me.

if there is code that searches for the config svc now, this change will confuse it (as the instance svcs will now have this property that formerly was unique to the config svc).  i think `create_or_update_service` will stop working correctly.

a possible solution would be to create a separate label name for instance svcs to use.

bfjelds (changes requested):
see comments regarding using AKRI_CONFIGURATION_LABEL_NAME in instance svcs.

bfjelds:
> > this change worries me.
> > if there is code that searches for the config svc now, this change will confuse it (as the instance svcs will now have this property that formerly was unique to the config svc). i think `create_or_update_service` will stop working correctly.
> > a possible solution would be to create a separate (other than AKRI_CONFIGURATION_LABEL_NAME) label name for instance svcs to use.
> 
> Good catch. How about this for the label:
> 
> ```
> pub const AKRI_SERVICE_CONFIGURATION_LABEL_NAME: &str = "akri.sh/service-configuration";
> ```

that could work.

another alternative would be to change the code.  an instance svc would have config and instance labels set, a config svc would only have config label set.

the benefit to changing the code is that there would be one label that is used for configurations in pods and svcs (and it would be consistent).

the lousy part of changing the code is that it'd be more work.

kate-goldenring:
> > > this change worries me.
> > > if there is code that searches for the config svc now, this change will confuse it (as the instance svcs will now have this property that formerly was unique to the config svc). i think `create_or_update_service` will stop working correctly.
> > > a possible solution would be to create a separate (other than AKRI_CONFIGURATION_LABEL_NAME) label name for instance svcs to use.
> > 
> > 
> > Good catch. How about this for the label:
> > ```
> > pub const AKRI_SERVICE_CONFIGURATION_LABEL_NAME: &str = "akri.sh/service-configuration";
> > ```
> 
> that could work.
> 
> another alternative would be to change the code. an instance svc would have config and instance labels set, a config svc would only have config label set.
> 
> the benefit to changing the code is that there would be one label that is used for configurations in pods and svcs (and it would be consistent).
> 
> the lousy part of changing the code is that it'd be more work.

You're saying change the code in  `create_or_update_service`?

bfjelds:
> > > > this change worries me.
> > > > if there is code that searches for the config svc now, this change will confuse it (as the instance svcs will now have this property that formerly was unique to the config svc). i think `create_or_update_service` will stop working correctly.
> > > > a possible solution would be to create a separate (other than AKRI_CONFIGURATION_LABEL_NAME) label name for instance svcs to use.
> > > 
> > > 
> > > Good catch. How about this for the label:
> > > ```
> > > pub const AKRI_SERVICE_CONFIGURATION_LABEL_NAME: &str = "akri.sh/service-configuration";
> > > ```
> > 
> > 
> > that could work.
> > another alternative would be to change the code. an instance svc would have config and instance labels set, a config svc would only have config label set.
> > the benefit to changing the code is that there would be one label that is used for configurations in pods and svcs (and it would be consistent).
> > the lousy part of changing the code is that it'd be more work.
> 
> You're saying change the code in `create_or_update_service`?

At least that function...might take more, I'm not sure.

kate-goldenring:
Closing due to being inactive. Will re-open when circle back to providing support.

PR #171, opened 2020-12-11
Add instance polling to streaming sample

jiria:
This PR extends the streaming app sample so that it can dynamically poll for new cameras and refresh the view to automatically reflect changes.

It also contains minor fixes that remove various warnings.

**If applicable**:


samples/apps/video-streaming-app/streaming.py
```diff
@@ -16,54 +16,47 @@
...
 def index():
-    return render_template('index.html', camera_count = camera_count)
+    global camera_frame_queues
+    return render_template('index.html', camera_count=len(camera_frame_queues)-1)
+    
+@app.route('/camera_list')
+# Returns the current list of cameras to allow for refresh
+def camera_list():
+    global small_frame_sources
+    logging.info(small_frame_sources)
```
kate-goldenring, line 53: Maybe append a description to this log

deployment/helm/templates/rbac.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
   name: akri-agent-sa
 ---
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
bfjelds, line 13: we should update the version for this change

samples/apps/video-streaming-app/streaming.py
```diff
@@ -16,54 +16,47 @@
...
-    for camera_id in range(1, camera_count + 1):
-        url = "{0}:80".format(os.environ['CAMERA{0}_SOURCE_SVC'.format(camera_id)])
-        small_frame_sources.append(url)
-
-for camera_id in range(camera_count + 1):
-    camera_frame_queues.append(queue.Queue(1))
+def get_camera_list(configuration_name):
+    camera_list = []
+    config.load_incluster_config()
+    coreV1Api = client.CoreV1Api()
```
kate-goldenring, line 28: Couldn't we query the env vars set on the node periodically for services (like in the previous implementation) instead of connecting to the K8s API?

bfjelds: a node's env vars are not updated.

kate-goldenring: Is it that the container only receives the state of the nodes env vars when first deployed?

bfjelds:
fwiw, i think with this change, we could add this to the helm chart.

bfjelds:
should we apply this to samples/apps/anomaly-detection-app as well?

samples/apps/video-streaming-app/streaming.py
bfjelds: i'd be speculating on the hows or the whys, but I found that if you:

1. start serviceA
1. start podA, it has serviceA details in its env vars
1. start serviceB
1. start podB, it has serviceA and serviceB details in env vars
1. podA never gets serviceB details in env vars

jiria: My understanding is that the envs reflect only the state at the start of the pod. That makes sense, as the env vars and injected into the process, no way to modify after the process starts.

deployment/helm/templates/rbac.yaml
jiria: Version bumped.

deployment/samples/akri-video-streaming-app.yaml
```diff
@@ -17,16 +17,18 @@ spec:
...
+        # your specific cameras) or 
+        # specify a Akri configuration name to pick up cameras automatically
         # - name: CAMERAS_SOURCE_SVC
-        #   value: "all-cameras-svc"
+        #   value: "akri-udev-video-svc"
         # - name: CAMERA_COUNT
         #   value: "2"
         # - name: CAMERA1_SOURCE_SVC
-        #   value: "akri-onvif-svc"
+        #   value: "akri-udev-video-5c8604-svc"
```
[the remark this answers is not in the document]
jiria: Switched to generic <id>

samples/apps/video-streaming-app/streaming.py
jiria: Dropped this log, this was just for debugging.

jiria:
> fwiw, i think with this change, we could add this to the helm chart.

As discussed offline, for this I might need to add some retry logic to poll for the configuration service.

bfjelds (approved):

kate-goldenring (approved):

PR #173, opened 2020-12-12
Cleanup streaming app sample

jiria:
This PR streamlines the streaming app sample so that both main camera feed and small camera feeds can be dynamically refreshed based on detected cameras

Further work could be done to integrate with Helm and patch up the RBAC to target any namespace


samples/apps/video-streaming-app/streaming.py
```diff
@@ -19,13 +19,148 @@
...
+    ret = coreV1Api.list_service_for_all_namespaces(watch=False)
+    for svc in ret.items:
+        if svc.metadata.name == configuration_name + "-svc":
+            grpc_ports = list(
+                filter(lambda port: port.name == "grpc", svc.spec.ports))
+            if (len(grpc_ports) == 1):
+                url = "{0}:{1}".format(svc.spec.cluster_ip, grpc_ports[0].port)
+                camera_display.main_camera = CameraFeed(url)
+
     ret = coreV1Api.list_service_for_all_namespaces(watch=False)
```
bfjelds, line 164: can we consolidate the 2 for loops here?  or call list_service_for_all_namespaces once?

jiria: Thanks, you are right. They were originally at two different places and I missed refactoring those. Coming up shortly.

bfjelds (approved):

samples/apps/video-streaming-app/streaming.py
```diff
@@ -19,13 +19,148 @@
...
+        self.queue = queue.Queue(1)
+        self.thread = None
+        self.stop_event = global_stop_event
+
+    def __eq__(self, other):
+        if other is None:
+            return False
+        return self.url == other.url
+
+    def StartHandler(self):
```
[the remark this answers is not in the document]
jiria: Thanks for pointing out, will fix.

kate-goldenring (approved):

jiria:
What do you propose for the Helm integration? Or leave that for another PR? @kate-goldenring @bfjelds

bfjelds:
> What do you propose for the Helm integration? Or leave that for another PR? @kate-goldenring @bfjelds

helm integration to me is moving the samples yaml into the deployment/helm/templates folder and adding some parameterization.  i think it is fine to be a separate PR.  i also think we should update the ocpua monitoring app with this same type of logic if we can (and then add it to helm)

kate-goldenring:
> > What do you propose for the Helm integration? Or leave that for another PR? @kate-goldenring @bfjelds
> 
> helm integration to me is moving the samples yaml into the deployment/helm/templates folder and adding some parameterization. i think it is fine to be a separate PR. i also think we should update the ocpua monitoring app with this same type of logic if we can (and then add it to helm)

A separate PR to do these two steps makes sense to me. the parameters could be `deploy-streaming-sample` and `deploy-anomaly-sample` or something along those lines

kate-goldenring:
@jiria, can you bump the version twice with `version.sh -u -p`. You want it to be 0.0.1 greater than https://github.com/deislabs/akri/blob/main/version.txt.

jiria:
Version updated, sorry for the delay.

romoh (approved):

romoh (approved):
The workflow is failing on the security audit check. I sent a [PR to fix this](https://github.com/deislabs/akri/pull/247).

bfjelds:
it looks like this needs another version bump

PR #190, opened 2021-01-07
Prometheus

kate-goldenring:
Adds support for Akri monitoring via [Prometheus](https://prometheus.io/). 

The Akri Controller and Agent publish metrics to port 8080 at a `/metrics` endpoint. All the [default process metrics](https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics), such as Agent or Controller total CPU time usage (`process_cpu_seconds_total`) and RAM usage (`process_resident_memory_bytes`) are exposed, along with the following custom metrics, all of which are prefixed with `akri`. 
| Metric Name | Metric Type | Metric Source | Buckets |
|---|---|---|---|
| akri_instance_count | IntGaugeVec | Agent | Configuration, shared | 
| akri_discovery_response_time_bucket | HistogramVec | Agent | Configuration | 
| akri_broker_pod_count | IntGaugeVec | Controller | Configuration, Node |

While these metrics will always be served by the Agent and Controller -- this could be configurable in the future -- they are exposed to Prometheus when `prometheus.enabled=true` is set. 

Includes documentation on how to install Prometheus, run Akri with Prometheus support, access Grafana, and expose metrics from brokers.

The [rust prometheus crate](https://github.com/tikv/rust-prometheus) uses a more recent version of [rust protobuf](https://github.com/stepancheg/rust-protobuf), which [requires a more recent version of rust](https://github.com/stepancheg/rust-protobuf/issues/534). Therefore, this PR also increases our rust build version to the latest v1.49.0.

**If applicable**:


Cargo.lock
```diff
@@ -3181,6 +3562,34 @@ dependencies = [
  "try-lock 0.2.3",
 ]
 
+[[package]]
```
jiria, line 3565: Curious, with all the new dependencies, how much did we regress on footprint?

deployment/helm/templates/prometheus.yaml
```diff
@@ -0,0 +1,61 @@
...
+  labels:
+    release: prometheus
+spec:
+  selector:
+    matchLabels:
+      app: akri-controller-metrics
+  endpoints:
+  - port: metrics
+---
+{{- /* If Prometheus is enabled, create a service to target the metrics */}}
```
jiria, line 30: Could we use PodMonitor to not have to create separate services? Or is there a downside to that approach?

https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/design.md#podmonitor

docs/prometheus.md
```diff
@@ -0,0 +1,152 @@
...
+    kubectl get service/prometheus-grafana  --namespace=monitoring --output=jsonpath='{.spec.ports[?(@.name=="service")].nodePort}' && echo
+    ```
+1. SSH port forwarding can be used to access Grafana. Open a new terminal enter your ssh command to access the machine
+   running Akri and Prometheus followed by the port forwarding request. The following command will use port 50000 on the
+   host. Feel free to change it if it is not available. Be sure to replace `<Grafana Service port>` with the port number
+   outputted in the previous step.
+    ```sh
+    ssh someuser@<IP address> -L 50000:localhost:<Grafana Service port>
+    ```
+1. Navigate to `http://localhost:50000/` and enter Grafana's default username `admin` and password `prom-operator`. Now,
```
jiria, line 65: We might want to put some disclaimer somewhere to let folks know to change the password for security reasons.

jiria (approved):

docs/prometheus.md
```diff
@@ -0,0 +1,152 @@
...
+> guide](./user-guide.md) to determine whether the distribution you are using requires crictl path configuration.
+
+## Visualizing metrics with Grafana
+Now that Akri's metrics are being exposed to Prometheus, they can be visualized in Grafana. 
+1. Determine the port that the Grafana Service is running on, specifying the namespace if necessary, and save it for the
+   next step.
+    ```sh
+    kubectl get service/prometheus-grafana  --namespace=monitoring --output=jsonpath='{.spec.ports[?(@.name=="service")].nodePort}' && echo
+    ```
+1. SSH port forwarding can be used to access Grafana. Open a new terminal enter your ssh command to access the machine
```
kate-goldenring, line 58: "and"

deployment/helm/templates/prometheus.yaml
kate-goldenring: Thanks for pointing this out. This would be simpler. I'll try it out.

deployment/helm/values.yaml
```diff
@@ -88,7 +101,7 @@ debugEcho:
   - "foo1"
   # shared defines whether instances created as a result of
   # applying this debugEcho configuration as shared
-  shared: true
+  shared: false
```
bfjelds, line 104: Why is this included?

bfjelds:
This is so cool!!!

bfjelds (approved):

deployment/helm/values.yaml
kate-goldenring: Good catch. This was an accident. Will remove. Though i do think that `shared` in DebugEcho should default to `false`.

Cargo.lock
kate-goldenring: So sorry! I thought I replied to this:
Without Prometheus:
- agent: 24,448,848 
- controller: 17,427,432
- udev-broker: 9,873,408 
Each of the components increase by the following amount with Prometheus support added:
agent : 2,437,008
controller: 3,107,632
udev-video-broker: 2,525,352

PR #194, opened 2021-01-09
Allow overwriting controller.nodeSelectors

jiayihu:

Setting `nodeSelectors` last allows to overwrite `node-role.kubernetes.io/master: ""` to `node-role.kubernetes.io/master: "true"`, which is the actual value on k3s. Otherwise changing the label on the node might cause issues with other Pods expecting it to be the default value. It's can also be more troublesome to run `kubectl label node` if one has automated Akri deployment via Ansible like I do.

In the worst case, everything  is still working as before, so everyone should be happy 🤔

**If applicable**:


bfjelds:
I thunk this change is fine, but I think I'd document it differently.  If someone wants to override the existing `controller.onlyOnControlPlane' setting, I'd say they should specify `controller.onlyOnControlPlane=false` and the new nodesSelector values.

bfjelds:
Please run `version.sh -u -p` to update the versions for this PR.

jiayihu:
Yup that makes sense, although I feel it could be equally confusing having `controller.onlyOnControlPlane=false` and then a nodeSelector which states the contrary. In that case the chart change is also not needed because `node-role.kubernetes.io/master: ""` wouldn't be added in first place. Only the docs should be updated.

Btw the shell script `./version.sh -u -p` throws:

```
Updating to version: 0.1.6
Update ./shared/Cargo.toml [^version = .*] => [version = "0.1.6"]
sed: 1: "./shared/Cargo.toml": invalid command code .
```

kate-goldenring:
> sed: 1: "./shared/Cargo.toml": invalid command code .

What OS are you using? It looks like [different OSes (specifically OS X) expect an extension specifier](https://stackoverflow.com/questions/19456518/invalid-command-code-despite-escaping-periods-using-sed) with some sed commands, which we don't include. We may want to update our script to take in more parameters.

jiayihu:
And MacOS is indeed what I'm using 😄

kate-goldenring:
> And MacOS is indeed what I'm using 😄

Thanks for testing it out on a different OS. We wouldn't have found this otherwise. What happens if you change the two `sed -i` commands in `version.sh` to `sed -i '' -e ...`? For me, it works but also complains `sed: can't read : No such file or directory`.

docs/user-guide.md
```diff
@@ -96,9 +96,11 @@ helm install akri akri-helm-charts/akri
     controller.allowOnControlPlane=false` to your install command below. Conversely, if you only want the Controller to
     run on control plane nodes, add `--set controller.onlyOnControlPlane=true`. This will guarantee the Controller only
     runs on nodes with the label (key, value) of (`node-role.kubernetes.io/master`, ""), which is the default label for
-    the control plane node for Kubernetes. However, control plane nodes on MicroK8s and K3s do not have this label by
+    the control plane node for Kubernetes.
+    
+    However, control plane nodes on MicroK8s and K3s may not have this exact label by
     default, so you can add it by running `kubectl label node ${HOSTNAME,,} node-role.kubernetes.io/master=
-    --overwrite=true`. 
+    --overwrite=true`. Or alternatively, in K3s, you can set `controllers.nodeSelectors` to `"node-role.kubernetes.io/master": "true"` using a `values.yaml` file and passing `--set values.yaml` to the install command below. Any other `--set` flag mentioned so far can also be written in `values.yaml` as described in [Helm Value Files documentation](https://helm.sh/docs/chart_template_guide/values_files/).
```
kate-goldenring, line 103: Instead of passing a custom `values.yaml`, it may be simpler to document how the values can be set when installing via `--set`. The dots in the key [need to be escaped](https://stackoverflow.com/questions/49522776/helm-how-to-override-value-with-periods-in-name). For example, to enable the controller on the master of a K3s cluster:
```sh
   --set controller.nodeSelectors."node-role\.kubernetes\.io/master"=true
```
This can be tested locally and the manifest can be outputted by running:
```sh
helm template akri deployment/helm/ \
   --set controller.nodeSelectors."node-role\.kubernetes\.io/master"=true  \
   --debug
```

jiayihu: I think spent one hour trying to get how to property escape the inline flag. Btw I've updated the documentation accordingly to how you suggest.

I personally feed it's cleaner using `values.yml` because of the many non-trivial flags one has to pass but maybe it's easier to only use the CLI on the first run.

version.sh
```diff
@@ -46,7 +46,7 @@ update_file_version()
     NEW_LINE=$3
 
     echo "Update $FILE [$LINE_PATTERN] => [$NEW_LINE]"
-    sed -i s/"$LINE_PATTERN"/"$NEW_LINE"/g $FILE
+    sed -i ''  s/"$LINE_PATTERN"/"$NEW_LINE"/g $FILE
```
kate-goldenring, line 49: These sed changes cause it to not work on GNU. I think requesting that Mac OS users install GNU sed via `brew install gnu-sed` may be the best bet unless @bfjelds has any idea as to how to make the sed magic happen.

jiayihu: Okay let me try later with `gnu-sed`

jiayihu:
Okay I've tested with gnu-sed and it's working. I've reverted changes to `version.sh` and added instead a note for MacOS contributors. Let me know if there's anything else

kate-goldenring:
> Yup that makes sense, although I feel it could be equally confusing having `controller.onlyOnControlPlane=false` and then a nodeSelector which states the contrary. In that case the chart change is also not needed because `node-role.kubernetes.io/master: ""` wouldn't be added in first place. Only the docs should be updated.
> 
> Btw the shell script `./version.sh -u -p` throws:
> 
> ```
> Updating to version: 0.1.6
> Update ./shared/Cargo.toml [^version = .*] => [version = "0.1.6"]
> sed: 1: "./shared/Cargo.toml": invalid command code .
> ```

I still worry about the contradiction you mention here. The scenario where someone does the following is weird:
```sh
helm install akri akri-helm-charts/akri    --set controller.nodeSelectors."node-role\.kubernetes\.io/master"=true --set controller.onlyOnControlPlane=true
```

Should we get rid of `onlyOnControlPlane` and specify the node selector in-line per distribution? Or maybe we can set the value of `node-role.kubernetes.io/master:` to be parameterized something like the following (there may be a name more descriptive than `value` here):
```yaml
        {{- if .Values.controller.onlyOnControlPlane }}
        node-role.kubernetes.io/master: {{ .Values.controller.onlyOnControlPlane.value }}
        {{- end }}
```
Then, the install command for K3s (assuming a default value of ` .Values.controller.onlyOnControlPlane.value` to be `""`) would be:
```sh
helm install akri akri-helm-charts/akri    --set controller.onlyOnControlPlane=true --set controller.onlyOnControlPlane.value=true 
```

jiayihu:
You actually don't need to use both flags, since `onlyOnControlPlane` is false by default. So you can just use

```
helm install akri akri-helm-charts/akri    --set controller.nodeSelectors."node-role\.kubernetes\.io/master"=true
```

So you can decide whether to use only `--set onlyOnControlPlane=true` or `--set controller.nodeSelectors."node-role\.kubernetes\.io/master"=true` without contradictions

bfjelds:
what if `onlyOnControlPlane` was renamed to something like `useDefaultControlPlaneNodeSelector`?  then it wouldn't be logically confusing to specify useDefaultControlPlaneNodeSelector=false and `nodeSelector."node-role\.kubernetes\.io/master"=true`

jiayihu:
Another case of "Naming is hard". Btw one could argue that `useDefaultControlPlaneNodeSelector` leaves one wondering what's actually the default nodeSelector that will be applied 😄 The name is also extremely long.

But you know better, let me know what you prefer and I'll update the PR!

bfjelds:
> Another case of "Naming is hard". Btw one could argue that `useDefaultControlPlaneNodeSelector` leaves one wondering what's actually the default nodeSelector that will be applied 😄 The name is also extremely long.
> 
> But you know better, let me know what you prefer and I'll update the PR!

LOL.  So true on all counts.

How about:

values.yaml
```yaml
controller:
  controlPlaneSelectorValue=""
  useControlPlaneNodeSelector=false
```

controller.yaml:
```yaml
      {{- if or .Values.controller.linuxOnly .Values.controller.useControlPlaneNodeSelector.Values.controller.nodeSelectors }}
      nodeSelector:
        {{- if .Values.controller.nodeSelectors }}
          {{- toYaml .Values.controller.nodeSelectors | nindent 8 }}
        {{- end }}
        {{- if .Values.controller.linuxOnly }}
        "kubernetes.io/os": linux
        {{- end }}
        {{- if .Values.controller.useControlPlaneNodeSelector}}
        node-role.kubernetes.io/master: {{ .Values.controller.controlPlaneSelectorValue | quote }}
        {{- end }}
      {{- end }}
```

Then, you can specify `controlPlaneSelectorValue="true"` or whatever you like, can specify `useControlPlaneNodeSelector=false` or bring your own `nodeSelector`s?

jiayihu:
I would personally vote for leaving `onlyOnControlPlane `. From the perspective of a dev who has read the docs but doesn't know the Chart impl:

- If set `onlyOnControlPlane=true` it will apply `node-role.kubernetes.io/master: ""`
- I don't set it then the nodeSelector won't be applied. I can then apply my own one using `--set controller.nodeSelectors."node-role\.kubernetes\.io/master"=true`.

The point is that you don't actually need to pass `onlyOnControlPlane=false` because it's the default value. And since it's not passed it doesn't feel contradictory to me.

(Also it requires the less rework, leaving the PR as it is)

bfjelds (approved):

bfjelds:
I'm convinced :)

kate-goldenring (approved):
Thanks for making this a better experience for K3s users. After https://github.com/deislabs/akri/pull/202 merges, you'll need to rebase to get these tests passing. You will also need to either increase the version of this PR or https://github.com/deislabs/akri/pull/196/files, depending on which is the first to merge.

PR #196, opened 2021-01-10
Allow conditional compilation of agent protocols

jiayihu:

This is an initial draft to allow avoiding to compile all the deps when they are not actually needed by one's use case. This has two main advantages that comes to my mind:

1. The agent is more lightweight, avoiding to pack unnecessary deps in production
2. It allows to build and test the agent locally on non-linux OS, before deploying to the cluster. Otherwise I have to cross-build (which is much slower than native compilation), push the images, redeploy the chart (which involves pulling the image again).

The agent can't run locally on MacOS yet since it needs to be inside a k8s node, but this could be a first step towards that direction. I could maybe add a dumb agent that just calls the discovery handler in order to test the behaviour. In the worst case it's still helpful to have quick build errors instead of relying on `cross` for development.

Once the protocol handler is locally tested, one can deploy to the cluster for better testing. Otherwise I can't even write unit tests because the cargo can't compile the dependencies without `libudev` and in future there might be other unique env dependencies based on the protocol.

Let me know what you think about the idea 😄

**If applicable**:


agent/Cargo.toml
```diff
@@ -35,19 +34,23 @@ serde_json = "1.0.45"
 serde_yaml = "0.8.11"
 serde_derive = "1.0.104"
 akri-shared = { path = "../shared" }
-sxd-document = "0.3.0"
-sxd-xpath = "0.4.0"
 tempfile = "3.1.0"
```
jiayihu, line 37: I removed these deps and `mime` which didn't seem to be used anywhere

bfjelds (approved):

jiayihu:
@kate-goldenring Actually I think the PR is ready, I don't have anything else on my mind. The biggest annoyance is that one still needs to manually remove `"samples/brokers/udev-video-broker"` from the workspace `Cargo.toml`. Unfortunately there is no way to make it automatic, there is no concept of workspace level features yet.

Nevertheless I've been using the changes in the last days while developing the CoRE broker and I haven't noticed any issue. I also don't think that writing a dump agent to allow testing locally is necessary anymore, because of the transition to a gRPC/REST agent in #198

agent/src/protocols/mod.rs
```diff
@@ -80,13 +83,35 @@ fn inner_get_discovery_handler(
...
+        config => {
+            log::info!("No handler found for configuration {:?}, the default NoopDiscoveryHandler will be used", config);
+
+            Ok(Box::new(NoopDiscoveryHandler::default()))
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct NoopDiscoveryHandler {}
```
bfjelds, line 105: can this be removed?

agent/src/protocols/mod.rs
```diff
@@ -80,13 +83,35 @@ fn inner_get_discovery_handler(
...
         ProtocolHandler::onvif(onvif) => Ok(Box::new(onvif::OnvifDiscoveryHandler::new(&onvif))),
+        #[cfg(feature = "udevf")]
         ProtocolHandler::udev(udev) => Ok(Box::new(udev::UdevDiscoveryHandler::new(&udev))),
+        #[cfg(feature = "opcua")]
         ProtocolHandler::opcua(opcua) => Ok(Box::new(opcua::OpcuaDiscoveryHandler::new(&opcua))),
         ProtocolHandler::debugEcho(dbg) => match query.get_env_var("ENABLE_DEBUG_ECHO") {
             Ok(_) => Ok(Box::new(debug_echo::DebugEchoDiscoveryHandler::new(dbg))),
             _ => Err(failure::format_err!("No protocol configured")),
         },
+        config => {
```
bfjelds, line 96: i think we'd favor an error (for no configs being present) over having a noop config.

agent/Cargo.toml
```diff
@@ -35,19 +34,24 @@ serde_json = "1.0.45"
...
 
 [build-dependencies]
 tonic-build = "0.1.1"
+
+[features]
+default = ["onvif", "opcua", "udevf"]
+
+onvif = ["xml-rs", "yaserde", "yaserde_derive"]
+opcua = ["opcua-client"]
+udevf = ["pest", "pest_derive", "udev"]
```
bfjelds, line 57: does this need to be `udevf` or can it be `udev`? (not sure if the 'f' is intentional because of a naming conflict or if it is accidental)

agent/src/protocols/mod.rs
```diff
@@ -80,13 +83,35 @@ fn inner_get_discovery_handler(
...
+            log::info!("No handler found for configuration {:?}, the default NoopDiscoveryHandler will be used", config);
```
kate-goldenring, line 97: What was your thought process around adding this?

bfjelds (changes requested):
see comments around 'udevf' and 'noop'

kate-goldenring:
This is a great addition! 
You mention in the description:
>The agent can't run locally on MacOS yet since it needs to be inside a k8s node, but this could be a first step towards that direction. I could maybe add a dumb agent that just calls the discovery handler in order to test the behaviour. In the worst case it's still helpful to have quick build errors instead of relying on cross for development.

The [development doc](https://github.com/deislabs/akri/blob/main/docs/development.md#local-builds-and-tests) explains how to run the controller locally by passing it the kubeconfig file. You can also run the agent locally, by also specifying some additional runtime parameters needed for slot reconciliation. For example (for MicroK8s):
```sh
sudo RUST_LOG=info KUBECONFIG=~/.kube/config AGENT_NODE_NAME=kagold HOST_CRICTL_PATH=/usr/local/bin/crictl HOST_RUNTIME_ENDPOINT=/var/snap/microk8s/common/run/containerd.sock HOST_IMAGE_ENDPOINT==/var/snap/microk8s/common/run/containerd.sock ./target/debug/agent
```
> **Note**: the agent must run privileged for some protocols, such as udev
Then, add the configuration in another terminal:
```sh
kubectl apply -f configuration.yaml
```

This would be great to add to the documentation, but can also be added in another PR.

Also, currently the Agent is unix only due to the underlying rust UnixStreamer and UnixListener libraries used for the Device Plugins

jiayihu:
@bfjelds replaced noop with panic and renamed `udevf` to more meaningful `udevfeat`. It was indeed because of a conflict. Cargo already automatically creates a feature for every optional dependency, so `udev` can't be used to define what dependencies have to be included.

@kate-goldenring Sorry I wasn't clear in the description. I meant to run the agent locally without being in the cluster, although running in Microk8s would be a good compromise I guess. Personally I'd prefer to avoid for now. Every k8s distro has its own quirks because of different architectures, such as Kind vs Microk8s vs Minikube vs K3s. I'd prefer sticking with my k3s cluster which I'm getting to know better 😄

Btw IIRC Microk8s doesn't run on MacOS, it's Linux-only.

kate-goldenring:
> @bfjelds replaced noop with panic and renamed `udevf` to more meaningful `udevfeat`. It was indeed because of a conflict. Cargo already automatically creates a feature for every optional dependency, so `udev` can't be used to define what dependencies have to be included.
> 
> @kate-goldenring Sorry I wasn't clear in the description. I meant to run the agent locally without being in the cluster, although running in Microk8s would be a good compromise I guess. Personally I'd prefer to avoid for now. Every k8s distro has its own quirks because of different architectures, such as Kind vs Microk8s vs Minikube vs K3s. I'd prefer sticking with my k3s cluster which I'm getting to know better 😄
> 
> Btw IIRC Microk8s doesn't run on MacOS, it's Linux-only.

If you want to use K3s, I believe you should only need to substitute into the command above `KUBECONFIG=/etc/rancher/k3s/k3s.yaml` and `HOST_RUNTIME_ENDPOINT=/run/k3s/containerd/containerd.sock"`. But i understand the desire to not be tied to any cluster when testing.

agent/Cargo.toml
```diff
@@ -35,19 +34,24 @@ serde_json = "1.0.45"
...
 
 [build-dependencies]
 tonic-build = "0.1.1"
+
+[features]
+default = ["onvif", "opcua", "udevfeat"]
+
+onvif = ["xml-rs", "yaserde", "yaserde_derive"]
+opcua = ["opcua-client"]
+udevfeat = ["pest", "pest_derive", "udev"]
```
bfjelds, line 57: i haven't used features before, would it make sense to be consistent (even if it is not needed):

default = ["onvif-feature", "opcua-feature", "udev-feature"]
onvif-feature = ...
opcua-feature = ...
udev-feature = ...

jiayihu:
@kate-goldenring I can't do it because `/run/k3s/containerd/containerd.sock` is available only within the cluster. I guess I should be more precise when saying "locally". I develop on MacOS and deploy on a local k3s cluster running on 4 Rasp 3B+ node (actually 1 node is reserverd for [krustlet](https://github.com/deislabs/krustlet) :P)

agent/Cargo.toml
jiayihu: Sure, I changed to `-feat` suffix. Usually the feature names doesn't mention the word "feature", e.g. [smoltcp/Cargo.toml](https://github.com/smoltcp-rs/smoltcp/blob/master/Cargo.toml), but consistency >> convention 😄.

kate-goldenring (approved):
This is going to be a great addition! We are working on updating our workflow files so that the tests are no longer failing. When https://github.com/deislabs/akri/pull/202 merges, could you rebase? Then, hopefully the x's will disappear!

jiayihu:
@kate-goldenring this one is also ready to go 🎉

bfjelds (approved):

PR #197, opened 2021-01-11
Get Agent logs before testing their contents [SAME VERSION]

kate-goldenring:
End to end tests are failing on PRs stating that "Akri failed to successfully connect to crictl via the CRI socket"; however, the logs show crictl is successfully querying the CRI socket. It seems that `kubectl` is failing to get the logs. This PR attempts to grab the Agent's logs several times before testing the log's contents.

**If applicable**:

### DIFF
--- a/test/run-end-to-end.py
+++ b/test/run-end-to-end.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 import shared_test_code
-import json, os, time, yaml
+import json, os, subprocess, time, yaml
 from kubernetes import client, config
 from kubernetes.client.rest import ApiException
 
@@ -107,7 +107,7 @@ def do_test():
     temporary_agent_log_path = "/tmp/agent_log.txt"
     print("Check logs for Agent slot-reconciliation for pod {}".format(shared_test_code.agent_pod_name))
     for x in range(3):
-        log_result = os.system('sudo {} logs {} >> {}'.format(kubectl_cmd, shared_test_code.agent_pod_name, temporary_agent_log_path))
+        log_result = os.system('sudo {} logs {} > {}'.format(kubectl_cmd, shared_test_code.agent_pod_name, temporary_agent_log_path))
         if log_result == 0:
             break
         print("Failed to get logs from {} pod with result {} on attempt {} of 3".format(shared_test_code.agent_pod_name, log_result, x))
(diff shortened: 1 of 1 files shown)


test/run-end-to-end.py
```diff
@@ -104,10 +104,18 @@ def do_test():
...
+    for x in range(3):
+        log_result = os.system('sudo {} logs {} >> {}'.format(kubectl_cmd, shared_test_code.agent_pod_name, temporary_agent_log_path))
+        if log_result == 0:
+            break
+        print("Failed to get logs from {} pod with result {} on attempt {} of 3".format(shared_test_code.agent_pod_name, log_result, x))
+        if x == 3:
+            return False
+    grep_result = os.system('grep "get_node_slots - crictl called successfully" {} | wc -l | grep -v 0'.format(temporary_agent_log_path))
+    if grep_result != 0:
+        print("Akri failed to successfully connect to crictl via the CRI socket with stdout result of {}", grep_result)
```
bfjelds, line 118: maybe include a `os.system('sudo {} get pods,services,akric,akrii --show-labels'.format(kubectl_cmd))` here?

test/run-end-to-end.py
```diff
@@ -104,10 +104,18 @@ def do_test():
     #
     # Check that slot reconciliation is working on agent
     # 
+    temporary_agent_log_path = "/tmp/agent_log.txt"
     print("Check logs for Agent slot-reconciliation for pod {}".format(shared_test_code.agent_pod_name))
-    result = os.system('sudo {} logs {} | grep "get_node_slots - crictl called successfully" | wc -l | grep -v 0'.format(kubectl_cmd, shared_test_code.agent_pod_name))
-    if result != 0:
-        print("Akri failed to successfully connect to crictl via the CRI socket")
+    for x in range(3):
+        log_result = os.system('sudo {} logs {} >> {}'.format(kubectl_cmd, shared_test_code.agent_pod_name, temporary_agent_log_path))
```
bfjelds, line 110: maybe use `>` instead of `>>`

test/run-end-to-end.py
bfjelds: (so that if there is an error, we can see if the desired pod is running)

bfjelds: maybe we should call `os.system('grep "get_node_slots" {}'.format(temporary_agent_log_path))` in the error condition too, just to try to see what is going on when this fails.

kate-goldenring: From what i understand, `os.system` does not output anything. It just returns the return code. But we can use the [`subprocess` lib](https://stackoverflow.com/questions/4760215/running-shell-command-and-capturing-the-output) to output the results.

bfjelds: i think we see the output of the helm install command and it seems to use os.system

```
    helm_install_command = "helm install akri {} --set debugEcho.enabled=true --set debugEcho.name={} --set debugEcho.shared=false --set agent.allowDebugEcho=true {} {} --debug ".format(helm_chart_location, shared_test_code.DEBUG_ECHO_NAME, cri_args, extra_helm_args)
    print("Helm command: {}".format(helm_install_command))
    os.system(helm_install_command)
```

kate-goldenring: Was overthinking it. You're right. Thanks!

Issue #89, opened 2020-10-30
failure is deprecated


DazWilkin:
**Is your feature request related to a problem? Please describe.**

Rust `failure` crate is [deprecated](https://github.com/rust-lang-nursery/failure#failure---a-new-error-management-story)

**Describe the solution you'd like**

Consider replacing `failure`.

kate-goldenring:
Thanks for catching this, @DazWilkin.

PR #201, opened 2021-01-12
Use anyhow instead of the deprecated failure crate

romoh:
Use [anyhow](https://github.com/dtolnay/anyhow) for error handling instead of failure crate which is [deprecated](https://github.com/rust-lang-nursery/failure#failure---a-new-error-management-story). anyhow has parity with failure so the change is 1:1 and no major changes were required.
#89 

None

**If applicable**:
- [X] added code adheres to standard Rust formatting (`cargo fmt`)
- [X] inline documentation builds (`cargo doc`)
- [X] version has been updated appropriately (`./version.sh`)


kate-goldenring (approved):
Thanks for fixing this, @romoh! We are working on updating our workflow files so that the tests are no longer failing. When https://github.com/deislabs/akri/pull/202 merges, could you rebase? Then, (fingers crossed!) the tests will all be passing.

kate-goldenring:
> Can you run `version.sh -u -p` again?

PR train has been moving in reasonably quickly. @romoh, can you update the version until its greater than [main](https://github.com/deislabs/akri/blob/main/version.txt) currently 0.1.9?

agent/Cargo.toml
```diff
@@ -13,7 +13,7 @@ blake2 = "0.8.0"
 chrono = "0.4.10"
 cfg-if = "0.1"
 env_logger = "0.6.1"
-failure = "0.1.5"
+anyhow = "1.0.38"
```
jiria, line 16: nit: should we fix the ordering?

kate-goldenring: good catch, @jiria. we should. would you be willing to put in the fix?

Issue #198, opened 2021-01-11
Modularize Discovery Protocols as Pods 


kate-goldenring:
**Is your feature request related to a way you would like Akri extended? Please describe.**
Currently, protocol discovery is all implemented in the Akri Agent, which as explained in the [protocol extension proposal](https://github.com/deislabs/akri/blob/main/docs/proposals/simple-protocol-extension.md#background) makes the Agent bigger than needed (if someone is not using all the protocols), creates a larger attack surface, and requires discovery protocols to be written in rust.
Akri should be able to be deployed with only the desired discovery protocols, and discovery protocols should be able to be added without changing Akri's core components (the Agent, Controller, and Configuration CRD). This would make extending Akri easier. 

**Describe the solution you'd like**
This could be done by implementing idea 3 of the [protocol extension proposal](https://github.com/deislabs/akri/blob/main/docs/proposals/simple-protocol-extension.md#possible-ideas), wherein protocols are implemented as pods and exposed over a gRPC interface. The DiscoveryHandler's two methods `are_shared` and `discover` could be exposed over this interface. 

Additionally, the Configuration CRD would need to be modified so protocols are generic rather than explicitly defined in the CRD. Potentially as a map, as mentioned in [this thread](https://github.com/deislabs/akri/issues/132#issuecomment-748548127).

Also, the Agent should still be able to be compiled with DiscoveryHandlers in case users want the smaller footprint of one Agent pod rather than Agent pod + 1 pod for each DiscoveryHandler.

**Describe alternatives you've considered**
Other ideas listed in the [protocol extension proposal](https://github.com/deislabs/akri/blob/main/docs/proposals/simple-protocol-extension.md#possible-ideas), such as exposing the Agent as a library.

**Additional context**
The Zeroconf protocol #163  increases the Agent's size considerably and would benefit from running in its own Pod.

jiayihu:
I'll try to add some thoughts I have on my mind. I personally don't know gRPC, I've used only old RPC in my past so I don't have anything against it. I'd like just to point some benefits of REST over gRPC:

- I think it's easier for people working with REST/HTTP rather than gRPC
- The agent don't even need to actually run as pod inside the cluster. As far as agent and handler are able to communicate HTTP in some way, they can talk to each other. Not sure if gRPC can also do it, maybe yes since I think it has been designed at L7 level for that purpose.
[...]

PR #203, opened 2021-01-12
[Extensibility] CoAP CoRE

jiayihu:

This is a WIP PR which adds support to the CoAP CoRE protocol. It has at least reachable a POC point that I feel worth sharing.  I've added more information in the proposal doc. Sorry for any grammar error or type 😄 I'm going to check with Grammarly before merging.

**If applicable**:


agent/Cargo.toml
```diff
@@ -22,11 +24,10 @@ hyper = "0.11"
...
 k8s-openapi = { version = "0.6.0", features = ["v1_16"] }
 log = "0.4"
-mime = "0.3"
 mockall = "0.6.0"
-opcua-client = "0.7.0"
-pest = "2.0"
-pest_derive = "2.0"
+opcua-client = { version = "0.7.0", optional = true }
+pest = { version = "2.0", optional = true }
+pest_derive = { version = "2.0", optional = true }
```
jiayihu, line 30: These diffs should disappear when #196 is merged

agent/Cargo.toml
```diff
@@ -12,6 +12,8 @@ async-trait = "0.1.0"
 blake2 = "0.8.0"
 chrono = "0.4.10"
 cfg-if = "0.1"
+coap = { version = "0.9", optional = true }
+coap-lite = { version = "0.3", optional = true }
 env_logger = "0.6.1"
```
jiayihu, line 17: I've added support to CoRE parsing in `coap-lite` 0.3.4, just published today. But `coap` is explicitly using `0.3.2` so I think that duplicates the dependency. I'll open a PR in `coap`.

agent/src/protocols/coap_core/discovery_handler.rs
```diff
@@ -0,0 +1,145 @@
...
+            let endpoint = format!("coap://{}:5683/well-known/core", ip_address);
+            info!("Discovering resources on endpoint {}", endpoint);
+
+            // TODO: make timeout configurable
+            let response = CoAPClient::get_with_timeout(endpoint.as_str(), Duration::from_secs(5));
+
+            match response {
+                Ok(response) => {
+                    let payload = String::from_utf8(response.message.payload)
+                        .expect("Receive payload is not a string");
```
jiayihu, line 44: Ops a typo here

agent/src/protocols/coap_core/discovery_handler.rs
```diff
@@ -0,0 +1,145 @@
...
+    //             tokio::runtime::Runtime::new()
+    //                 .unwrap()
+    //                 .block_on(async move {
+    //                     let mut server = Server::new("127.0.0.1:5683").unwrap();
+
+    //                     server.run(request_handler).await.unwrap();
+    //                 })
+    //         })
+    //         .unwrap()
+    // }
```
jiayihu, line 106: I'm not sure how testing could be done. Something like [httptest](https://docs.rs/httptest/0.14.0/httptest/) would be nice, but it's easier said than done for CoAP. I have tried this implementation but didn't work. The server was not up when the test tries to connect later. Maybe the server is "dropped" as soon as the function returns?

I resorted to creating a server on another terminal tab to run the tests. I'm very open to any idea.

agent/src/protocols/coap_core/discovery_impl.rs
```diff
@@ -0,0 +1,73 @@
...
+            debug!("attr {} value {}", attr, value);
+
+            match attr {
+                "rt" => resource.rtype = Some(value),
+                "if" if value.as_str() == "sensor" => {
+                    // Only "sensor" interface is supported for now
+                    resource.interface = Some(value)
+                }
+                _ => {
+                    // Other attributes are not supported yet
```
jiayihu, line 39: I should probably panic here

deployment/helm/values.yaml
```diff
@@ -287,4 +287,52 @@ udev:
...
+    type: ClusterIP
+    # port is the service port of the instance service
+    port: 80
+    # targetPort is the service targetPort of the instance service
+    targetPort: 8083
+    # protocol is the service protocol of the instance service
+    protocol: TCP
+  # createConfigurationService is specified if a single service should automatically be
+  # created for all broker pods of a Configuration
+  createConfigurationService: true
```
jiayihu, line 327: This probably should be `false` by default, doesn't make a lot of sense to have a service of services in CoAP I think.

samples/brokers/coap-core-broker/Dockerfile.coap-core
```diff
@@ -0,0 +1,21 @@
...
+COPY ./build/container-images-legal-notice.md .
+
+RUN apt-get update && apt-get install -y --no-install-recommends libssl-dev openssl && apt-get clean
+COPY ./target/${CROSS_BUILD_TARGET}/release/coap-core-broker /coap-core-broker
+
+# Expose port used by broker service
+EXPOSE 8083
+
+ENV RUST_LOG coap_core_broker
+CMD ["./coap-core-broker"]
```
jiayihu, line 21: There are some inconsistencies on the Docker configuration side. I tried following the extensibility documentation, which didn't explain cross compilation for the broker however. So I looked at existing Docker configuration and there's a mix of things from the extensibility doc and things from the existing configs.

E.g. this file should be in `/build/containers`

bfjelds: This is definitely a shortcoming of the extensibility docs.

I would suggest updating build\akri-containers.mk to include your new broker container.   Then create something like .github\workflows\build-udev-video-broker-container.yml for your broker.

docs/proposals/coap-core.md
```diff
@@ -0,0 +1,74 @@
...
+
+1. Akri agent sends a `GET /well-known/core` request to the device at `coap://{IP_ADDRESS}:5683`. The standard defines that CoAP devices which mean to support resource discovery, must be reachable with the default port `:5683` and expose endpoint `/well-known/core` implementing the CoRE Link Format.
+2. The device responds with the list of supported resources. An example of response in link format is the following:
+    ```
+    </sensors/temp>;rt="oic.r.temperature";if="sensor",
+    </sensors/light>;rt="oic.r.light.brightness";if="sensor"
+    ```
+
+    The example is stating that the device has 2 REST resources, `/sensors/temp` and `/sensors/light` which are of type `oic.r.temperature` and `oic.r.light.brightness` respectively. `rt` values are defined in [IANA](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#rt-link-target-att-value) to have same standardization, although vendor values can probably be used. Then `if` means that the interface description of the resource is of type `sensor`. Currently only resources with interface `sensor` are supported by the agent.
```
kate-goldenring, line 34: Do you see Akri supporting [other interfaces](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#if-link-target-att-value), such as the actuator one?

docs/proposals/coap-core.md
```diff
@@ -0,0 +1,74 @@
+# [CoAPCore](https://tools.ietf.org/html/rfc6690#:~:text=well-known%2Fcore) Protocol Implementation
```
kate-goldenring, line 1: this is nit but also a helpful clarification to me as I learn about CoRE, but is this better named CoAP CoRE?

deployment/helm/crds/akri-configuration-crd.yaml
```diff
@@ -100,11 +100,20 @@ spec:
                               type: array
                               items:
                                 type: string
+                    coapcore: # {{CoAPCoREDiscoveryHandler}}
+                      type: object
+                      properties:
+                        ipAddresses: # {{DeviceIPs}}
+                          type: array
```
kate-goldenring, line 107: Maybe add some filtering options so people can specify that they only want to discover devices that support certain [resource types](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml) -- say "I only want to find printers".

deployment/helm/templates/coap-core.yaml
```diff
@@ -0,0 +1,57 @@
...
+kind: Configuration
+metadata:
+  name: {{ .Values.coapcore.name }}
+spec:
+  protocol:
+    coapcore:
+      {{- if .Values.coapcore.ipAddresses}}
+      ipAddresses: 
+        {{- toYaml .Values.coapcore.ipAddresses | nindent 8 }}
+      {{- else }}
```
kate-goldenring, line 12: I dont think this is necessary since `ipAddresses` defaults to `[]` if nothing is specified

docs/proposals/coap-core.md
```diff
@@ -0,0 +1,74 @@
...
+
+  The Broker acts as a HTTP to CoAP Proxy. It translates RESTful HTTP requests into RESTful CoAP requests and viceversa for the response. "Cross-Protocol Proxying between CoAP and HTTP" is defined in section 10 of RFC 7252. Currently, the Broker forwards only GET requests.
+  
+  It's also in a good position to cache CoAP responses. "Unlike HTTP, the cacheability of CoAP responses does not depend on the request method, but it depends on the Response Code." Currently, the Broker caches 2.05 Content responses, which are returned if the device is not reachable during the connection.
+
+## Outstanding Questions
+
+- Is there a better way to store the discovered resources as Configuration in the cluster?
+
+I think the current implementation would need a controller to accept queries about available resources and return the broker which can communicate to the device. The device is listed as generic `akri.sh/coap-core-021dd7` resource on the node, which is pretty much useless. A better label would be `akri.sh/oic.r.temperature-021dd7`, that is the discovered resource. This would allow using the K8s controller for scheduling pods which need the resource.
```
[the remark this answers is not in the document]
jiayihu: > are you thinking of adding support for filter in/out devices based on resource type, similar to Zeroconf

I can surely do it if there some use cases for it 🤔

About the rest of the point, I think that filtering to allow only a low fixed numbers of resource types could be limiting. If I were a platform like Azure, who wants to offer edge clusters with IoT devices as resources, I'd prefer supporting a large diverisity of resource types because that's the nature of IoT. 
I think filtering makes sense if the use case is a company is specialized in one kind of devices.

One of the ideas of my thesis, is building a platform which can offer local devices as services in edge computing. Just like today you can provision databases and other software as service in the cloud. But even in a small areas like a neighbourhood, you could have a large number of resource types. Also by filtering, that would mean that you would have two separate Configurations for temperature and brightness and thus two brokers. Having two Configurations doesn't look bad per se, whereas two brokers for each resource in each device feels more wrong. That number `brokers x types x devices` would be explosive I believe.

docs/proposals/coap-core.md
```diff
@@ -0,0 +1,74 @@
...
+
+Resource Discovery can be performed either unicast or multicast.
+
+When a server’s IP address is already known, either a priori or resolved via the Domain Name System (DNS), unicast discovery is performed in order to locate the entry point to the resource of interest. In this specification, this is performed using a GET to `/.well-known/core` on the server, which returns a payload in the CoRE Link Format.
+
+Multicast Resource Discovery is useful when a client needs to locate a resource within a limited scope, and that scope supports IP multicast. A GET request to the appropriate multicast address is made for `/.well-known/core`. In order to limit the number and size of responses, a query string is recommended with the known attributes. Typically, a resource would be discovered based on its Resource Type and/or Interface Description, along with possible application-specific attributes.
+
+### With Akri
+
+The Akri Configuration defines a list of IP addresses to use for resource discovery, implementing thus unicast discovery. Adding support to multicast discovery is surely desiderable.
```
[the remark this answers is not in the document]
jiayihu: A CoAP broadcast request is just an IP multicast, the increase of difficulty lies in handling responses from multiple devices but it shouldn't be hard. I already planned to consider this feature as the highest priority. This would allow discovery of every device within the same network.

Unfortunately I'll have less time to work on akri in the next days because I'm moving to other tasks in my thesis, but I'll give a try asap :)

docs/proposals/coap-core.md
jiayihu: Do you mean if there's a better name than CoAP CoRE? Officially the discovery protocol is CoRE, but the implementation is tailored for CoRE under CoAP. As the RFC6690 states, CoRE can be applied also to other protocols like HTTP, so it could be named only CoRE.

For the time being I chose CoAP CoRE to be explicit about the expected communication protocol used by the devices. But I can imagine in future that the same discovery handler could support also HTTP or even MQTT etc.

agent/src/protocols/coap_core/discovery_handler.rs
bfjelds: the most complete way that comes to mind would be to wrap `CoAPClient::get_with_timeout` in a trait and create a mock version of it.  i would then create an internal version of discover that accepts the trait as an argument.  this would allow you to test all of the code (what get_with_timeout expects as arguments, how the code in discover reacts to various return values, etc).

bfjelds: `MockOnvifQueryImpl` is an example of that

agent/src/protocols/coap_core/discovery_impl.rs
bfjelds: we favor passing errors up through the return (rather than panic)

agent/src/protocols/mod.rs
```diff
@@ -80,13 +85,39 @@ fn inner_get_discovery_handler(
...
+        #[cfg(feature = "coapcore")]
+        ProtocolHandler::coapcore(coapcore) => Ok(Box::new(
+            coap_core::CoAPCoREDiscoveryHandler::new(&coapcore),
+        )),
         ProtocolHandler::debugEcho(dbg) => match query.get_env_var("ENABLE_DEBUG_ECHO") {
             Ok(_) => Ok(Box::new(debug_echo::DebugEchoDiscoveryHandler::new(dbg))),
             _ => Err(failure::format_err!("No protocol configured")),
         },
+        config => {
+            log::info!("No handler found for configuration {:?}, the default NoopDiscoveryHandler will be used", config);
```
bfjelds: when you rebase this code, can you replace the `panic!("No handler found for configuration {:?}", config);` that is in here with `Err(failure::format_err!(No handler found for configuration {:?}", config))`

docs/proposals/coap-core.md
jiayihu: I'd personally leave this for someone else for the time being. Maybe someone else will have an use case for it in future and will contribute :)

deployment/helm/values.yaml
bfjelds: it seems like that would depend on how the broker is structuring the data it is sending.  if it just sends a number without context (72, 34), then i would agree.  but if there is some context (temp=72, humidity=34), then it seems fine to have config service.  i haven't fully wrapped my head around what this protocol means yet, but if a single config might discover many instances that are all the same type (5 instances for 5 thermometers), then the config service would allow someone to view all of them.

docs/proposals/coap-core.md
```diff
@@ -0,0 +1,74 @@
...
+2. The device responds with the list of supported resources. An example of response in link format is the following:
+    ```
+    </sensors/temp>;rt="oic.r.temperature";if="sensor",
+    </sensors/light>;rt="oic.r.light.brightness";if="sensor"
+    ```
+
+    The example is stating that the device has 2 REST resources, `/sensors/temp` and `/sensors/light` which are of type `oic.r.temperature` and `oic.r.light.brightness` respectively. `rt` values are defined in [IANA](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#rt-link-target-att-value) to have same standardization, although vendor values can probably be used. Then `if` means that the interface description of the resource is of type `sensor`. Currently only resources with interface `sensor` are supported by the agent.
+3. For each device, the agent returns a `DiscoveryResult` which will have the following properties:
```
bfjelds, line 35: i'm having mixed feelings about how Akri should represent the discovery results.

the proposal seems to be that there is a 1:1 relationship between Instance and IP.  (in the example below, there is an Instance corresponding to 192.168.1.126)

i wonder if there shouldn't be an Instance for each IP+sensor.  (in the example below, there would be 2 Instances generated, 1 for 192.168.1.126+brightness and 1 for 192.168.1.126+temp)

i suspect that there are pros and cons to each ... i'd be interested in hearing why you decided on what you did.

bfjelds: maybe this is a duplicate of https://github.com/deislabs/akri/pull/203/files#r556583235

docs/proposals/coap-core.md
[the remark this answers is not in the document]
kate-goldenring: I think the main question I have is whether each resource type should be returned as a separate `DiscoveryResult` [here](https://github.com/jiayihu/akri/blob/6aee1161d6eccdbc5a80051f38560f6b24692bbe/agent/src/protocols/coap_core/discovery_handler.rs#L49) and therefore gets a separate Instance (and could also get a separate broker). Currently, we have one broker deployment strategy of one broker per instance, but this could change in the future to where someone could request a core_coap temperature and light sensor per broker, but this does get weird, since that could result in getting a temperature sensor from one endpoint and a light sensor from another instead of the temperature and light sensors from the same endpoint. It would also be hard to manage device usage if multiple resources of the device are being exposed as instances.

I am not familiar with the use cases for coap_core, so I differ to you. The more we discuss it, I am falling in line with your thoughts of each service being a `DiscoveryResult` rather than breaking it down further by resource type. Thanks for being patient with me and helping me type through my thoughts.

As for filtering, I still think it could be useful. People might only want to discover services that expose a certain resource type, since their brokers may specifically be interested in that resource type. Say you have a broker that grabs the temperature reading from temperature sensors and serves it to some anomaly detection app, you'd want to ensure that the service/device it is grabbing the temperature from does in fact have a temperature resource. Then, in the Configuration, you could specify the filter and `discover` would simply only return devices that have that resource type.

Also, your thesis sounds cool. Do you intend to use existing tech, such as CoAP to expose the local devices as services? Do you intend to just use CoAP or discover the devices via multiple protocols?

docs/proposals/coap-core.md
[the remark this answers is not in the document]
kate-goldenring: Awesome! And no worries. We will still be here.

docs/proposals/coap-core.md
kate-goldenring: Good point. This could definitely come in a later iteration.

docs/proposals/coap-core.md
kate-goldenring: Cool. I was specifically referring to changing Core to CoRE, but this has reinforced the fact that I need to learn more about CoRE and how it can be applied to other protocols.

agent/src/protocols/coap_core/discovery_handler.rs
jiayihu: Thanks for the tip, I followed it

docs/proposals/coap-core.md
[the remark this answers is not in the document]
jiayihu: > someone could request a core_coap temperature and light sensor per broker, but this does get weird, since that could result in getting a temperature sensor from one endpoint and a light sensor from another instead of the temperature and light sensors from the same endpoint

Actually I don't think that becomes necessary weird. It would allow to virtualize the IoT resources. If you are asking for temperature and brightness, one could not care from which device it comes from. That would allow some virtualization benefits. For instance, we could use the brightness from another device if it fails in the current one. Since we're talking about edge, one may probably want to restrict the location of the resources though (e.g. "temperature and brightness in this room"). And if you restrict enough, the limit case would having the two resources from the same device. It's just an idea, maybe silly.

What I mean is that I think it depends on the use case. In other cases, I agree it does get weirds, e.g. if I need both camera recording and infrared recording from the same device.

The interesting point that has come to my mind while typing is that maybe we should discuss about what virtualization means for akri in the context of edge devices. Currently, akri allows to add device instances as resources and it's thus a virtualization of device instances. However, virtualizing resources instead of devices seems to be more natural to me if we think about the cloud.

For instance, let's be guided by how we expect users to use Akri instances. An application would never request a CoAP CoRE device, because that doesn't make sense. It would more likely happier to request a resource type, like temperature. How do we implement that with Akri? If we have an instance/broker per device, then we need to write a new controller which scans the list of device instances to find a result with the temperature sensor as metadata. This doesn't feel very natural to me. 

Is an instance per resource better? Maybe. Is also a broker per resource instance better? I don't think so because it would be an explosive overhead.

I think "the main question I have is whether each resource type should be returned as a separate DiscoveryResult" can be reduced further to what virtualization means for Akri and, even before that, what's the expected user experience for application developers. How will developers use CoRE resources in your opinion?

Do these questions make sense for the other protocols? Unfortunately I have no experience with the other protocols, although I saw OPC UA mentioned in some papers. My feeling is that maybe these questions haven't occurred before because many protocols assume what the underlying resource type is. E.g. an onvif instance is the same as a camera instance and thus requesting an onvif instance to Akri is fine. But if we think about it, the discovery protocol should be an implementation detail. If you need a camera recording, the cluster should be able to provision it independently from the protocol. If I have an OVIF camera, a udev camera and a CoAP CoRE camera, the three of them should be usable by the application. And this is also why my CoAP broker acts as a proxy converting between HTTP to CoAP and viceversa. Any application can request the CoAP resources without knowing about CoAP.

> Also, your thesis sounds cool. Do you intend to use existing tech, such as CoAP to expose the local devices as services? Do you intend to just use CoAP or discover the devices via multiple protocols?

Thanks :) Yes, I instead to expose devices as services and I thought that it's also the goal of Akri. I read a nice paper the other day, which explains the same vision: [Combining Cloud and sensors in a smart city environment](Combining Cloud and sensors in a smart city environment). It was written in 2012 when Cloud wsa becoming popular. I have a broader vision which merges cloud, fog and edge computing but the main ideas is the same. If you have some time, give it a read. I think it's worth it, but I'm biased of course 😄

docs/proposals/coap-core.md
jiayihu: I'll correct the name to `CoAP CoRE` in the proposal title 👍

docs/proposals/coap-core.md
jiayihu: Yes I feel the same and we are discussing it here: https://github.com/deislabs/akri/pull/203#discussion_r560601025 

I'll resolve this conversation, feel free to reopen if I was mistaken in understanding.

docs/proposals/coap-core.md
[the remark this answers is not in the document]
kate-goldenring: I think you are right that these questions are more unique to CoRE. 
> How will developers use CoRE resources in your opinion?

I am certainly not well versed in CoRE. What are some of the main use cases you've seen?

> If you are asking for temperature and brightness, one could not care from which device it comes from. That would allow some virtualization benefits. For instance, we could use the brightness from another device if it fails in the current one. 

You make a good point here about trade offs. 

> If we have an instance/broker per device, then we need to write a new controller which scans the list of device instances to find a result with the temperature sensor as metadata.

This could be an issue even if we have instance per resource. For example, say the Configuration was to find and expose all brightness and temperature resources (maybe a filter of brightness || temperature). Instances would be created for each brightness and each temperature resource, but the name of those instances would be the same: <configuration-name>-digest. So how would Pods request a specific resource? There would need to be a convention of creating separate Configurations for each resource type. 

Maybe we should make the way the Agent names Instances more customizable, such that every instance of a Configuration does not have the same name (besides a digest id). The device plugin name == instance name, so if the instance name is more specific then a resource could be requested with a more specific name. Maybe in the properties map of a DiscoveryResult, there could be a "INSTANCE_NAME" key the Agent could check for. If it exists, the Agent names the instance with the value. Otherwise it gets the default name.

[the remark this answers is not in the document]
jiayihu: > I am certainly not well versed in CoRE. What are some of the main use cases you've seen?

In practise not very much since most works in embedded are close-sourced. But the literature use cases are of any IoT use case since the purpose of CoAP is to be a low-overhead, bandwidth and energy communication protocol yet allowing meaningful RESTful communication. E.g. RFID sensors and healthcare devices are two quite different use-cases I've read about.

One important feature of CoRE is being thought for M2M communication (compared e.g. to udev). That's why it's important for me to support provisioning by resource type, because it requires the least human works. I believe that keeping in mind M2M communication is fundamental when developing for IoT, but it's just my opinion.

In my use case, the cluster must be able to automatically discover resources, assign them to applications or handle availability in case of failure. The applications themselves must be able to work in any cluster with heterogeneous resources so they can only have expectations about the available resource types, nothing about a specific instance. Ideally people develop IoT applications locally and then deploy on a huge variety of edge clusters where Akri instances are running. That will truly bring the cloud to the edge.

If you think about it, in the current industry, a lot of manual repetitive work is done when dealing with IoT. For instance, flashing each embedded device is a nightmare and that's why deployment is always a pain. In the cloud, we can nowadays just push an image and it's done.

I hope this makes sense. I'm open to any criticism since all of you work on Azure and you know better than me.

> Instances would be created for each brightness and each temperature resource, but the name of those instances would be the same: -digest. So how would Pods request a specific resource? There would need to be a convention of creating separate Configurations for each resource type.

Did you mean that the name of those instance would be the same despite being of two different types? And thus how would a Pod request a specific **resource type**. Just clarifying since it's quite easy to get confused with all these similar terms.

> There would need to be a convention of creating separate Configurations for each resource type.

Yup, I thought a bit about it in https://github.com/deislabs/akri/issues/198#issuecomment-758291583. The issue is the potentially huge number of resource types. But maybe in the end it's the best compromise if we suppose that a cluster owner will limit the resource types to a restricted small group.

So far the choices for Configuration are: 1 Configuration x discovery protocol, 1 Configuration x device, 1 Configuration x resource, 1 Configuration per resource type. 

> Maybe we should make the way the Agent names Instances more customizable, such that every instance of a Configuration does not have the same name (besides a digest id). 

I think I understood what you mean, but I'll ask you to provide an example just to be sure 😅

> The device plugin name == instance name, so if the instance name is more specific then a resource could be requested with a more specific name.

That depends on what you mean with "more specific name". Because of the aforementioned M2M goal, a Pod should always just request by resource type like IANA `oic.r.temperature` and never written to need a specific instance like `temperature-digest` IMO.

docs/proposals/coap-core.md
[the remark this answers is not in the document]
jiayihu: I've added support to multicast today 🎉

agent/src/protocols/coap_core/discovery_impl.rs
jiayihu: My initial comment was wrong. After reading the code again, I think it should be left empty. The rational is that a device can list all of its resources and Akri only picks up the ones it supports. Devices are unaware of Akri and should not forced to limit themselves.

deployment/helm/values.yaml
jiayihu: We are discussing what a config means for CoAP. I'll leave as it is now then, for the time being.

deployment/helm/crds/akri-configuration-crd.yaml
jiayihu: I've just added support to query filtering https://tools.ietf.org/html/rfc6690#section-4.1. This will allow to filter discovered resources by setting:

```yml
queryFilter: 
  name: rt
  value: oic.r.temperature
```

From my understanding of the standard, only 1 filter is allowed and the value itself is extremely limited. That's the rational for having a filter as fixed struct instead of list of filters. Devices are also not required to support filtering because of the constrained environment, so some filtering is done by the agent.

agent/Cargo.toml
```diff
@@ -12,6 +12,8 @@ async-trait = "0.1.0"
 blake2 = "0.8.0"
 chrono = "0.4.10"
 cfg-if = "0.1"
+coap = { git = "https://github.com/jiayihu/coap-rs", branch = "receive_from", optional = true }
```
jiayihu, line 15: I'll remove my own fork when the PR https://github.com/Covertness/coap-rs/pull/60 is merged

agent/src/protocols/coap_core/discovery_handler.rs
```diff
@@ -0,0 +1,416 @@
...
+
+    fn configure_unicast_response(mock: &mut MockCoAPClient, timeout: Duration) {
+        mock.expect_get_with_timeout()
+            .withf(move |_url, tm| *tm == timeout)
+            .returning(|_url, _timeout| Ok(create_core_response()));
+    }
+
+    #[tokio::test]
+    async fn test_discover_resources_via_ip_addresses() {
+        // TODO: find better way than setting env variable, which could be shared to other tests
```
jiayihu, line 237: The issue lies in https://github.com/deislabs/akri/blob/1c81b6ca44cccb596f0b19a01a220d96154b7658/agent/src/protocols/mod.rs#L24

which needs to read the env variable. It could receive a trait as parameter, but it feels a bit hacky because it should be passed down all the way from the tested method so that we can pass a mock. I'm open tu suggestions.

samples/brokers/coap-core-broker/Dockerfile.coap-core
jiayihu: Thanks for the help, I've tried to create the CI config. Let's see if it works

jiayihu:
So @kate-goldenring @bfjelds , I think this is ready for some serious review for the first iteration. This PR currently supports:

- CoAP Core basic use cases
- Discovery via multicast and static IP addresses
- Query filtering to limit accepted devices
- Broker which acts as proxy translating HTTP-CoAP GET requests and caching responses

Maybe what's missing is:
- _Moar_ tests 😄
- Some documentation about how one can test this stuff, maybe in a section of the proposal markdown.

Let me know if there's anything else

docs/proposals/coap-core.md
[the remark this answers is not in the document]
kate-goldenring: Thanks for sharing the use cases.

> Did you mean that the name of those instance would be the same despite being of two different types? And thus how would a Pod request a specific resource type. 

Yup.

> Maybe we should make the way the Agent names Instances more customizable, such that every instance of a Configuration does not have the same name (besides a digest id).

A DiscoveryHandler (or a call to discover) could return the name that instance should get. So if each resource type is a discovery result, oic.r.temperature could be the returned instance name. That would still have a digest put after it, so ...

>Because of the aforementioned M2M goal, a Pod should always just request by resource type like IANA oic.r.temperature and never written to need a specific instance like temperature-digest IMO.

With Akri, we definitely want to support requesting resources on the Configuration level. so you could request `udev-video` instead of `udev-video-digest`. This brings up an interesting point though: if it isnt 1 Configuration x resource type, that solution would not solve your need to request `oic.r.temperature`

> If you think about it, in the current industry, a lot of manual repetitive work is done when dealing with IoT. For instance, flashing each embedded device is a nightmare and that's why deployment is always a pain. In the cloud, we can nowadays just push an image and it's done.

Definitely agree that there needs to be better device management and hopefully Akri could provide a path to that

deployment/helm/crds/akri-configuration-crd.yaml
kate-goldenring: Where in the Configuration would people pass their requested filters? Instead of using queryFilter and having to handle the case that devices might not support it, why not just filter the results in the DiscoveryHandler? So if the DiscoveryHandler discovers 3 devices [device1: temperature, brightness, and moisture sensors/resources; device2: moisture; device3: brightness], and the configuration requests moisture resources, then the DiscoveryHandler would filter out device3 and return the other 2.

docs/proposals/coap-core.md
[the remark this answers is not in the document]
jiayihu: > This brings up an interesting point though: if it isnt 1 Configuration x resource type, that solution would not solve your need to request oic.r.temperature

Exactly, this is the reason I started to question about the current 1 Configuration x discovery protocol and the consequent additional controller which looks for instances with requested metadata if things are kept as they are.

deployment/helm/crds/akri-configuration-crd.yaml
jiayihu: I'm currently passing `queryFilters` in the properties:

```yaml
properties:
  multicast: true
  multicastIpAddress: 224.0.1.187
  staticIpAddresses: []
  queryFilter:
    name: rt
    value: oic.r.temperature
  discoveryTimeoutSeconds: 5
```

> Instead of using queryFilter and having to handle the case that devices might not support it, why not just filter the results in the DiscoveryHandler

I'm already filtering the results nevertheless in the DiscoveryHandler, exactly because devices are not required to support filtering :) However, I preferred to support only the standard query filtering instead of some custom Akri filtering. This limits filtering because, as far as I understand the RFC, you can filter only one property and only one value. So you **cannot** write a filter like `rt=temperature,brightness`, only `rt=temperature` or `rt=brightness` or `rt=temp*`. Actually I haven't added support to wildcart yet because that would require an additional dependency to a regex crate, which I discovered is not in std in Rust. But I don't have anything against it.

The rational for supporting the standard filtering, although limited, is that it feels more suitable for a first iteration to keep things simple. When more uses will emerge, more custom configuration can be added. Otherwise, for the time being, it could feel counter-intuitive having custom filter rules because one might expect that they are transformed to a proper CoRE query but that cannot actually happen. Currently, I can instead happily see that `/well-known/core?rt=oic.r.temperature` is passed to my CoAP servers and the DiscoveryHandler filters the results anyway if the server ignores it. This is standard expected behaviour.

I can't deny I'm also biased because my first attempt was allowing a custom rule like `rt=temperature,brightness` but things can become quickly weird, both from user and Akri perspective. From the user developer perspective, one wonders how this is implemented since it's non-standard. The same doubt happens with any other kind of query structure. From Akri perspective, it become ugly to do query parsing and to do conversion between CoAP query and our custom query. For instance, if we allow a custom filter to allows only `temperature` and `brightness`, what would the CoAP request look like?

 - `/well-known/core?rt=oic.r.temperature` is wrong because it's missing the brightness type
 - `/well-known/core?rt=oic.r.brightness` has the reverse issue
 - `/well-known/core?rt=oic.r.temperature,oic.r.brightness` is not standard-compliant
 - `/well-known/core?rt=oic.r.temperature` and `/well-known/core?rt=oic.r.brightness`? That seems the only doable compromise, at the cost of duplicating each discovery request and increasing the overhead also on the edge device.

Therefore, in order to avoid complicating stuff without a clear use case for it, I think supporting the limited standard query filtering is better for the time being. There could be also the case that a future version of the CoAP standard allows slightly more complex queries and by sticking to the standard, we save a lot of headaches. What do you think?

kate-goldenring: Starting with standard query filtering and then potentially doing custom filtering as a vnext sounds good. Maybe add that information to the proposal so we can keep track of it.

agent/src/protocols/coap_core/discovery_handler.rs
kate-goldenring: Hi @jiayihu, we set up a mock `EnvVarQuery` struct that can be used in tests. Here is an example of a test that uses it and of breaking a function into `fn_name` and `inner_fn_name` to create a testable `inner_fn_name` that tests in the mock struct: https://sourcegraph.com/github.com/deislabs/akri/-/blob/agent/src/protocols/mod.rs#L121

jiayihu: Thanks for the tip! I have been using `EnvVarQuery` already in some tests, but this case is different. I've shown the problematic line of code in my previous comment, which is not written by me actually. So I would need to pass `EnvVarQuery` down to all the functions until the one actually reading the env variable, since it's an internal method of akri. That would make the API a bit awkward I guess.

I believe there's maybe a better way to write this Akri method, thus why I asked for input from you:
https://github.com/deislabs/akri/blob/1c81b6ca44cccb596f0b19a01a220d96154b7658/agent/src/protocols/mod.rs#L16-L37

kate-goldenring:
Hi @jiayihu! The new `DiscoveryHandler` model has been merged. Its a big change, so I put together [a document](https://hackmd.io/@akri/r1NjkIFQO) with the steps to get this PR into the new mold. Let me know how I can help. Once you are rebased -- this is going to be tricking and I wonder if a new PR is better --, I could put PR's into yours with the nitty stuff like a Helm template for the Configuration.

jiayihu:
Mmh I'll give it a try during the weekend maybe. I don't see many conflicts so I hope rebasing won't be a big issue. I'll ask for help in case :)

PR #207, opened 2021-01-15
Image pull policy

kate-goldenring:
Currently, the [`PullPolicy`](https://kubernetes.io/docs/concepts/containers/images/#updating-images) on our container images is `Always`. This means that if Akri's component containers continue to error out and retry due to some issue, the container images will continuously be pulled from our container registry. 

Due to the high number of pulls on our `udev-video-broker image`, the following scenario is probably happening: fake video devices are being created (as described in our [end-to-end-demo](https://github.com/deislabs/akri/blob/main/docs/end-to-end-demo.md)), but no stream is being passed through them, causing the brokers Pods to fail and Kubernetes to continually restart them. A new image should not be pulled each time in this scenario.

**If applicable**:


deployment/helm/values.yaml
```diff
@@ -46,7 +46,7 @@ controller:
     # controller.yaml will default to v(AppVersion)-dev
     tag:
     # pullPolicy is the Akri Controller pull policy
-    pullPolicy: Always
+    pullPolicy: IfNotPresent
```
jiria, line 49: Should we set this only if we do not have useLatestContainers? Otherwise folks would get stuck, no? Without useLatestContainers, that should be fine, because they are picking a specific release and we are not planning to update the tags (I think). What do you think?

kate-goldenring: Is Kubernetes not able to check whether the image digest has changed when using `ifNotPresent`? The [docs](https://kubernetes.io/docs/concepts/configuration/overview/#container-images) says you can specify a digest to make sure you always use the same version of an image. That makes me think that it could distinguish whether a "newer" latest container exists? Maybe that is too hopeful, though.

Just to get all the possibilities down on paper, if I am not mistaken, @bfjelds please correct me if I am wrong, there are 4 main containers that could be pulled (if not specifying an explicit tag). Say that the akri Helm chart was added (`helm repo add akri-helm-charts https://deislabs.github.io/akri/`) when the version was 0.1.5:
1. `akri-dev` chart: <image>:**v0.1.5-dev**
2. `akri-dev` chart with `useLatestContainers`: <image>:**latest-dev** (this could be greater than v0.1.5 if more versions have been created)
3. `akri` chart: <image>:**v0.1.5**
4. `akri` chart with `useLatestContainers`: <image>:**latest** (this could be greater than v0.1.5 if more versions have been created)

The user would need to do `helm repo update` to remove the incongruency between latest and non-latest. 

So, for 1 and 3, what you're saying makes sense, most likely those havent been updated.

jiria: Iirc, without specifying the pullPolicy, if the tag is latest, it defaults to Always, otherwise IfNotPresent. You can override by specifying the pullPolicy. That makes me think that IfNotPresent will not refresh. Though it should be relatively easy to test this.

https://kubernetes.io/docs/concepts/containers/images/#updating-images

kate-goldenring: Great. I can test it. Assuming the expected result you mention, maybe we remove ImagePullPolicy since it defaults to the behavior you requested

bfjelds: i think, at the least, we should leave the ability to configure pullPolicy, but maybe only if a value is provided (so the default is "")

deployment/helm/values.yaml
```diff
@@ -109,7 +109,7 @@ debugEcho:
     # container used by debugEcho
     image: nginx:latest
     # pullPolicy is the debugEcho pull policy
-    pullPolicy: Always
+    pullPolicy: ""
```
bfjelds, line 112: not sure i know what happens if this empty value is passed into a PodSpec.

we may want to update each occurence of `imagePullPolicy: {{ .Values.*.image.pullPolicy }}` to be something like 

{{- with .Values.*.image.pullPolicy }}
imagePullPolicy: {{ .Values.*.image.pullPolicy }}
{{- end }}

so that the imagePullPolicy is only configured if someone sets a value.

kate-goldenring: I tried it out with the following dummy deployment and it defaulted to IfNotPresent behavior
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: pull-policy-deployment
  labels:
    app: pull-policy
spec:
  selector:
    matchLabels:
      app: pull-policy
  template:
    metadata:
      labels:
        app: pull-policy
    spec:
      containers:
      - name: pull-policy
        image: ghcr.io/kate-goldenring/pull-policy:latest
        imagePullPolicy: ""
```

kate-goldenring: But that seems to contradict this line in their documentation: "When imagePullPolicy is defined without a specific value , it is also set to Always", so maybe I should either test again or go with what you're suggesting

kate-goldenring: It says "Container image "ghcr.io/kate-goldenring/pull-policy:v0" already present on machine" when i deploy for the second time. Then, if I change to the latest tag, it pulls it: "Pulling image "ghcr.io/kate-goldenring/pull-policy:latest""

bfjelds: looks like you're getting the behavior you want ... i don't feel super strongly, but if it were my PR, i'd probably avoid `imagePullPolicy: ""`.

kate-goldenring: Looks much cleaner your way! Fixed

bfjelds (approved):

PR #214, opened 2021-01-20
Simplify mockall usage

jiayihu:
Updating mockall to latest allows to remove warnings https://github.com/asomers/mockall/pull/124 and also suppoort `async_trait`.

Unfortunately, ideally the mocks should be generated only during tests but `cfg(test)` doesn't work across dependencies https://github.com/rust-lang/rust/issues/59168. In the following code, `MockSlotQuery` cannot be imported from another crate:

```rust
#[cfg(test)]
use mockall::{automock, predicate::*};

#[cfg_attr(test, automock)]
#[async_trait]
pub trait SlotQuery {
    async fn get_node_slots(&self) -> SlotQueryResult;
}
```

If `cfg(test)` worked across deps, `mockall` could also be listed as dev-dep:

```toml
[dev-dependencies]
mockall = "0.9.0"
```

However I think this is already the current behaviour and since actual tests are behind `#[cfg(test)]`, mocks are stripped in release build.

**If applicable**:


bfjelds:
love, love, love the idea of this ... the async mockall code felt so hackish!

bfjelds:
looks like `cargo fmt` needs to be run on the PR

kate-goldenring:
Thanks for updating mockall! This is great. One more version update should do the trick.

jiayihu:
Ready to go 🚀

bfjelds (approved):

PR #213, opened 2021-01-19
Update crates with security vulnerabilities

romoh:
**What this PR does / why we need it**:
Updates dependencies caught by cargo-audit. This includes changes by `cargo update` as well as a manual update to hyper crate. With the changes, running `cargo-audit audit` does not yield any errors.
I could have used `cargo-audit` with fix enabled, but decided it is a good chance to also update our dependencies through cargo-update.

Akri builds successfully and test passes.
Closes #208 

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] added code adheres to standard Rust formatting (`cargo fmt`)
- [x] code builds properly (`cargo build`)
- [x] code is free of common mistakes (`cargo clippy`)
- [x] all Akri tests succeed (`cargo test`)
- [ ] inline documentation builds (`cargo doc`)
- [x] version has been updated appropriately (`./version.sh`)

PR #216, opened 2021-01-20
Add security audit workflow [SAME VERSION]

### LINKED ISSUE #212: Automate running cargo-audit to catch dependencies with security vulnerabilities
It is related to #208 

No, it is related to keeping Akri dependencies secure.

cargo-audit is a tool which inspects Cargo.lock files and compares them against the RustSec Advisory Database, a community database of security vulnerabilities. It can detect dependencies with security risks and thus it will be useful to automatically run cargo-audit with pull requests through gitactions to catch issues for newly added dependencies.

N/A

**Additional context**
N/A

romoh:
Add a security-audit workflow to catch security vulnerabilities. The workflow uses an [existing GitHub action](https://github.com/marketplace/actions/rust-audit-check) maintained by the cargo-audit team.
For now, I chose to trigger it on pull requests and pushes. The workflow can later be expanded to include other security auditing.

The workflow was tested by creating a pull request to a forked branch that introduces a security vulnerability (undo of #213) and the [workflow failed](https://github.com/romoh/akri/pull/2/checks?check_run_id=1737690827). Upon fixing it, the [workflow succeeded](https://github.com/romoh/akri/pull/2/checks?check_run_id=1737708699).

Since this is not a product related, no version update required.
Closes #212 

@bfjelds - Please chime in with your experience around token permissions. Not sure if that will be problematic.
Note: The action states that in case of token permission issues, the action will output all found advisories to the stdout instead. Another option is to switch the workflow to a daily cadence instead.

**If applicable**:


.github/workflows/security-audit.yml
```diff
@@ -0,0 +1,24 @@
+name: Security audit
```
bfjelds, line 1: should this be part of https://github.com/deislabs/akri/blob/main/.github/workflows/check-rust.yml?

romoh: I thought to group all security audits in one workflow and later we can expand it to include other security checks too. No strong opinion thou, I can move it to check-rust.yml since they run on the same triggers.

kate-goldenring (approved):

.github/workflows/check-rust.yml
```diff
@@ -20,6 +20,13 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  security_audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/audit-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
bfjelds, line 29: i'm not sure we should be adding our token here.  can we do this in a way that would work from a fork PR?

kate-goldenring: @romoh what happens if the token is removed?

romoh: The token is only required for generating a report of the vulnerabilities and warnings as an output of the action. If the token is not found an stdout error is shown but doesn't impact the result of the audit step.  I sent a test PR to confirm that this is indeed the case.

@kate-goldenring, to answer your question, We can remove the token if we don't care about the reporting part. I would keep it there just in case github figured how to get this working at some point, the reporting will light up automatically.

kate-goldenring: @romoh thanks for checking. That sounds like a good plan to me

bfjelds: you are still going to add a comment here with that information, right?

PR #228, opened 2021-01-27
use akri bot for dependencies autoupdate workflow [SAME VERSION]

romoh:
The [automated dependencies update PR](https://github.com/deislabs/akri/pull/226) did not trigger the test workflow as expected.
Upon looking at this further, it is a by-design behavior and a limitation enforced by GitHub to prevent infinite triggering of workflows, If one of the workflows uses the default GITHUB_TOKEN, it is not allowed to trigger other workflows. 
 
`When you use the repository's GITHUB_TOKEN to perform tasks on behalf of the GitHub Actions app, events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs.`

Therefore, I've created an akri-bot that has write access to the repo. The workflow will now run using its PAT instead as a workaround. 
Verified this on a forked repo.

**If applicable**:

### DIFF
--- a/.github/workflows/auto-update-dependencies.yml
+++ b/.github/workflows/auto-update-dependencies.yml
@@ -27,6 +27,6 @@ jobs:
     - name: Check for dependency updates
       uses: romoh/dependencies-autoupdate@v1.1
       with:  
-        token: ${{ secrets.GITHUB_TOKEN }}
+        token: ${{ secrets.AKRIBOT_TOKEN }}
         update-command: "'cargo update && cargo test'"
         on-changes-command: "'./version.sh -u -p'"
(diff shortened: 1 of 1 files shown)


bfjelds (approved):

jiria (approved):
Nice catch!

.github/workflows/auto-update-dependencies.yml
```diff
@@ -27,6 +27,6 @@ jobs:
     - name: Check for dependency updates
       uses: romoh/dependencies-autoupdate@v1.1
       with:  
-        token: ${{ secrets.GITHUB_TOKEN }}
+        token: ${{ secrets.AKRIBOT_TOKEN }}
```
kate-goldenring, line 30: super nit: but think this would look better as AKRI_BOT_TOKEN.

kate-goldenring (approved):

PR #232, opened 2021-02-03
Allow unreachable code introduced with feature-driven protocols [SAME VERSION]

bfjelds:
The catchall match is there in case someone turns off one of the handlers via features ... but at runtime the disabled feature is configured.  The build warning is annoying.


romoh (approved):
Do you want to add a comment in the code on why this is unreachable/added?

romoh (approved):

kate-goldenring (approved):

PR #236, opened 2021-02-05
Reduce frequency of dependency update [SAME VERSION]

bfjelds:
Our dependencies update on a daily basis ... not sure it makes sense for us to update our version each time someone else's do.  I chose a monthly cadence because it would, at most, cause us to update our version 12x per year, which doesn't seem too bad.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
 # https://help.github.com/en/articles/about-code-owners#codeowners-syntax
 
-*   @bfjelds @kate-goldenring @jiria @britel
+*   @bfjelds @kate-goldenring @jiria @britel @romoh
--- a/.github/workflows/auto-update-dependencies.yml
+++ b/.github/workflows/auto-update-dependencies.yml
@@ -2,7 +2,7 @@ name: Dependencies autoupdate
 
 on:
   schedule:
-    - cron: '0 0 * * *' # run daily at 12:00 am
+    - cron: '0 0 * * 1' # run on first day of each month at 12:00 am
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/auto-update-dependencies.yml
+++ b/.github/workflows/auto-update-dependencies.yml
@@ -2,7 +2,7 @@ name: Dependencies autoupdate
 
 on:
   schedule:
-    - cron: '0 0 * * 1' # run on first day of each month at 12:00 am
+    - cron: '0 0 1 * *' # run on first day of each month at 12:00 am UTC time
 
 env:
   CARGO_TERM_COLOR: always


jiria:
Is there some way to figure out if any dependency has an actively misused vulnerability? So that we could ensure those go in faster. If not, it is probably fine to keep per month.

bfjelds:
> Is there some way to figure out if any dependency has an actively misused vulnerability? So that we could ensure those go in faster. If not, it is probably fine to keep per month.

I could be wrong, but I was thinking that the vulnerability check was in the check-rust workflow: https://github.com/deislabs/akri/blob/7181bd32e41c87804dbcece910d2c1cd8a32b650/.github/workflows/check-rust.yml#L23

jiria:
> > Is there some way to figure out if any dependency has an actively misused vulnerability? So that we could ensure those go in faster. If not, it is probably fine to keep per month.
> 
> I could be wrong, but I was thinking that the vulnerability check was in the check-rust workflow (though, that only gets run for PRs and Releases):
> 
> https://github.com/deislabs/akri/blob/7181bd32e41c87804dbcece910d2c1cd8a32b650/.github/workflows/check-rust.yml#L23

Ah, ok. Is somebody getting an email when the check-rust workflow finds something so we could manually update? Or could that also auto-trigger the update PR?

romoh (approved):

romoh:
> > > Is there some way to figure out if any dependency has an actively misused vulnerability? So that we could ensure those go in faster. If not, it is probably fine to keep per month.
> > 
> > 
> > I could be wrong, but I was thinking that the vulnerability check was in the check-rust workflow (though, that only gets run for PRs and Releases):
> > https://github.com/deislabs/akri/blob/7181bd32e41c87804dbcece910d2c1cd8a32b650/.github/workflows/check-rust.yml#L23
> 
> Ah, ok. Is somebody getting an email when the check-rust workflow finds something so we could manually update? Or could that also auto-trigger the update PR?

This is only triggered on a pull request flow

jiria:
> This scenario can benefit from a scheduled workflow that does a security audit and send us an email or so. Jiri, do you want to track that with an issue?

Sure, we can prioritize separately. Can you please create the issue?

romoh (changes requested):
Looking at [GitHub documentation](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule). Looks like your change means "on every Monday of the week" 

`┌───────────── minute (0 - 59)
│ ┌───────────── hour (0 - 23)
│ │ ┌───────────── day of the month (1 - 31)
│ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
│ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
│ │ │ │ │                                   
│ │ │ │ │
│ │ │ │ │
* * * * *`

Instead it should be '0 0 1 * *'  
Also, nit - can you update the comment to state this is 12:00 am UTC time.

romoh (approved):

PR #244, opened 2021-02-12
Simplify matrix; Use all versions 1.16+ for all distros

### LINKED ISSUE #243: Failures of end-to-end tests on K8s 1.16|1.17
**Describe the bug**

Extensive (sorry!) testing of the end-to-end tests for the Webhook for Configurations uncovered issues with the existing end-to-end tests on the following Kubernetes distro-versions.

This bug identifies and tracks their resolution.

|Distro|Version|Failure|Link|
|--------|:-----:|-------|----|
|K3s|1.16|Akri not running in expected state|[1889124406](https://github.com/deislabs/akri/runs/1889124406?check_suite_focus=true)|
|MicroK8s|1.16|Nothing to do for helm3|[1882806752](https://github.com/deislabs/akri/runs/1882806752?check_suite_focus=true)|
|MicroK8s|1.17|Nothing to do for helm3|[1882806782](https://github.com/deislabs/akri/runs/1882806782?check_suite_focus=true)|

> **NOTE** The above links are non-exhaustive and assumed (!) to be consistent

**To Reproduce**

Run GitHub Actions workflow: Test K3s, Kubernetes, and MicroK8s

**Expected behavior**

Success

DazWilkin:

Attempting to resolve #243 

**If applicable**:


bfjelds:
it looks like on k3s 1.16, the controller isn't starting ... i wonder if there is something different about how k3s 1.16 handles scheduling on the master.

DazWilkin:
> it looks like on k3s 1.16, the controller isn't starting ... i wonder if there is something different about how k3s 1.16 handles scheduling on the master.

I created a VM running K3s 1.16 earlier today and am testing it.

I'm almost out of time today... it's that time of the day where Freddie and I go walking .... but will resume next week

bfjelds:
> it looks like on k3s 1.16, the controller isn't starting ... i wonder if there is something different about how k3s 1.16 handles scheduling on the master.

i misread the the test output, the controller is running.  it fails because it can't find the instances.

.github/workflows/run-test-cases.yml
```diff
@@ -88,211 +88,223 @@ jobs:
...
+          INSTALL_K3S_VERSION: ${{ matrix.kube.version }}
+        run: |
+          sudo curl -sfL https://get.k3s.io | sh -
+          sudo addgroup k3s-admin
+          sudo adduser $USER k3s-admin
+          sudo usermod -a -G k3s-admin $USER
+          sudo chgrp k3s-admin /etc/rancher/k3s/k3s.yaml
+          sudo chmod g+r /etc/rancher/k3s/k3s.yaml
+          sudo chmod 666 /etc/rancher/k3s/*
+          if [ "${{ matrix.kube.runtime }}" == "K3s-1.16" ]; then
```
bfjelds, line 179: this block of code worked for me if it ran after `until kubectl get node ${HOSTNAME,,}` ... maybe the rancher device-plugins folder takes a little time to be created?

DazWilkin: Ah, useful. Thanks. Will move it.

I dislike adding the exception for the version but, short of making `K3s-1.16` an entirely distinct step, this does the trick.

However, the workflow isn't (heavily) commented and I wonder what future readers will make of the change? 

Should we create a document that attempts to document what it all (!) does?

I feel the need to note somewhere your discovery of the K3s 1.16 device-plugins issue and record that as part of this change history (other than in the commits). Thoughts? [I have written a document for the `./test/run-webhook.py` script for the same reason]

bfjelds:
i think the two failures that aren't related to the rancher k3s device-plugin folder are essentially a case where the agent creates an instance before the controller is started, and for some reason the controller doesn't handle the previously created instance.  i created an issue: https://github.com/deislabs/akri/issues/248 for these failures.

.github/workflows/run-test-cases.yml
bfjelds: > However, the workflow isn't (heavily) commented and I wonder what future readers will make of the change?
> Should we create a document that attempts to document what it all (!) does?

I 100% think we should have a page for each distro with detailed steps ... and any quirks we find.

While I was shoveling snow (the first time), I imagined refactoring all of the shell commands in the test workflow into actual script files one to set up k3s, one to set up microk8s, etc) ... and those scripts could serve as documented how-to.  That would feed the effort to move the python tests to bash as well.  It sounded like a great plan while I was shoveling.

DazWilkin: I like this idea alot!

When I was walking Freddie through the icy-slush this morning on the trail (:smile: ), I was thinking, if each step were more reproducible (script|container), it would be easier to transform the workflow into a script that builds e.g. VMs on Azure|GCP that could the be more easily introspected (GitHub workflows are reproducible across runs but very difficult to reproduce outside of GitHub; my K3s and MicroK8s VMs on GCE behave different than the tests even though I'm attempting to use the same scripts).

DazWilkin: > this block of code worked for me if it ran after `until kubectl get node ${HOSTNAME,,}` ... maybe the rancher device-plugins folder takes a little time to be created?

Fixed: https://github.com/deislabs/akri/pull/244/commits/7ab2d18d632d39280c48ef2d681bbfaebb951e9e

DazWilkin:
> @DazWilkin, Thanks for doing this change. I'll add it to [the recent dependencies PR](https://github.com/deislabs/akri/pull/247) to unblock the test run.
> You may want to rebase after the other change is merged.

@romoh Thank you! Let me know when I should rebase and I shall.

DazWilkin:
@bfjelds I took a stab at documenting the workflow: https://github.com/deislabs/akri/pull/244/commits/15270acc234cce4a040aab799d583768de143894

bfjelds (approved):

romoh:
> > @DazWilkin, Thanks for doing this change. I'll add it to [the recent dependencies PR](https://github.com/deislabs/akri/pull/247) to unblock the test run.
> > You may want to rebase after the other change is merged.
> 
> @romoh Thank you! Let me know when I should rebase and I shall.

It is merged now. 
Thanks!

docs/workflows-run-test-cases.md
```diff
@@ -0,0 +1,120 @@
+# Test K3s, Kubernetes (Kubeadm) and MicroK8s
```
romoh, line 1: This looks great. Thanks for the write up.

romoh (approved):

docs/workflows-run-test-cases.md
DazWilkin: You're welcome. Thank you!

Issue #253, opened 2021-02-24
Simplify and optimize builds (Makefiles and Workflows)


bfjelds:
The Akri builds are based on Makefile.  The Makefile can be run locally and are run from our Workflows.  They work.  And that might be the best thing that can be said about them.

But they are hard to maintain and easy to get incorrect and have no linting/validation.

This is a pattern that is all over the scripts that seems ridiculous (there must be a better way, a more function-like way with some reuse):

```
ifeq (1, ${BUILD_ARM32})
	docker build $(CACHE_OPTION) -f $(DOCKERFILE_DIR)/Dockerfile.controller . -t $(PREFIX)/controller:$(LABEL_PREFIX)-$(ARM32V7_SUFFIX) --build-arg PLATFORM=$(ARM32V7_SUFFIX) --build-arg CROSS_BUILD_TARGET=$(ARM32V7_TARGET)
endif
```

The automated Workflow builds are based on the Makefiles.  For each Rust container, we build the entire source tree for a given platform, then put the appropriate binary into the appropriate container (so, we build the whole tree to grab the agent, the whole tree to grab the controller, etc).  It would be better to only build once if possible.

bfjelds:
Also, at a minimum, we should provide much more documentation around how the existing system works.

kate-goldenring:
@bfjelds can this PR be closed due to https://github.com/deislabs/akri/pull/270 which makes it so we only build our rust components once in our Workflows?

bfjelds:
I think #270 is only part of this issue ... part of this issue was intended to address the fact that the Makefiles were created by someone who doesn't know make (me).  There must be better ways to create the Makefiles that are more manageable and clear.

kate-goldenring:
That **Makes** sense. We can see this issue as a call for help for optimizing our makefiles

Issue #211, opened 2021-01-16
More fully document Akri's build process


DazWilkin:
**Is your feature request related to a problem? Please describe.**

There are many, independent but necessary changes required when extending Akri.

The [Development](./docs/development.md) doc does a good job but it is insufficient.

The build process is quite involved.

**Is your feature request related to a way you would like Akri extended? Please describe.**

More fully document the steps needed to extend Akri's build process with new components.

Use these documented steps as informal unit tests when changes are made to the build process.

Consider improvements to the build process reviewing the documented steps.

E.g. the Makefiles are very repetitive and would benefit from refactoring.

**Describe the solution you'd like**

I took a stab at documenting extending the build process for new components since I've just done this for the Webhook:

https://hackmd.io/@akri/S16Xoixyd

Feedback very welcome.

**Describe alternatives you've considered**

None.

**Additional context**

E.g. see https://github.com/deislabs/akri/pull/206

kate-goldenring:
This issue brings up a good point: Akri's development documentation could be more organized and comprehensive. 
Currently Akri's main development document is: https://docs.akri.sh/development/development
We also have one focused on using our makefiles: https://docs.akri.sh/development/building
And then the additional documentation @DazWilkin provided above:  https://hackmd.io/@akri/S16Xoixyd
This documentation should be brought together and reorganized and put into [Akri's documentation repository](https://github.com/deislabs/akri-docs)

PR #256, opened 2021-02-25
Simplify containers makefile

bfjelds:
Try to make it easier to add a new container and harder to make a mistake.

We have a lot of almost identical constructs in our makefile...try abstracting those into something that can be reused.

With this change, we'd be able to add a new rust container with one line (rather than lots and lots):

```
# arguments: new-thing corresponds to the Makefile target, new-thing-broker corresponds to the Dockerfile and container label
$(eval $(call add_rust_targets,new-thing,new-thing-broker))
```

Takes one tiny step towards helping with https://github.com/deislabs/akri/issues/253.

**If applicable**:


build/akri-containers.mk
```diff
@@ -58,339 +46,17 @@ ifeq (1, ${BUILD_ARM64})
...
-ifeq (1, ${BUILD_ARM32})
-	$(ENABLE_DOCKER_MANIFEST) docker manifest create --amend $(PREFIX)/video-streaming-app:$(LABEL_PREFIX) $(PREFIX)/video-streaming-app:$(LABEL_PREFIX)-$(ARM32V7_SUFFIX)
-endif
-ifeq (1, ${BUILD_ARM64})
-	$(ENABLE_DOCKER_MANIFEST) docker manifest create --amend $(PREFIX)/video-streaming-app:$(LABEL_PREFIX) $(PREFIX)/video-streaming-app:$(LABEL_PREFIX)-$(ARM64V8_SUFFIX)
-endif
+# Rust targets
+$(eval $(call add_rust_targets,controller,controller))
+$(eval $(call add_rust_targets,agent,agent))
+$(eval $(call add_rust_targets,udev,udev-video-broker))
```
kate-goldenring, line 52: To simplify, should we make the convention of having `add_rust_targets` only have one parameter, being the final container tag, and change the `udev` build name to `udev-video-broker`?

build/akri-dotnet-containers.mk
```diff
@@ -0,0 +1,102 @@
+define add_onvif_target
```
kate-goldenring, line 1: can this be done similarly to `add_rust_targets` so it is `add_dotnet_targets`?

kate-goldenring (commented):
We've been wanting to reuse our Makefile code for awhile! Thank you for "making" this happen :) Could a line be added to development.md on how to add a container to the build process? Or we could leave that to another PR which addresses this issue https://github.com/deislabs/akri/issues/211.

build/akri-containers.mk
bfjelds: we could do that.  might require changes to the udev workflow too.  i will check.

build/akri-dotnet-containers.mk
bfjelds: if i knew how to, i would.  really, the difference between the various add_*_targets is ONLY the differing number and values of the --build-args.

for rust, there are these 2 (with variants per platform): `--build-arg PLATFORM=$(AMD64_SUFFIX) --build-arg CROSS_BUILD_TARGET=$(AMD64_TARGET)`

for onvif, there are these 2 (with variants per platform): `--build-arg OUTPUT_PLATFORM_TAG=$(USE_OPENCV_BASE_VERSION)-$(AMD64_SUFFIX) --build-arg DOTNET_PUBLISH_RUNTIME=linux-x64`

for opcua, there are these 2 (with variants per platform): `--build-arg OUTPUT_PLATFORM_TAG=3.1-buster-slim --build-arg DOTNET_PUBLISH_RUNTIME=linux-x64`

for python, there is 1 (with variants per platform): `--build-arg PLATFORM=$(AMD64_SUFFIX)`

i'm sure there is a way to just have one add_targets, but i don't know it :(

build/akri-containers.mk
bfjelds: it looks like there are 4 things with differing makefile target prefixes and dockerfile/container names:

* udev
* onvif
* anomaly-detection
* streaming

bfjelds: i think i'll leave that simplification for the next round

romoh (commented):
I assume adding documentation will come in the next PR, correct?

build/akri-dotnet-containers.mk
```diff
@@ -0,0 +1,102 @@
...
+  ifeq (1, ${BUILD_AMD64})
+	  $(ENABLE_DOCKER_MANIFEST) docker manifest create --amend $(PREFIX)/$(2):$(LABEL_PREFIX) $(PREFIX)/$(1):$(LABEL_PREFIX)-$(AMD64_SUFFIX)
+  endif
+  ifeq (1, ${BUILD_ARM32})
+	  $(ENABLE_DOCKER_MANIFEST) docker manifest create --amend $(PREFIX)/$(2):$(LABEL_PREFIX) $(PREFIX)/$(1):$(LABEL_PREFIX)-$(ARM32V7_SUFFIX)
+  endif
+  ifeq (1, ${BUILD_ARM64})
+	  $(ENABLE_DOCKER_MANIFEST) docker manifest create --amend $(PREFIX)/$(2):$(LABEL_PREFIX) $(PREFIX)/$(1):$(LABEL_PREFIX)-$(ARM64V8_SUFFIX)
+  endif
+
```
romoh, line 46: Suggestion - Looks like a lot of the code here and the rest of the file can be further simplified by using different parameters for each of the build architectures initially then calling the docker commands once.

bfjelds: i don't know how to do that with a makefile.  if someone does, i'm happy to learn.

jiria: Haven't tried this, but would this work?

https://www.gnu.org/software/make/manual/html_node/Eval-Function.html

jiria: (Specifically the combination with the foreach construct.)

bfjelds: that's the mechanism i'm using already.  i don't know how to make a nested define and get it working.

jiria: My bad, for some reason I only saw one iteration. Second option would be to use CMake, that I think has more capabilities. But I think that is out of scope for now.

romoh: Honestly, It's been a while since I did that too so it was a suggestion. Deferring that to later PRs makes sense.

romoh (approved):

PR #257, opened 2021-02-26
Fix Simplify containers makefile [SAME VERSION]

bfjelds:
Fix manifest create mistake

### DIFF
--- a/build/akri-containers.mk
+++ b/build/akri-containers.mk
@@ -30,7 +30,7 @@ install-cross:
 .PHONY: akri
 akri: akri-build akri-docker-all
 akri-build: install-cross akri-cross-build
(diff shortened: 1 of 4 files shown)


build/akri-containers.mk
```diff
@@ -30,7 +30,7 @@ install-cross:
 .PHONY: akri
 akri: akri-build akri-docker-all
 akri-build: install-cross akri-cross-build
-akri-docker-all: akri-controller akri-agent akri-udev akri-onvif akri-streaming akri-opcua-monitoring akri-anomaly-detection akri-webhook-configuration
+akri-docker-all: akri-docker-controller akri-docker-agent akri-docker-udev akri-docker-onvif akri-docker-streaming akri-docker-opcua-monitoring akri-docker-anomaly-detection akri-docker-webhook-configuration
```
bfjelds, line 33: avoid running akri-build for each container ... `make akri` will build everything once and then create the containers (rather than building the rust source code before each container build).

PR #264, opened 2021-03-03
Move security audit checks to a daily schedule [SAME VERSION]

romoh:
As part of check-rust workflow, we also run `cargo audit` to detect any security vulnerabilities in Akri's dependencies. This was added to catch newly added dependencies imposing security risks to Akri. However, we ended up with those checks blocking PRs & code merges because of existing dependencies being out of date or a new vulnerability is discovered in a version of a crate that Akri is already consuming, so this check ended up being very noisy and hence this PR.

This change:
* Removes security audit checks from the 'check-rust'
* Creates a new workflow that runs daily that takes care of running the security audit and sends akri team an email in the event a security vulnerability was detected.

A v2 of this change would be to add another step in check-rust workflow to audit newly added dependencies. Another potential improvement is to attempt re-running the audit after a cargo update and in case that succeeds the workflow can kick off the autoupdate-dependencies workflow instead of sending an email to minimize manual intervention. 

Note: The audit github action will create an issue when a vulnerability is detected. I'm not very happy with that as that is not a recommended/secure approach. Usually in the event a security vulnerability is detected in some codebase, it is best to fix asap rather than documenting it through issues where others become aware of that and can leverage to compromise the code. I opened an [issue](https://github.com/actions-rs/audit-check/issues/180) on audit-check and will investigate other github action options.

**If applicable**:

### DIFF
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -30,7 +30,7 @@ jobs:
           password: ${{secrets.AKRI_BOT_PASSWORD}}
           subject: "Security vulnerability detected in ${{github.repository}}"
           body: |-
-            A security vulnerability was detected in one or more of Akri's. For more details, check the output of the [security audit workflow](https://github.com/${{github.repository}}/runs/actions)
+            A security vulnerability was detected in one or more of Akri's dependencies. For more details, check the output of the [security audit workflow](https://github.com/${{github.repository}}/runs/actions)
 
             Hint: In most cases, running the [auto-update dependencies](https://github.com/deislabs/akri/actions/workflows/auto-update-dependencies.yml) workflow will fix the issue.
 
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -1,0 +1,42 @@
+# This is a security audit workflow that runs security audit checks and send an email in case any vulnerabilities are detected.
+
+name: Security Audit
+on:
+  schedule:
+  - cron: '0 0 * * *' #runs daily at 12:00 am UTC
+    
+jobs:
+  security_audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Run security audit check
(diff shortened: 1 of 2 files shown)


.github/workflows/security-audit.yml
```diff
@@ -0,0 +1,42 @@
...
+        if: steps.cargo-audit.outcome != 'success'
+        uses: dawidd6/action-send-mail@v2
+        with:
+          server_address: smtp-mail.outlook.com
+          server_port: 587
+          username: ${{secrets.AKRI_BOT_EMAIL}}
+          password: ${{secrets.AKRI_BOT_PASSWORD}}
+          subject: "Security vulnerability detected in ${{github.repository}}"
+          body: |-
+            A security vulnerability was detected in one or more of Akri's dependencies. For more details, check the output of the [security audit workflow](https://github.com/${{github.repository}}/runs/actions)
```
bfjelds, line 33: very cool ... eventually, it would be cool to get a link to the actual action invocation that triggered this.

bfjelds (approved):

.github/workflows/security-audit.yml
romoh: I tried to use the github.run_id and github.run_number but then turned out I need an API call to get the id that can be used as the url.. so I abandoned the idea out of laziness :)

romoh: Actually I just noticed I might have been using a wrong url. Let me try again.

kate-goldenring (approved):
Will this automatically post to the same GitHub Issue each time or is that a vNext?

romoh:
> Will this automatically post to the same GitHub Issue each time or is that a vNext?

Based on the current github action implementation, for each vulnerability it will search the existing issues and check if an issue is already created, if not then it will create an issue for that vulnerability. 

While we can use other actions (with no issue creation), t the one above is the one recommended by rs tools and cargo audit so I'm leaning towards keeping it at the moment and maybe later we can contributed back to make issue creation optional.

PR #270, opened 2021-03-09
Try to ONLY build rust once per platform

bfjelds:
We currently have 4 rust containers.  Each of them build for 3 different platforms.  For each container-platform combination (12), we build ALL of the rust binaries.  If each build is 30 minutes, we are using 360 minutes of github action time when we only need 90 minutes.

This isn't super-impactful because the workflows run at the same time.  But it is wasteful.  And if we increase our platforms (by adding windows, etc) or our containers (highly likely), this number could really explode.

To address this:
1. I created a `build the rust code` workflow which runs for all of the platforms and exports the executable files in the applicable target/<platform>/release folder as a tar in an artifact.
2. I made the rust container build workflows triggered by the `build the rust code` workflow and had them download and expand the appropriate artifact tar

> Note: there is no supported way to download another workflow's artifacts, so this makes use of someone's attempt to enable this

I don't know if this will work.


.github/workflows/build-rust-code.yml
```diff
@@ -0,0 +1,92 @@
...
+        cross --version
+        make akri-cross-build-${{ matrix.arch.make-target }}
+
+    - name: Package build binaries
+      run: |
+        tar_manifest='/tmp/tar-contents.txt'
+        > $tar_manifest
+        for f in target/${{ matrix.arch.rust-target-path }}/release/*; do filetype=$( file "$f" ); case "$filetype" in *ELF*) echo "$f" >> $tar_manifest ;; esac; done
+        tar -cvf /tmp/rust-${{ matrix.arch.label }}-binaries.tar `cat $tar_manifest`
+
```
romoh, line 94: Nice, so looks like tar helps us retain permissions.

.github/workflows/build-rust-code.yml
```diff
@@ -0,0 +1,92 @@
...
+    - name: Upload target binaries as artifact
+      uses: actions/upload-artifact@v2
```
romoh, line 96: Do you know if there are storage restrictions on the artifacts in the repo?
We might want to configure the [retention period](https://github.com/actions/upload-artifact#retention-period) if that's a concern. Default is 90 days and seems unnecessary for what we are doing here.

romoh (approved):

.github/workflows/build-rust-code.yml
```diff
@@ -0,0 +1,101 @@
+name: Build Production Rust Code
```
kate-goldenring, line 1: This runs on tests too right? Should "Production" be removed?

kate-goldenring (approved):

.github/workflows/build-rust-code.yml
bfjelds: it is only production.  tarpaulin and tests build differently.

bfjelds:
I'm not sure how to test this without merging it.  It looks like the `workflow_run` executes in the branch context, so a **new** workflow that uses `workflow_run` will not execute until it is part of the branch (after it has been merged).  Of course.

bfjelds:
> i guess we can merge it and i can merge it into my pr and we can test it that way

i say you go ahead and merge your PR.  we can take a run at this tomorrow or sometime when i can pay attention to the build and fix what i messed up :)

kate-goldenring:
> > i guess we can merge it and i can merge it into my pr and we can test it that way
> 
> i say you go ahead and merge your PR. we can take a run at this tomorrow or sometime when i can pay attention to the build and fix what i messed up :)

Sounds like tomorrow is the day! I have to get the webhook certs fixed up via [this PR](https://github.com/deislabs/akri/pull/271) too.

kate-goldenring:
The end-to-end tests timed out when the last PR merged: https://github.com/deislabs/akri/actions/runs/640226875. Could you also extend the timeout on them in this PR? Will this optimize the e2e tests too because components will build faster?

bfjelds:
> The end-to-end tests timed out when the last PR merged: https://github.com/deislabs/akri/actions/runs/640226875. Could you also extend the timeout on them in this PR?

Sure.

> Will this optimize the e2e tests too because components will build faster?

Probably not.  The time between when a PR is merged and when the containers are pushed will probably not change.  That time will still be [build rust]+[build & push container].

romoh (commented):
Since this is only workflow changes, is the version update needed?

bfjelds:
> Since this is only workflow changes, is the version update needed?

The changes to the workflows will generate new binaries with new containers, I think it makes sense to rev the version.

kate-goldenring:
I have been looking at how to support slim and all-in-one agent builds. By default, no features would be set in the Agent, so running `cargo build` will always build the slim agent. So when building Akri's components in `akri-containers.mk`, either (1) both the slim and all-in-one agent could be built or (2) only one would be built. The former seems to fit better with the goal of this pr of running this build once. The latter would require build rust again for one of the builds. Correct me please @bfjelds if this summary doesn't seem right.

1. If BUILD_AGENT_ALL_IN_ONE is specified, an agent with all discovery handlers embedded will be built before building all other components. This executable will then be renamed to `agent-all-in-one` which it's dockerfile will reference. Then all other components are built including a slim `agent`.
```mk
ifeq (1, $(BUILD_AMD64))
ifeq (1, ${BUILD_AGENT_ALL_IN_ONE})
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(AMD64_TARGET) --manifest-path agent/Cargo.toml \
	--features "agent-all-in-one akri-udev akri-opcua akri-udev" && \
	mv target/$AMD64_TARGET/release/agent target/$AMD64_TARGET/release/agent-all-in-one
endif
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(AMD64_TARGET)
endif
akri-cross-build-arm32:
ifeq (1, ${BUILD_ARM32})
ifeq (1, ${BUILD_AGENT_ALL_IN_ONE})
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(ARM32V7_TARGET) --manifest-path agent/Cargo.toml \
	--features "agent-all-in-one akri-udev akri-opcua akri-udev" && \
	mv target/$ARM32V7_TARGET/release/agent target/$ARM32V7_TARGET/release/agent-all-in-one
endif
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(ARM32V7_TARGET)
endif
akri-cross-build-arm64:
ifeq (1, ${BUILD_ARM64})
ifeq (1, ${BUILD_AGENT_ALL_IN_ONE})
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(ARM64V8_TARGET) --manifest-path agent/Cargo.toml \
	--features "agent-all-in-one akri-udev akri-opcua akri-udev" && \
	mv target/$ARM64V8_TARGET/release/agent target/$ARM64V8_TARGET/release/agent-all-in-one
endif
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(ARM64V8_TARGET)
endif
```

2. A single agent is built and it is either all-in-one or slim based on BUILD_AGENT_ALL_IN_ONE but the executable is `agent` regardless
```mk
akri-cross-build: akri-cross-build-amd64 akri-cross-build-arm32 akri-cross-build-arm64
akri-cross-build-amd64:
ifeq (1, $(BUILD_AMD64))
ifeq (1, ${BUILD_AGENT_ALL_IN_ONE})
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(AMD64_TARGET) --workspace --exclude agent && \
	cross build --release --target=$(AMD64_TARGET) --manifest-path agent/Cargo.toml \
	--features "agent-all-in-one akri-udev akri-opcua akri-udev"
else
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(AMD64_TARGET)
endif
endif
akri-cross-build-arm32: 
ifeq (1, $(BUILD_ARM32))
ifeq (1, ${BUILD_AGENT_ALL_IN_ONE})
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(ARM32V7_TARGET) --workspace --exclude agent && \
	cross build --release --target=$(ARM32V7_TARGET) --manifest-path agent/Cargo.toml \
	--features "agent-all-in-one akri-udev akri-opcua akri-udev"
else
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(ARM32V7_TARGET)
endif
endif
akri-cross-build-arm64:
ifeq (1, ${BUILD_ARM64})
ifeq (1, ${BUILD_AGENT_ALL_IN_ONE})
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(ARM64V8_TARGET) --workspace --exclude agent && \
	cross build --release --target=$(ARM64V8_TARGET) --manifest-path agent/Cargo.toml \
	--features "agent-all-in-one akri-udev akri-opcua akri-udev"
else
	PKG_CONFIG_ALLOW_CROSS=1 cross build --release --target=$(ARM64V8_TARGET)
endif
endif
```

bfjelds:
i've been delaying merging this trying to think of a way to make this work with the e2e tests ... so that the e2e tests
1. do not do their own build during a PR
2. do not wait an arbitrary number of seconds before running on merge/release
3. run immediately if the helm charts are changed without any rust (or rust build) code changes
4. run immediately if the e2e test code is changed without any rust (or rust build) code changes

generally, i think that 1 & 2 are relatively easy ... we can make the e2e test workflow wait for the rust build workflow (for merge/release, we will need to have an arbitrary wait for the containers to build, but it'd be shorter).

for 3 & 4, it is harder ... mostly, i think, from a maintenance perspective.  we can create triggers for 3 & 4, but it would be something like ((run if changes in helm chart) && (run if NO changes in rust)).  we would need to essentially copy and paste the rust code trigger in the rust build and the e2e (negated).

i worry that this might start to be more complicated than it is worth.

kate-goldenring:
> 3. run immediately if the helm charts are changed without any rust (or rust build) code changes
> 4. run immediately if the e2e test code is changed without any rust (or rust build) code changes
> for 3 & 4, it is harder ... mostly, i think, from a maintenance perspective. we can create triggers for 3 & 4, but it would be something like ((run if changes in helm chart) && (run if NO changes in rust)). we would need to essentially copy and paste the rust code trigger in the rust build and the e2e (negated).

For 3 and 4 is the issue here that the build workflows are not kicked off by only helm chart or e2e test code changes? So there is no workflow to do the building and therefore need to specify the specific scenario in rust build workflow?

bfjelds:
> For 3 and 4 is the issue here that the build workflows are not kicked off by only helm chart or e2e test code changes? So there is no workflow to do the building and therefore need to specify the specific scenario in rust build workflow?

the issue of 3 & 4 is:
1.  we want a trigger that fires immediately (no need to wait for the non-existent rust build)
2.  if someone changes rust code AND the e2e code (or the helm charts), we need to wait for the rust code to build (i.e. `#1` cannot be triggered)

i think that the basic flow chart we need to cover is:

```
----
 |--- PR
   |--- Rust code changed
      |--- always use output of Build Production Rust Code
   |--- No rust code changed
      |--- e2e code changed
        |--- run tests based on published containers for contents of version.txt
      |--- helm code changed
        |--- run tests based on published containers for contents of version.txt
 |--- Merge | Release
   |--- Rust code changed
      |--- WAIT for ~5min after trigger for Build Production Rust Code (need containers to be built)
   |--- No rust code changed
      |--- e2e code changed
        |--- NO WAIT, just run tests based on published helm+containers for contents of version.txt
      |--- helm code changed
        |--- WAIT ~5min (need helm to be built), run tests based on newly published helm (use previously published containers reflecting contents of version.txt)
```

kate-goldenring:
>    |--- Rust code changed
>       |--- WAIT for ~5min after trigger for Build Production Rust Code (need containers to be built)

This is a beautiful chart and sounds like a great approach. The only thing that worries me is the 5 min wait seems to short. It feel like our builds normally take longer than that

bfjelds:
> This is a beautiful chart and sounds like a great approach. The only thing that worries me is the 5 min wait seems to short. It feel like our builds normally take longer than that

it feels too complicated to me.  and i'm not sure how we can do it without ending up with an unmanageable yaml file.  and at the end of it all, we still have some sleeps.  booooooooooo!

kate-goldenring:
@bfjelds can we make this PR two phases? With this first phase supporting one rust build but not solving the e2e test scenario?

bfjelds:
> @bfjelds can we make this PR two phases? With this first phase supporting one rust build but not solving the e2e test scenario?

sure.  if you have the time to fix whatever doesn't work after this gets merged, i can merge it now :)

kate-goldenring:
> sure. if you have the time to fix whatever doesn't work after this gets merged, i can merge it now :)
@bfjelds sounds like a plan.

bfjelds:
> > sure. if you have the time to fix whatever doesn't work after this gets merged, i can merge it now :)
> > @bfjelds sounds like a plan.

ok ... i've rebased and bumped the version again.  feel free to merge whenever the CI stops, or i will.

PR #282, opened 2021-03-23
Ensure Discovery Handler is marked as available again after Configuration deletions

kate-goldenring:
Currently, when using Akri, if a configuration is applied, deleted, and then reapplied, the Discovery Handler will not be called on the second application because it's status is not being changed from `Active` to `Waiting` upon Configuration deletion. This is because the thread is being canceled [here](https://sourcegraph.com/github.com/deislabs/akri/-/blob/agent/src/util/discovery_operator.rs#L677) before it's status can be changed. This moves the `select!` statement into `internal_do_discover` to ensure the DH's status is reset. 

**If applicable**:


bfjelds:
is there a test that can be modified to include this scenario?

kate-goldenring:
> is there a test that can be modified to include this scenario?

@bfjelds, great suggestion. I added an extensive test: 5d5eb440c73db73a

romoh (approved):

PR #284, opened 2021-03-24
Enable Canceling Previous Workflows on PRs from Forks [SAME VERSION]

kate-goldenring:
Our cancel workflow previously only worked on PRs not PRs from forks. Using `workflow_run` enables it to cancel previous workflows on PRs from forks. 
Added conditions so it does not cancel workflows on main or from releases.

### DIFF
--- a/.github/workflows/cancel-previous-pr-workflows.yml
+++ b/.github/workflows/cancel-previous-pr-workflows.yml
@@ -1,11 +1,14 @@
 name: Cancel Previous PR Workflows
 on:
-  pull_request:
-    branches: [ main ]
+  workflow_run:
+    workflows: ["Dependencies autoupdate", "Build Agents", "Build Anomaly Detection App Container", "Build Controller", "Build ONVIF Broker (.NET)", "Build OPC UA Monitoring Broker", "Build OpenCV Base", "Build Production Rust Code", "Build Rust CrossBuild", "Build UDEV Broker", "Build Video Streaming App Container", "Build Webhook Configuration", "Check Rust", "Check versioning", "Helm", "Tarpaulin Code Coverage", "Test K3s, Kubernetes, and MicroK8s"]
+    types:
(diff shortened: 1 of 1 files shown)


.github/workflows/cancel-previous-pr-workflows.yml
```diff
@@ -1,11 +1,14 @@
...
-  pull_request:
-    branches: [ main ]
+  workflow_run:
+    workflows: ["Dependencies autoupdate", "Build Agents", "Build Anomaly Detection App Container", "Build Controller", "Build ONVIF Broker (.NET)", "Build OPC UA Monitoring Broker", "Build OpenCV Base", "Build Production Rust Code", "Build Rust CrossBuild", "Build UDEV Broker", "Build Video Streaming App Container", "Build Webhook Configuration", "Check Rust", "Check versioning", "Helm", "Tarpaulin Code Coverage", "Test K3s, Kubernetes, and MicroK8s"]
+    types:
+      - requested
 jobs:
   cancel:
     runs-on: ubuntu-latest
+    if: (github.event.workflow_run.event != 'release') && (github.event.workflow_run.event != 'push')
```
bfjelds, line 10: nit: since this is titled "Cancel Previous **PR** Workflows", you might make this conditional just for PRs ... `if: startsWith(github.event_name, 'pull_request')` or `if: (github.event_name == 'pull_request')`

bfjelds: (functionally, my suggestion is probably the same now, but if we added a trigger for some other event like `label added`, this would need to be updated)

kate-goldenring: Great point! Adjusted in acb3083e631870c8. I used exact match instead of `startsWith` since we are avoiding `pull_request_target`.

romoh (approved):

PR #286, opened 2021-03-26
Build discovery handler images [SAME VERSION]

kate-goldenring:
No Discovery Handler images are currently being built, meaning that the flow in the `user-guide.md` fails due to no image being found for the Discovery Handler Pods.

This builds those images.

**If applicable**:

### DIFF
--- a/.github/workflows/build-discovery-handlers.yml
+++ b/.github/workflows/build-discovery-handlers.yml
@@ -1,0 +1,107 @@
+name: Build Discovery Hanlders
+
+on:
+  workflow_run:
+    workflows: ["Build Production Rust Code"]
+    branches: [main]
+    types: 
+      - completed
+
+jobs:
+
+  per-arch:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 40
+    strategy:
+      matrix:
+        arch:
+          - arm64v8
+          - arm32v7
+          - amd64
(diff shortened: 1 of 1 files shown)


.github/workflows/build-discovery-handlers.yml
```diff
@@ -0,0 +1,107 @@
...
+        container_registry_base_url: ghcr.io
+        container_registry_username: ${{ secrets.crUsername }}
+        container_registry_password: ${{ secrets.crPassword }}
+        makefile_component_name: ${{ matrix.akri-component }}
+        platform: ${{ matrix.arch }}
+        build_rust: "0"
+
+  multi-arch:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }} && ((github.event.workflow_run.event == 'release') || (github.event.workflow_run.event == 'push' && github.ref == 'refs/heads/main'))
+    needs: per-arch
```
bfjelds, line 72: this seems fine, but we are moving away from a workflow for each component (started slowly with agent, now this is more).  i wonder if we don't want to move to a singular build-rust-container workflow that has a really long list of matrix.akri-component?

bfjelds: the "needs" line here probably means that none of the DH multi-arch will run until all of the DH per-arch have run ... this is slightly different than our typical "componentX multi-arch will run after all platform per-arch for componentX finish"

bfjelds (approved):

PR #287, opened 2021-03-26
Increase e2e test timeout [SAME VERSION]

kate-goldenring:
On pushes to main, the e2e tests waits 25 minutes for containers to build before starting. They then have 10 minutes to complete. This doesn't seem to be sufficient. Adding 10 minutes more.

### DIFF
--- a/.github/workflows/run-test-cases.yml
+++ b/.github/workflows/run-test-cases.yml
@@ -98,7 +98,7 @@ jobs:
   test-cases:
     needs: build-containers
     runs-on: ubuntu-18.04
-    timeout-minutes: 45
+    timeout-minutes: 60
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/run-test-cases.yml
```diff
@@ -98,7 +98,7 @@ jobs:
   test-cases:
     needs: build-containers
     runs-on: ubuntu-18.04
-    timeout-minutes: 35
+    timeout-minutes: 45
```
romoh, line 101: Do we want to increase that generously to avoid having to increase it later again?

romoh: maybe 60 mins?

kate-goldenring: makes sense. bumped it to 60

PR #290, opened 2021-03-31
Remove un-utilized CRD field

kate-goldenring:
This PR removes two fields that existed in our CRDs that were not being used.
~~1. `Configuration.properties`
  Our Configuration CRD has a `properties` field that was originally intended to be a place where operators could set properties that would ultimately be set as environment variables in broker Pods; however, no support is currently in place for propagating these properties into brokers and broker Pod environment variables can be set in the broker PodSpec. As it is not being leveraged and may not need to, it makes sense to remove it.~~ Support of this field is instead being added in ref #288 
2. `Instance.rbac`
  This field was a place holder for rbac support, which Akri now provides via creation of ClusterRoles for the Agent and Controller for accessing instances, when enabled via Helm.

As this is a breaking change, it bumps the minor version.

**If applicable**:


romoh (approved):

agent/src/util/device_plugin_service.rs
```diff
@@ -497,7 +497,6 @@ async fn try_create_instance(
         nodes: vec![dps.node_name.clone()],
         device_usage,
         metadata: dps.device.properties.clone(),
-        rbac: "rbac".to_string(),
```
jiria, line 500: Separate question, should we have some RBAC support for brokers that we deploy? And how about DH?

kate-goldenring: I could see Akri supporting NetworkPolicies for limiting access to our broker Services, as mentioned in [the security proposal](https://github.com/deislabs/akri/blob/main/docs/proposals/security.md#device-access-via-broker-services). What rbac settings would you impose on the broker itself or the DHs?

jiria: I was wondering in case the broker or DH had some need of their own to access ApiServer, then it would be good to be able to pass some RBAC rules.

kate-goldenring: I think it makes sense to remove this rbac setting until we flush out a scenario for what you are mentioning. These RBAC settings could be applied to the cluster. I dont see why the broker or DH Pods would need to be aware of them. I could see helm templates being updated to support this though.

bfjelds (approved):

romoh (approved):

PR #292, opened 2021-04-01
Use unshared debug echo in e2e tests [SAME VERSION]

kate-goldenring:
Before our latest DH changes, we used the debug echo Configuration in our e2e test to discover **unshared** devices ([past state](https://github.com/deislabs/akri/blob/5924ff72219fd9eb91b835e81c1115d04a533ef4/test/run-end-to-end.py#L34)). Currently we are discovering shared devices in e2e which is leading to a 5 min delay before devices are marked offline. This will speed up our tests and revert back to the previous pattern.

If we want to do shared, than we should change the test to one 5 minute sleep instead of checks every 20 seconds.

### DIFF
--- a/test/run-end-to-end.py
+++ b/test/run-end-to-end.py
@@ -31,7 +31,7 @@ def main():
     print("Providing Akri Helm chart with CRI args: {}".format(cri_args))
     extra_helm_args = shared_test_code.get_extra_helm_args()
     print("Providing Akri Helm chart with extra helm args: {}".format(extra_helm_args))
-    helm_install_command = "helm install akri {} --set agent.full=true --set debugEcho.configuration.enabled=true --set debugEcho.configuration.name={} --set agent.allowDebugEcho=true {} {} --debug ".format(helm_chart_location, shared_test_code.DEBUG_ECHO_NAME, cri_args, extra_helm_args)
+    helm_install_command = "helm install akri {} --set agent.full=true --set debugEcho.configuration.enabled=true --set debugEcho.configuration.shared=false --set debugEcho.configuration.name={} --set agent.allowDebugEcho=true {} {} --debug ".format(helm_chart_location, shared_test_code.DEBUG_ECHO_NAME, cri_args, extra_helm_args)
     print("Helm command: {}".format(helm_install_command))
     os.system(helm_install_command)
     


romoh (approved):

romoh (approved):
Good find.

kate-goldenring:
I wonder if instead the default Helm value should be `shared=false`.

romoh:
> I wonder if instead the default Helm value should be `shared=false`.

I expect the majority of resources that Akri discovers to be over the network so defaulting to 'true' makes sense to me.. 
@bfjelds thoughts?

kate-goldenring:
> > I wonder if instead the default Helm value should be `shared=false`.
> 
> I expect the majority of resources that Akri discovers to be over the network so defaulting to 'true' makes sense to me..
> @bfjelds thoughts?

Only reason im partial to unshared is that the way debug echo marked devices as offline is on a per agent basis which is confusing with multi-node clusters. If you have 2 workers and mark the devices offline in one agent but not the other, the isntances will remain because the other agent still sees them as online.

romoh:
> > > I wonder if instead the default Helm value should be `shared=false`.
> > 
> > 
> > I expect the majority of resources that Akri discovers to be over the network so defaulting to 'true' makes sense to me..
> > @bfjelds thoughts?
> 
> Only reason im partial to unshared is that the way debug echo marked devices as offline is on a per agent basis which is confusing with multi-node clusters. If you have 2 workers and mark the devices offline in one agent but not the other, the isntances will remain because the other agent still sees them as online.

I see your point. 
We can unshare it then and maybe later we can come up with a way to easily test shared resources (or make debug echo offline state shared).

bfjelds:
generally, i think that debugecho and its ONLINE/OFFLINE is specifically for testing.  i'm not too concerned about what the default is because of that.  our test should be explicit rather than defaulting to a value, so the default shouldn't impact the test (and the test ought to be testing shared=true because that scenario is more complex).

romoh:
> generally, i think that debugecho and its ONLINE/OFFLINE is specifically for testing. i'm not too concerned about what the default is because of that. our test should be explicit rather than defaulting to a value, so the default shouldn't impact the test (and the test ought to be testing shared=true because that scenario is more complex).

Totally agree. If debug echo is the way we test our scenarios then we should use shared=true as this is what we expect most devices to look like. At the same time, it seems the way we set it to be offline doesn't align with that fact that's shared. 
I'm trying to think of ways to make the offline state global without too much work. Can we share the file that we add "OFFLINE" to over network & then all DHs can reference that?

bfjelds:
> Totally agree. If debug echo is the way we test our scenarios then we should use shared=true as this is what we expect most devices to look like. At the same time, it seems the way we set it to be offline doesn't align with that fact that's shared.
> I'm trying to think of ways to make the offline state global without too much work. Can we share the file that we add "OFFLINE" to over network & then all DHs can reference that?

definitely sounds cool.  i'm not sure how to justify worrying about this though.  i don't think we have a test that would exercise this.

kate-goldenring:
Closing due to decision to continue to test shared scenario

PR #304, opened 2021-04-13
Remove udev directory mount from Agent

kate-goldenring:
It appears that the udev directory does not need to be mounted into the full Agent or udev discovery handlers in order to search it. The udev Discovery Handler are already functioning without the mount. This removes the mount by default, allowing it to be added back in the agent with `--set udev.discovery.host.udev=/run/udev` and in the udev Discovery Handler with `--set udev.discovery.host.udev=/run/udev`.

**If applicable**:


kate-goldenring:
@bfjelds does this change to make it default not mounted but remountable via helm seem good to you?

deployment/helm/templates/udev-discovery-handler.yaml
```diff
@@ -49,6 +49,10 @@ spec:
         volumeMounts:
         - name: discovery-handlers
           mountPath: /var/lib/akri
+        {{- if .Values.udev.discovery.host.udev }}
```
bfjelds, line 52: nit: i'd probably prefer a more specific name ... maybe `mountUdev`

bfjelds (approved):

PR #305, opened 2021-04-16
Set compute resource limits and requests for all Akri components

kate-goldenring:
Currently, Akri's component Pods are not being constrained with resource requests and limits. A memory and CPU usage analysis of Akri's components was done leveraging [a blog by @DazWilkin](https://pretired.dazwilkin.com/posts/210305/) that investigated Akri resource usage using Virtical Pod Autoscaler (VPA). 

Reference [this document](https://hackmd.io/DSVfdh82TQS_Tz-SgxFrJw) for more details on the results of the resource analysis using VPA. While the upper bound resource limit changed overtime, the request and target resource usage stayed the same. The VPA seemed to choose the minimum values for Akri's components as their actual usage was way below these minimums. 

More analysis should be done to see if/how the recommendations change as Akri is stressed.

**If applicable**:


jiria:
This is great, thanks Kate. Left few comments in the HackMD doc.

deployment/helm/values.yaml
```diff
@@ -59,6 +59,17 @@ controller:
   # nodeSelectors is the array of nodeSelectors used to target nodes for the Akri Controller to run on
   # This can be set from the helm command line using `--set controller.nodeSelectors.label="value"`
   nodeSelectors: {}
+  resources:
+    # memoryRequest defines the minimum amount of RAM that must be available to this Pod
+    # for it to be scheduled by the Kubernetes Scheduler
+    memoryRequest: 100Mi
```
bfjelds, line 65: if we are going to have a universal value, i wonder if it wouldn't make sense to have it declared once ... and have a per-component override.

something like 

```
memoryRequest: 100mi
...
agent:
  resources:
    memoryRequest: 
...
```

then the template can choose .Values.agent.resources.memoryRequest if it is provided ...  otherwise it can choose .Values.memoryRequest.

with that, users could with one setting adjust all the pods to a limit.

kate-goldenring: After reducing the minimums, the values have diverged -- expect for the Discovery Handlers. I updated the [HackMD](https://hackmd.io/@akri/rkmkDJ88O) document and push the latest values. Since they are different, it seems a universal value is not needed/applicable?

kate-goldenring:
> This is great, thanks Kate. Left few comments in the HackMD doc.

@jiria I lowered the limits like you suggested and got a better set of values. I've updated the MD with the findings and the PR with the values

deployment/helm/templates/agent.yaml
```diff
@@ -43,6 +43,13 @@ spec:
         {{- with .Values.agent.image.pullPolicy }}
         imagePullPolicy: {{ . }}
         {{- end }}
+        resources:
```
bfjelds, line 46: should we define these for the configuration brokerPodSpecs as well?

kate-goldenring: Since they are user provided, i am not sure how we can make the best estimate.

bfjelds: i think w e could provide estimates for the brokers we provide: udev, opcua, debug echo, and onvif

kate-goldenring: So we would default to using the requirements of our containers?

bfjelds: our provided brokers (as specified by our provided configuration files) would have cpu and memory requests/limits like all of our other containers.  there would be configurable settings in the values yaml, like our other settings.  unless that doesn't make sense?

deployment/helm/values.yaml
bfjelds: fair enough :)

bfjelds (approved):

PR #308, opened 2021-04-16
Update cluster setup documentation

kate-goldenring:
Clarify cluster setup documentation. Be sure to signal that only one of the three options needs to be installed.
Also, emphasize that adding nodes is optional and discourage it in the demos.

### DIFF
--- a/docs/end-to-end-demo.md
+++ b/docs/end-to-end-demo.md
@@ -123,6 +123,13 @@ Look at the Configuration and Instances in more detail.
 1. Deploy a video streaming web application that points to both the Configuration and Instance level services that were automatically created by Akri.
     ```sh
     kubectl apply -f https://raw.githubusercontent.com/deislabs/akri/main/deployment/samples/akri-video-streaming-app.yaml
+    ```
+    For MicroK8s
+    ```sh
+    watch microk8s kubectl get pods
+    ```
+    For K3s and vanilla Kubernetes
+    ```sh
     watch kubectl get pods
     ```
 1. Determine which port the service is running on. Be sure to save this port number for the next step.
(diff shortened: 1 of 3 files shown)


docs/end-to-end-demo.md
```diff
@@ -69,7 +69,7 @@ The following will be covered in this demo:
     > ```
 
 ## Setting up a cluster
-Reference our [cluster setup documentation](./setting-up-cluster.md) to set up a cluster or adapt your currently existing cluster. 
+Reference our [cluster setup documentation](./setting-up-cluster.md) to set up a **single-node** cluster or adapt your currently existing cluster. This documentation assumes you are using a single-node cluster, so if installing K3s or MicroK8s, you can skip the last step of the installation instructions.
```
jiria, line 72: Curious about the emphasis? Can we not emphasize that or also mention that you can bring your own cluster with any number of nodes? Otherwise it might sound that BYOC needs to be a single node too.

kate-goldenring: In a way it does need to be single node too. Or at least to follow the documentation exactly.  I want the demo to be as quick as possible to do, so i'd prefer people dont try to set up multinode clusters if they dont already have a cluster

jiria: Ok, perhaps then add a disclaimer that you can use multi node cluster this demo, but be aware that you might see multiple instances where we show one. For deployments outside of this specific e2e demo, multi node clusters are definitely supported. Or something like that.

kate-goldenring: @jiria how do the changes in 8ddc545ee7 look? I am open to adjusting again so it makes the most sense!

jiria: Looks good, thanks Kate.

bfjelds (approved):

PR #311, opened 2021-04-16
Add documentation on setting a Configuration's `brokerProperties`

kate-goldenring:
Previously, we had documentation on setting broker Pod environment variables via a Configuration's broker PodSpec. The convention now is to use a Configurations `brokerProperties` for setting env vars in Pods that request akri resources, whether the Pods are deployed by the Akri Controller (with a specified broker PodSpec) or manually.

Also the `udev-video-sample.md` document is a bit redundant to the end to end demo doc and `udev-configuration.md` doc and should potentially be removed.

### DIFF
--- a/docs/broker-development.md
+++ b/docs/broker-development.md
@@ -75,3 +75,22 @@ helm install akri akri-helm-charts/akri-dev \
     --set udev.configuration.brokerPod.image.repository="ghcr.io/brokers/camera-broker" \
     --set udev.configuration.brokerPod.image.tag="v0.0.1" 
 ```
+
+## Specifying additional broker environment variables in a Configuration
+You can request that additional environment variables are set in Pods that request devices discovered via an Akri
+Configuration. These are set as key/value pairs in a Configuration's `brokerProperties`. For example, take the scenario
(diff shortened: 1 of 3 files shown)


docs/broker-development.md
```diff
@@ -74,4 +74,23 @@ helm install akri akri-helm-charts/akri-dev \
     --set udev.configuration.discoveryDetails.udevRules[0]='KERNEL=="video[0-9]*"' \
     --set udev.configuration.brokerPod.image.repository="ghcr.io/brokers/camera-broker" \
     --set udev.configuration.brokerPod.image.tag="v0.0.1" 
+```
+
+## Specifying additional broker environment variables in a Configuration
+You can request that additional environment variables are set in Pods that request devices discovered via an Akri
+Configuration. These are set as key/value pairs in a Configuration's `brokerProperties`. For example, take the scenario
+of brokers being deployed to USB cameras discovered by Akri. You may wish to give the brokers extra image about the
```
bfjelds, line 82: maybe: "... may wish to give the brokers extra **information** about the ... "

kate-goldenring: good catch. fixed in fc2b066

PR #316, opened 2021-04-22
Rename Configuration Validating Webhook

kate-goldenring:
Akri supports a webhook for validating configurations. This is described as "webhook configuration" in our Helm chart and rust binary. A name that described what it does, act as a "configuration validator", would be preferable.

I noticed that that "webhook-configuration" is misleading when renaming Configuration templates in this PR https://github.com/deislabs/akri/pull/315.
The name could be interpreted as an Akri Configuration for "discovering" webhooks.

**If applicable**:


romoh (approved):

bfjelds (approved):

kate-goldenring:
So after looking further, it looks like webhook-configuration is a fine naming choice in the Kubernetes world, as our is a [`ValidatingWebhookConfiguration`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#configure-admission-webhooks-on-the-fly) as can also be noted by looking at the [Helm template for the webhook](https://github.com/deislabs/akri/blob/main/deployment/helm/templates/webhook-configuration.yaml#L115). I am now thinking a name change may not be needed or appropriate. @romoh @bfjelds thoughts?

bfjelds:
> So after looking further, it looks like webhook-configuration is a fine naming choice in the Kubernetes world, as our is a [`ValidatingWebhookConfiguration`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#configure-admission-webhooks-on-the-fly) as can also be noted by looking at the [Helm template for the webhook](https://github.com/deislabs/akri/blob/main/deployment/helm/templates/webhook-configuration.yaml#L115). I am now thinking a name change may not be needed or appropriate. @romoh @bfjelds thoughts?

i suppose it is safer to, when in doubt, not make a change.  less chance of introducing a bug.  we can always change it later if someone decides it isn't appropriate.

kate-goldenring:
Closing this. Consider reopening if request for name change occurs

PR #327, opened 2021-04-28
Remove branch specification from dependent container build workflows [SAME VERSION]

kate-goldenring:
Upon creating a release, our workflows that are dependent on the Build Production Rust Code workflow are not triggered due to the `main` branch specification.

### DIFF
--- a/.github/workflows/build-agent-container.yml
+++ b/.github/workflows/build-agent-container.yml
@@ -3,7 +3,6 @@ name: Build Agents
 on:
   workflow_run:
     workflows: ["Build Production Rust Code"]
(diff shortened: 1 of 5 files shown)


romoh (commented):
Removing the 'main' restriction will trigger it on forks as well. How about adding another condition `on release`

romoh:
> Removing the 'main' restriction will trigger it on forks as well. How about adding another condition `on release`

Actually I just noticed this only executes only when the other workflow completes and the other one has the on release trigger.

romoh (approved):

Issue #294, opened 2021-04-06
Create an Akri namespace for deploying Akri's components instead of default namespace


romoh:
**Is your feature request related to a problem? Please describe.**
Currently Akri's components are deployed to the default namespace which is not conventional. 

**Describe the solution you'd like**
Instead we should create an Akri namespace where the CRDS, agent, DHs and controller will be deployed. Brokers can live in their own namespace as defined by the user.
This also ties back to #268 where users can type `kubectl get all -A` & see Akri's components or specify the Akri namespace for that.

rodz:
Been working on it since yesterday, stuck trying to make the new namespace (named akri) still communicate with the rest of the cluster. Might have a pr solving this one soon.

kate-goldenring:
Closing due to the conclusion in PR ref #340 to stick to helm conventions of installing a package into a namespace via `-n` (ie `helm install akri ... -n akri-ns`)

kate-goldenring:
Closing due to the conclusion in PR ref #340 to stick to helm conventions of installing a package into a namespace via `-n` (ie `helm install akri ... -n akri-ns`)

PR #340, opened 2021-06-09
Create akri as the default namespace

rodz:

Solves #294 

 This creates a new default namespace (named "akri") for akri components and would be used if the user does not specify a particular namespace. 

 This particular implementation comes with pros and cons when dealing with helm templates, here is a summary.

**Pros**:

 If a user doesn't specify a namespace it makes sense for it to be arranged in an organized namespace and creates a good standard for Akri usage.

 By default Akri won't pollute the cluster's default namespace.

**Cons**:

 Helm doesn't support creating new default namespace other than "default", so this particular implementation creates it via template.

 What that means is that if the user specify the namespace during Akri's helm installation (for example "-n iot") it would still be deployed to the "akri" namespace. It is still possible to change it, but it is required to override the `.Values.namespace.name` variable to set a custom namespace, contradicting the helm already established conventions.

 Since now a namespace creation is part of the template it could result in some problems, for example if the user uninstall Akri the namespace it was present would also be deleted, even if some non-akri related components were deployed there.

**Conclusion**:

 My conclusion is that we should stick with Helms patterns and good practice conventions and not set a namespace via templates for now, leaving it to the user decide where to deploy Akri. It is still a good feature to keep in mind and this Issue could definitely be reviewed if we see any updates to this aspect of Helm. 

 We should still update our documentation in order to incentivize the creation of an Akri namespace during installation, so we could still hit our goal to cut the number of components on the default namespace.

### DIFF
--- a/deployment/helm/templates/rbac.yaml
+++ b/deployment/helm/templates/rbac.yaml
@@ -15,7 +15,6 @@ kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: "akri-controller-role"
-  namespace: {{ .Values.namespace.name }}
 rules:
 - apiGroups: [""]
   resources: ["pods", "services"]
@@ -34,7 +33,6 @@ kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: "akri-agent-role"
-  namespace: {{ .Values.namespace.name }}
 rules:
 - apiGroups: [""]
   resources: ["pods"]
@@ -50,7 +48,6 @@ apiVersion: 'rbac.authorization.k8s.io/v1'
 kind: 'ClusterRoleBinding'
 metadata:
   name: 'akri-controller-binding'
-  namespace: {{ .Values.namespace.name }} # Previous .Release.Namespace
 roleRef:
   apiGroup: ''
   kind: 'ClusterRole'
@@ -64,7 +61,6 @@ apiVersion: 'rbac.authorization.k8s.io/v1'
 kind: 'ClusterRoleBinding'
 metadata:
   name: 'akri-agent-binding'
-  namespace: {{ .Values.namespace.name }} # Previous .Release.Namespace
 roleRef:
   apiGroup: ''
   kind: 'ClusterRole'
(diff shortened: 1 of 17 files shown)


deployment/helm/templates/rbac.yaml
```diff
@@ -46,27 +48,25 @@ apiVersion: 'rbac.authorization.k8s.io/v1'
 kind: 'ClusterRoleBinding'
 metadata:
   name: 'akri-controller-binding'
-  namespace: {{ .Release.Namespace }}
```
romoh, line 49: Since we are following helm guidelines, we should remove these as part of updating our documentation with our namespace guidelines.

rodz: Nice to remember! Will be sure to remove them there.

deployment/helm/values.yaml
```diff
@@ -25,6 +25,11 @@ rbac:
   # enabled defines whether to apply rbac to Akri
   enabled: true
 
+# namespace is the Kubernetes namespace that will be created for storing
+# Akri related deployments
+namespace:
```
bfjelds, line 30: why are introducing two levels here (namespace.name) rather than just one level (namespace)?

kate-goldenring:
> My conclusion is that we should stick with Helms patterns and good practice conventions and not set a namespace via templates for now

@rodz given your conclusion -- and i agree it makes sense to use helm conventions -- should we close this PR and re-open if conventions ever change and instead create a PR that adds to our documentation setting namespace during helm installs:
```sh
helm instal akri ... -n akri
```

deployment/helm/values.yaml
rodz: Just to make more clear and follow the patterns from the other values (service_name.data), but I guess it could've be in a single level variable.

rodz:
> > My conclusion is that we should stick with Helms patterns and good practice conventions and not set a namespace via templates for now
> 
> @rodz given your conclusion -- and i agree it makes sense to use helm conventions -- should we close this PR and re-open if conventions ever change and instead create a PR that adds to our documentation setting namespace during helm installs:
> 
> ```shell
> helm instal akri ... -n akri
> ```

Sure Kate! Will also make sure to reference this PR on the next one. Thanks for everyone that gave their feedback.

Issue #106, opened 2020-11-05
Support of MQTT


didier-durand:
**Is your feature request related to a way you would like Akri extended? Please describe.**

[MQTT ](https://mqtt.org/)is a key protocol in the IoT world. It would be extremely beneficial if Akri could directly communicate with devices implementing this protocol to report data / get instructions. That would make the path between device and kubernetes application much shorter than it is for other projects lilke KubeEdge whose [architecture](https://kubeedge.io/en/docs/kubeedge/) entails more elements between device and K8s application.

Many [use cases ](https://mqtt.org/use-cases/)show how it's used today or will be used in the future.

**Describe the solution you'd like**

Akri may rely on existing [protocol implementations](https://mqtt.org/software/) to deliver this feature rapidly

**Describe alternatives you've considered**

The only one known as of now is Kubeedge with the disadvantage mentioned above.

**Additional context**

[https://dzone.com/articles/why-mqtt-has-become-the-de-facto-iot-standard](https://dzone.com/articles/why-mqtt-has-become-the-de-facto-iot-standard)

DazWilkin:
@didier-durand I'm intrigued by this because MQTT was my initial thought on a protocol too but now, I'm unsure.

Assuming devices are able to publish to MQTT topics, what benefit do you see in writing Akri-specific protocol support for (an) MQTT (broker) over an above, running an MQTT broker in the cluster?

didier-durand:
@DazWilkin:

Hi Daz,

This is why I suggested MQTT:

1. When I compare Akri to [Kubeedge](https://kubeedge.io/en/) also aimed at IoT, the big difference from my perspective is the architecture: Akri tries to shorten to a very minimum the "distance" between K8s and the connected object while Kubeedge has additional [layers and components](https://kubeedge.io/en/docs/kubeedge/#architecture), which make the architecture less homogeneous.
2. Those specific components (they include an MQTT broker) require then specific "out-of-K8s" management procedures and tools for patching, monitoring, etc.
3. So, I suggested supporting MQTT in Akri to leverage this advantage of the shortest possible to K8s. Similarly to your sayings, you could envision a solution to relay the video stream of the e2e demo (in this repo) to a remote K8s without the use of v4l2 loopback. But, the use of this solution allows working from within K8s straight away. That's the same idea for MQTT from my perspective.

This has numerous advantages coming from the normal benefits of a pure cloud-native architecture:

1. you can update the entire chain of components very fast by pushing new versions of the corresponding container images.
[...]

PR #341, opened 2021-06-14
docs: add MQTT Proposal Draft v0.1.0

shantanoo-desai:
- Add detailed description of possible Integration
  of Akri and MQTT
- Add diagrams for different descriptions

This PR introduces an initial draft to integrate MQTT Protocol to Akri. 
Since there is already a discussion thread for Support of MQTT (see #106 ), this proposal builds upon the discussion and helps
provide a more detailed description of the protocol under question (MQTT) and how a possible integration might look like.

This documentation is written while considering some already provided proposals under the `docs/proposals` directory.
It contains some questions pertaining to implementation as well as integration aspects of MQTT with Akri.

This proposal is bound to be edited and changed based on discussions/feedback/criticisms.

This PR should not break any existing Test Pipelines. 

**If applicable**:


docs/proposals/mqtt.md
```diff
@@ -0,0 +1,234 @@
...
+Agent implements [MQTT Protocol](https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.html), a light-weight
+publish/subscribe messaging protocol widely used in Internet of Things (IoT).
+
+Supporting MQTT as a protocol in Akri provides a mechanism by which Application Developers (IoT, Kubernetes)
+can provide scalable applications based on current status of the IoT Devices within an environment. 
+
+Depending on the IoT Device's availability the application pertaining to the data such a device produces
+can be scaled accordingly.
+
+## What can Akri solve?
```
kate-goldenring, line 13: Should this be narrowed based on the discovery strategy. For example, Akri + MQTT Broker discovery makes it so an operator can say, run this client (Akri "broker") on nodes that can see an MQTT Broker with x,y, and z topics. How do MQTT clients normally respond to topics no longer being published (devices going away)? It seems like with this implementation, if a topic goes away that a client/akri broker was using, Akri would automatically bring down the client. Is that ideal?

docs/proposals/mqtt.md
```diff
@@ -0,0 +1,234 @@
...
+
+> NOTE: MQTT is completely data-agnostic. The Payload and Topic strings can be developed according to the application developer's
+        perspective.
+
+
+## Discovery Process
+
+Since MQTT bases itself over the TCP/IP stack for networking, the specifications provide _handshaking_ mechanisms which can be leveraged
+to discover IoT Devices that communicate with the MQTT Broker. Since the only mediating entity for message exchange is the MQTT Broker,
+all client only require the IP Address and the MQTT port as a basic form of initiation of communication. Default port for unsecure MQTT
```
kate-goldenring, line 132: nit: all clients
Also, maybe clarify that the clients require the "IP address and MQTT port **of the broker** as a ..."

docs/proposals/mqtt.md
```diff
@@ -0,0 +1,234 @@
...
+Broker is 1883 and for a secure Broker is 8883.
+
+The Akri Discovery Handler hence needs to have the URL Information for the MQTT Brokers and particular topics that need to be subscribed to
+in a configuration file.
```
kate-goldenring, line 136: This is an Akri Configuration right? or some other configuration?

docs/proposals/mqtt.md
```diff
@@ -0,0 +1,234 @@
...
+
+
+## Akri's Custom Broker Interfacing with an MQTT Broker Server
+
+> Note: due to conflicting naming for Akri's Broker as well as MQTT Broker, we refer to Akri's custom broker for MQTT as **MQTT Monitoring Broker**
```
kate-goldenring, line 141: Great note here!

docs/proposals/mqtt.md
```diff
@@ -0,0 +1,234 @@
...
+5. If the IoT Devices already have been publishing data prior to connection of Monitoring Broker with an assumption that they publish their data with QoS 1,
+   the Broker Server would publish all the information to the Monitoring Broker (persistent storage). If IoT Devices are yet to publish on their dedicated
+   topics, the MQTT Broker Server will publish data to the Monitoring Broker once the data is available.
+
+A possible integration of Akri's MQTT Monitoring Broker and MQTT Broker Server looks as follows:
+
+![Akri MQTT Monitoring Broker and MQTT Broker Server System Diagram](../media/akri_mqtt_broker_server.png)
+
+
+## Questions
```
kate-goldenring, line 188: To add to the questions, for each device Akri discovers, a kubernetes resource and an Akri Instance are made. For example, for udev with a filtering rule to only find usb cameras, a resource and Instance would be created for each instance. Then Pods can request these resources. Akri automates the deployment of Pods ("brokers") requesting the resources. What would be a kubernetes resource? (1) each discovered MQTT broker (2) each topic of each MQTT broker (3) MQTT broker with a set of topics (4) something else? How do we make it so a Pod can request the resource(s) that represent "a MQTT broker with a thermometer topic"?

docs/proposals/mqtt.md
```diff
@@ -0,0 +1,234 @@
...
+
+- Would a generic MQTT Monitoring Broker be possible to implement with Akri?
+    - If not, maybe would it be wise to begin with a Specification like [Sparkplug MQTT Specification](https://sparkplug.eclipse.org) where the Topic Structure
+      and the Payload are defined?
+
+- Is gRPC Server required within the MQTT Monitoring Broker? Does the Publish/Subscribe mechanism require a different technology?
+
+- How actually would Discovery Handler Look like? When one talks about Discovery Handler, is this pertaining to only MQTT Broker Server or the discovery of leaf
```
kate-goldenring, line 196: It seems like this should have its own section in the proposal with the following question also seeming to fit in a "What does discovery mean?" section. From previous discussions, it seems like Akri's role is discovering MQTT Broker Servers.

docs/proposals/mqtt.md
```diff
@@ -0,0 +1,234 @@
...
+```yaml
+apiVersion: akri.sh/v0
+kind: Configuration
+metadata:
+  name: mqtt
+spec:
+  discoveryDetails:
+    name: mqtt
+    discoveryDetails: |+
+      serverUrls:
```
kate-goldenring, line 220: What are the server urls? The urls to the MQTT brokers? Is there a way to discover them without being given their addresses? If you are modeling after OPC UA, the server urls can be explicitly provided but ideally that implementation will be expanded to do a network scan instead. Is something like that possible for MQTT?

kate-goldenring (commented):
Thank you for this awesome work. The background and diagrams are really helpful. My main questions that i think could be fleshed out in the proposal are:
What resources should Akri advertise?
Why is this an important scenario? As you pose in the question section, is this only useful on the Edge?
What does discovery look like? Network scan for MQTT brokers?
What language do you think you would implement the discovery handler in?

bfjelds:
It would help me to see some of the Akri concepts identified explicitly.  For example, how does the MQTT Monitoring Broker or MQTT Custom Application relate to Akri's Discovery Handler or Broker concepts (is it one or the other or both)?  

I'd also like a section to identify what existing challenge an Akri-MQTT solution would make easier.  For example, I could imagine a couple:

* Use the Akri-MQTT-Discovery-Handler to determine which nodes to deploy an MQTT kubernetes application to ... this approach would make the MQTT kuberenetes application an Akri broker (and the Akri configuration would likely not create any Akri services)
* Bring MQTT data to Kubernetes applications (applications that do not have MQTT client code in them) ... this approach would make the Akri broker truly a broker, converting the MQTT data stream into something that is consumed through the Akri services.

docs/proposals/mqtt.md
shantanoo-desai: There is no implementation MQTT Broker discovery. The only possible way to achieve this is AFAIK is through [Zeroconf/mDNS/DNS-SD Solution](https://github.com/mqtt/mqtt.org/wiki/broker_auto-discovery). Possible scan on standard ports like 1883 or 8883 might yield results of possible MQTT Brokers on devices however this is not standard practice.

There is however an alternative Protocol called [MQTT for Sensor Networks aka. MQTT-SN](https://www.oasis-open.org/committees/download.php/66091/MQTT-SN_spec_v1.2.pdf) which is an OASIS Standard which tackles the issue of making MQTT-SN more compatible for Low Power Devices in Lossy Networks that do not necessarily use TCP/IP Stack e.g. ZigBee, Bluetooth, LoRa etc.

docs/proposals/mqtt.md
shantanoo-desai: > How do MQTT clients normally respond to topics no longer being published (devices going away)?
There is a possibility of maintaining the connection session between the akri broker (containing mqtt client) however to my knowledge there is no mechanism in MQTT that provides a graceful disconnection if there is no data being published on the subscribed topics. I have to the check the MQTTv5 Specifications where they provide [Session Expiry and Message Expiry](https://www.hivemq.com/blog/mqtt5-essentials-part4-session-and-message-expiry/).
The Session Expiry feature might be able to trigger a graceful disconnection, which in turn leads to bringing down the Akri broker in a much more dynamic manner.

shantanoo-desai:
@kate-goldenring @bfjelds Closing this PR at the moment since I am stuck with some work and will come back to it once I dedicate some time for Akri. If there are some members willing to take this topic, I would gladly assist

kate-goldenring:
> @kate-goldenring @bfjelds Closing this PR at the moment since I am stuck with some work and will come back to it once I dedicate some time for Akri. If there are some members willing to take this topic, I would gladly assist

@shantanoo-desai sounds good. Thanks for the update!

PR #346, opened 2021-07-01
CoAP discovery handler

jiayihu:
**What this PR does / why we need it**: it adds support for the CoAP protocol, which is an important protocol for RESTful IoT devices.

**If applicable**:


deployment/helm/templates/custom-configuration.yaml
```diff
@@ -6,7 +6,19 @@ metadata:
 spec:
   discoveryHandler: 
     name: {{ required "A custom.configuration.discoveryHandlerName is required." .Values.custom.configuration.discoveryHandlerName }}
-    discoveryDetails: {{ .Values.custom.configuration.discoveryDetails }}
```
bfjelds, line 9: for custom-configuration,  i think the intent is that .Values.custom.configuration.discoveryDetails be yaml that can implicitly be what you are describing explicitly below.  if we make the explicit changes you have here, custom-configuration becomes coap-configuration (and can no longer be other configs)

bfjelds: this seems more like the start of a deployment/helm/templates/coap-configuration.yaml :)

kate-goldenring: Also the `custom.configuration` values should be added to a `coap` section of the values file: https://github.com/deislabs/akri/blob/main/deployment/helm/values.yaml. A `coap-discovery-handler.yaml` file should also be added.

Cargo.lock
```diff
@@ -1,5 +1,7 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
+version = 3
```
kate-goldenring, line 3: in case anyone else was curious about where this `version` came from: https://internals.rust-lang.org/t/upcoming-changes-to-cargo-lock/14017

kate-goldenring (commented):
Thanks for this great work @jiayihu! See my comments for my main questions around multicast discovery and async coap client. Also, can you add a coap-configuration.md document under the docs folder to explain installation and the `discoveryDetails` parameters provided for filtering? For example, here is [the one for ONVIF](https://github.com/deislabs/akri/blob/main/docs/onvif-configuration.md). Finally, are you interested in CoAP being able to be compiled into the agent? If so, I can point out where those changes need to happen in a subsequent PR.

kate-goldenring:
@jiayihu can you add some documentation on how to test your discovery handler? Do you happen to have a way you are mocking the coap devices?

docs/proposals/coap.md
```diff
@@ -0,0 +1,79 @@
...
+  ```
+  COAP_IP=192.168.1.126
+  oic.r.temperature=/sensors/temp
+  oic.r.light.brightness=/sensors/light
+  COAP_RESOURCE_TYPES=oic.r.temperature,oic.r.light.brightness
+  ```
+
+  The Broker acts as an HTTP-to-CoAP Proxy. It translates RESTful HTTP requests into RESTful CoAP requests and vice versa for the response. "Cross-Protocol Proxying between CoAP and HTTP" is defined in section 10 of RFC 7252. Currently, the Broker forwards only GET requests.
+  
+  The Broker is also in an excellent position to cache CoAP responses. "Unlike HTTP, the cacheability of CoAP responses does not depend on the request method, but it depends on the Response Code." Currently, the Broker only caches CoAP responses with status code equal to `2.05 Content`, which is returned if the device is not reachable during the connection.
```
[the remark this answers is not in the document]
jiayihu: Screenshot 2021-07-07 at 23 50 02
Sorry what character are you referring to?

docs/proposals/coap.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Outstanding Questions
+
+- Is there a better way to store the discovered resources as Configuration in the cluster?
+
+The current implementation would need a controller to accept queries about available resources and return the name of the Broker's service which can communicate to the device. The device is listed as a generic `akri.sh/coap-021dd7` resource on the node, which is too generic to be useful by any application. A better label would be `akri.sh/oic.r.temperature-021dd7`, the discovered resource. This would allow using the K8s controller for scheduling pods that need the resource.
+
+- Is there any way to avoid coupling a CoAP device to any node?
+
+Akri deploys an agent to each cluster node and each agent has a running CoAP discovery handler. This behaviour results in a single CoAP device being discovered by multiple nodes and listed as multiple virtual CoAP devices. An immediate consequence is that the discovery handlers overflow the device with discovery requests, which causes the device to use more energy and maybe incur in concurrency issues, especially in the case of real-time devices.
```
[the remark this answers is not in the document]
jiayihu: I'll remove this as it is being discussed in #348 and it was probably because I had `shared = false`

docs/proposals/coap.md
```diff
@@ -0,0 +1,79 @@
...
+  
+  The Broker is also in an excellent position to cache CoAP responses. "Unlike HTTP, the cacheability of CoAP responses does not depend on the request method, but it depends on the Response Code." Currently, the Broker only caches CoAP responses with status code equal to `2.05 Content`, which is returned if the device is not reachable during the connection.
+
+## Outstanding Questions
+
+- Is there a better way to store the discovered resources as Configuration in the cluster?
+
+The current implementation would need a controller to accept queries about available resources and return the name of the Broker's service which can communicate to the device. The device is listed as a generic `akri.sh/coap-021dd7` resource on the node, which is too generic to be useful by any application. A better label would be `akri.sh/oic.r.temperature-021dd7`, the discovered resource. This would allow using the K8s controller for scheduling pods that need the resource.
```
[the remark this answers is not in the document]
jiayihu: That is definitely an idea but the number of resource types is quite huge. Besides, as cluster administrator, who deploys the configurations, you may not know them in advance. Imagine you tasked to set up the clusters of a company with several geographical locations. It also means that you need two separate configurations, thus duplicate discovery handlers, for a same device that exposes both temperature and brightness as resources.

With that being said, I like the idea. It should also already be supported by using the `queryFilter` discoveryDetail like this:

```yaml
queryFilter:
  name: "rt" // resource type
  value: "oic.r.temperature"
```

It's part of the CoAP standard that the devices _can_ supported filtering by using a query filter. Even though I've implemented it, I never actually thought about using it for my use case 😄  It remains a workaround local deployments though, IMO.

docs/proposals/coap.md
```diff
@@ -0,0 +1,79 @@
...
+
+Resource Discovery can be performed either unicast or multicast.
+
+When a server's IP address is already known, either a priori or resolved via the Domain Name System (DNS), unicast discovery is performed to locate the entry point to the resource of interest. This specification is achieved using a GET to `/.well-known/core` on the server, which returns a payload in the CoRE Link Format.
+
+Multicast Resource Discovery is useful when a client needs to locate a resource within a limited scope, and that scope supports IP multicast. A GET request to the appropriate multicast address is made for `/.well-known/core`. To limit the number and size of responses, a query string is recommended with the known attributes. Typically, a resource would be discovered based on its Resource Type and/or Interface Description, along with possible application-specific attributes.
+
+### With Akri
+
+The Akri Configuration defines a list of IP addresses to use for resource discovery, implementing thus unicast discovery. Likewise, a multicast IP address can be used for multicast discovery. Both methods can be used at the same time.
```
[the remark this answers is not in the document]
jiayihu: A multicast discovery is just a normal CoAP request sent to the special IP address `224.0.1.187` (in the case of IPv4) and with the `broadcast` flag on the IP packet. CoAP devices should be instructed to reply if they receive that kind of broadcast packet. The discovery handler only supports IPv4 multicast for now. 
Static IP addresses are more for debugging or local clusters where you know which devices are available. For instance I use the static IP of my laptop when during development, the IP of my local embedded device when actually testing the whole system.

discovery-handler-modules/coap-discovery-handler/src/discovery_utils.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub fn new<A: ToSocketAddrs>(addr: A) -> Self {
+        CoAPClientImpl {
+            client: Client::new(addr).unwrap(),
+        }
+    }
+}
+
+impl CoAPClient for CoAPClientImpl {
+    fn get_with_timeout(&self, url: &str, timeout: Duration) -> Result<CoapResponse> {
+        Client::get_with_timeout(url, timeout)
```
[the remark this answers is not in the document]
jiayihu: You're right, I should wrap it into a tokio task and use a separate thread. On the other hand, the `async-coap` crate is basically dead, unfortunately. Maybe one day I'll write an async version myself. Most of the protocol implementation has been already extracted to [coap-lite](https://github.com/martindisch/coap-lite).

discovery-handler-modules/coap-discovery-handler/src/discovery_handler.rs
```diff
@@ -0,0 +1,257 @@
...
+                {
+                    error!(
+                        "discover - for CoAP failed to send discovery response with error {}",
+                        e
+                    );
+                    register_sender.send(()).await.unwrap();
+                    break;
+                }
+
+                delay_for(Duration::from_secs(DISCOVERY_INTERVAL_SECS)).await;
```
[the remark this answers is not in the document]
jiayihu: I just copied over the implementation from you. Make it work first, then improve it 😂
Jokes aside, 10s works in my experience but it's very personal. Less time can result in the device not being able to reply in time since we are dealing with very constrained devices and unstable networks.

> would lack of a loop and rather a polling for new devices be better?

Would you mind expanding your question? By polling, do you mean sending a request and waiting forever for its response like web clients used to do before WebSockets?

Cargo.lock
jiayihu: You have a keen eye for details.

jiayihu:
Sure, I'll add more documentation. To mock the CoAP devices you can use [coap-rs](https://github.com/Covertness/coap-rs) (I personally use [node-coap](https://github.com/mcollina/node-coap) because it's quicker to use an interpreted language for testing). Should I just provide the instructions in a `.md` file or setup an actual project? I think the latter would be better, but I'm not sure if you want something like that within the repo.

discovery-handler-modules/coap-discovery-handler/src/discovery_utils.rs
```diff
@@ -0,0 +1,122 @@
...
+use coap_lite::{CoapRequest, CoapResponse};
+use log::{debug, info};
+#[cfg(test)]
+use mockall::{automock, predicate::*};
+use std::net::{SocketAddr, ToSocketAddrs};
+use std::time::Duration;
+
+type Result<T> = std::io::Result<T>;
+
+#[cfg_attr(test, automock)]
```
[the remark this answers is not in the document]
jiayihu: Fortunately, the old me was wiser and I remembered that I already wrote the tests during the first PR.

docs/proposals/coap.md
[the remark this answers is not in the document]
kate-goldenring: Weird. It shows up on mine: 
image

docs/proposals/coap.md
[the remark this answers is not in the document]
kate-goldenring: what do you mean by ", thus duplicate discovery handlers"? There would still only be one coap DH running on the node but it would be called twice, once for each config

docs/proposals/coap.md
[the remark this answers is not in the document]
kate-goldenring: That sounds great!

discovery-handler-modules/coap-discovery-handler/src/discovery_handler.rs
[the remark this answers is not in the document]
kate-goldenring: I just want to make sure you arent tied to a 10 loop and rather doing whatever form of discovery best fits coap. For example, maybe you maintain the multicast socket and just check if new devices are discovered -- there is the issue of knowing when a device goes offline in this case though.

kate-goldenring:
> Sure, I'll add more documentation. To mock the CoAP devices you can use [coap-rs](https://github.com/Covertness/coap-rs) (I personally use [node-coap](https://github.com/mcollina/node-coap) because it's quicker to use an interpreted language for testing). Should I just provide the instructions in a `.md` file or setup an actual project? I think the latter would be better, but I'm not sure if you want something like that within the repo.

Are you referring to only documentation on mocking the devices? What would go in the "project" you're suggesting? A `coap-configuration.md` file should come with this pr that explains how to configure a Configuration for coap. If you are willing, an end-to-end demo would be fabulous that shows how the configuration, discovery handler, and broker come together, like the ones for opc ua and udev.

kate-goldenring:
@jiayihu give us a bump when you have addressed the comments and think this is ready for another round of reviews.

jiayihu:
@kate-goldenring I think this is finally ready for a second review 🎉

discovery-handler-modules/coap-discovery-handler/src/discovery_utils.rs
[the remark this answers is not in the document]
jiayihu: I went to another approach, i.e. wrapping with `tokio::task::spawn_blocking` the top-level request to discover a single endpoint, instead of making these low-level methods async. I tried the latter approach at first, but I wasn't unable to make it work because `coap::Client` itself doesn't implement `Send`, so making every method that uses it async was very troublesome.

kate-goldenring:
> @kate-goldenring I think this is finally ready for a second review 🎉
> I've seen that you've moved the doc to a separate repo. I can open a separate PR there after you have approved the files here.

Hi @jiayihu, this is great to see! I've been on vacation the past few days, but I'll start taking a look this week! Another PR with docs on the docs site is perfect.

docs/coap-demo.md
```diff
@@ -0,0 +1,165 @@
...
+To inspect more of the elements of Akri:
+
+- Run `kubectl get crd`, and you should see the CRDs listed.
+- Run `kubectl get akric`, and you should see `akri-coap`. 
+- The instances can be seen by running `kubectl get akrii` and
+  further inspected by running `kubectl get akrii akri-coap-<ID> -o yaml`
+
+## Deploy a Kubernetes application which requests the temperature via HTTP
+
+The Akri CoAP built-in Broker Pod is automatically deployed when an instance is created. The Broker exposes a RESTful endpoint which translates HTTP request to CoAP requests. To inspect the associated Kubernetes Service, run `kubectl get crd`.
```
kate-goldenring, line 150: I added a comment to `Values.yaml` about this, but the sample brokers should not be defaults rather than specified in the demos. Can you add a `--set` line to the installation to specify the broker?

docs/coap-demo.md
```diff
@@ -0,0 +1,165 @@
...
+
```
kate-goldenring, line 151: should this be `kubectl get services`?

docs/coap-demo.md
```diff
@@ -0,0 +1,165 @@
...
+akri-coap-920f97-svc   ClusterIP   10.96.190.197   <none>        80/TCP    10m
+akri-coap-svc          ClusterIP   10.96.87.25     <none>        80/TCP    10m
+```
+
+We can now deploy a simple application which just allows to send normal HTTP requests using the terminal.
+
+```
+kubectl run curl --image=radial/busyboxplus:curl -i --tty
+
+[ root@curl:/ ]$ curl http://akri-coap-920f97-svc/sensors/temp
```
kate-goldenring, line 163: specify that the id varies and should be changed when curling

samples/brokers/coap-broker/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+akri-shared = { path = "../../../shared" }
+coap = { git = "https://github.com/Covertness/coap-rs" }
+coap-lite = "0.5"
+env_logger = "0.8.4"
+log = "0.4.14"
+tokio = { version = "1", features = ["full"] }
```
kate-goldenring, line 15: nit: specify specific features to minimize unused code

samples/brokers/coap-broker/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "coap-broker"
+version = "0.1.0"
+authors = ["Jiayi Hu <jiayi.ghu@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+akri-shared = { path = "../../../shared" }
+coap = { git = "https://github.com/Covertness/coap-rs" }
```
kate-goldenring, line 11: specify crate instead of github

docs/coap-configuration.md
```diff
@@ -0,0 +1,83 @@
...
+
+Additional settings can be configured:
+
+| Helm Key | Type | Default | Description |
+|---|---|---|---|
+| coap.configuration.discoveryDetails.multicast | boolean | true | Enable IPv4 multicast discovery | 
+| coap.configuration.discoveryDetails.multicastIpAddress | string | 224.0.1.187 | The IPv4 to which the Discovery Handler sends the packets | 
+| coap.configuration.discoveryDetails.staticIpAddresses | Array of strings | [] | Additional static IP addresses to look for during the discovery. | 
+| coap.configuration.discoveryDetails.queryFilter | { name: string, value: string } | {} | Single name-value pair to filter the discovered resource, as described in [Section 4.1 of RFC 6690](https://datatracker.ietf.org/doc/html/rfc6690#section-4.1) |
```
kate-goldenring, line 39: Maybe break this out into two different fields for clarity, since it seems like the following is the appropriate was to configure the filter:
```
   --set coap.configuration.discoveryDetails.queryFilter.name="rt" \
   --set coap.configuration.discoveryDetails.queryFilter.value="oic.r.temperature"
```

kate-goldenring (commented):
I just finished looking through everything and wow! This is some awesome work. The `coap-demo.md` was smooth. I was able to get through it without any problems. I've made a lot of comments about nits and conventions, but overall this is looking great. One more push and we should be there. Once you have addressed the docs comments, go ahead and put a PR into akri-docs. Another note, maybe run `cargo clippy -p coap-discovery-handler` and `cargo clippy -p coap-broker` to grab any syntax fixes. 

Also, we havent ever hosted programs for mocking devices. I am not sure if `coap-device` needs to exist in the samples. It is really easy to quickly create via the demo. And you could also point to where you are hosting it @jiayihu if you are interested in that.

Cargo.toml
```diff
@@ -8,13 +8,16 @@ members = [
     "shared", 
     "agent", 
     "controller", 
+    "samples/apps/coap-device",
```
kate-goldenring, line 11: I am not sure if this should live in akri as it is a mock device. I feel like the `coap-demo.md` is sufficient for getting people going. Or we could put it on a branch of akri, or you could host it if interested. @bfjelds for thoughts

discovery-handler-modules/coap-discovery-handler/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+async-trait = "0.1.51"
+coap = { git = "https://github.com/Covertness/coap-rs" }
+coap-lite = "0.5"
+env_logger = "0.8.4"
+log = "0.4.14"
+serde = "1.0.104"
+serde_json = "1.0.45"
+serde_derive = "1.0.104"
+serde_yaml = "0.8.17"
+tokio = { version = "0.2", features = ["rt-threaded", "sync", "time", "stream", "fs", "macros", "uds"] }
```
kate-goldenring, line 19: Nit: We recently went to tokio 1.0. Could you bump to that version? It isnt required since it is not embedded in the agent.

deployment/helm/templates/debug-echo-configuration.yaml
```diff
@@ -28,8 +28,6 @@ spec:
           cpu: {{ .Values.debugEcho.configuration.brokerPod.resources.cpuRequest }}
         limits:
           {{`"{{PLACEHOLDER}}"`}} : "1"
-          memory: {{ .Values.debugEcho.configuration.brokerPod.resources.memoryLimit }}
```
bfjelds, line 31: why was this removed?

discovery-handler-modules/coap-discovery-handler/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "coap-discovery-handler"
+version = "0.6.8"
+authors = ["Jiayi Hu <jiayi.ghu@gmail.com>"]
+edition = "2018"
```
bfjelds, line 5: nit: 2018?  should this be 2021?

samples/brokers/coap-broker/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "coap-broker"
+version = "0.1.0"
```
bfjelds, line 3: should the versions match our overall versioning ("0.6.14" ... or better yet "0.7.0")?

bfjelds: and update version.sh for each new Cargo.toml verion location?

bfjelds:
there are a lot of `//TODO` comments.  i wonder if these shouldn't be converted to issues?

jiayihu:
Thank you to both of you for the comments. Unfortunately, vacation has ended and I'm not able to go back to the PR right now 🙃  It might take a while again

romoh:
@jiayihu, there have been a couple of changes since your PR. Let us know if you need help with rebase once you are ready to work on this again., we'll be glad to help.

PR #350, opened 2021-07-19
Make discovery handler map case insensitive

### LINKED ISSUE #344: Make Discovery Handler names non-case sensitive
When connecting a new version for the Debug Echo Discovery Handler the Akri's Agent couldn't recognize the DH since the name of the new version was written as "DebugEcho" instead of "debugEcho".

Make Akri's Agent Discovery Handler map be non-case sensitive and allow both `exampleDH` and `exampleDh` be linked to the same value on RegisteredDiscoveryHandlerMap.

We could make this more descriptive on the discovery handler development documentation, so the new developers are aware when creating a new name for their new DH.

**Additional context**
This is a hard problem to came across today but could became frequent as we develop new Discovery Handlers or new versions for already existing ones.

rodz:
This PR makes the Discovery Handler Map from Akri Agent case insensitive.
Now `RegisteredDiscoveryHandlerMap` is a `CaseInsensitiveHashMap` and allows keys with different capitalizations to be placed in the same map entry, having the same value.

This adds [case_insensitive_hashmap](https://github.com/PhilipDaniels/case-insensitive-hashmap) as a agent dependency. This is a wrapper around the regular Rust HashMap that uses [UniCase](https://docs.rs/unicase/2.6.0/unicase/) as the map key, enabling it to be case insensitive.

closes #344 

**If applicable**:


kate-goldenring:
Thanks @rodz for putting in a fix for this! What were your thoughts around using the crate -- it seems not widely used -- vs creating a wrapper in Akri for the HashMap light in this example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=24bd4bebfb5152cd476ed8c3f5487b72 ?

rodz:
> Thanks @rodz for putting in a fix for this! What were your thoughts around using the crate -- it seems not widely used -- vs creating a wrapper in Akri for the HashMap light in this example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=24bd4bebfb5152cd476ed8c3f5487b72 ?

Good point Kate, I wanted to use the UniCase object while making it case insensitive as it covers more options, for example `Maße` and `MASSE` been matched to the same key(This would not be the case if using the to_lowercase() strategy). This crate seemed like a good wrapper to allow us to do it and still use Strings as keys for querying the Map. 

Your idea of creating a wrapper itself for Akri seams also really good and solves the problem, do you feel like it's better instead of using the crate? As its maintained by a single developer and not widely used as you mentioned.

kate-goldenring:
>  for example Maße and MASSE been matched to the same key(This would not be the case if using the to_lowercase() strategy)

Good point. I am fine going with the crate just wanted to propose another opion

agent/src/util/registration.rs
```diff
@@ -135,7 +136,7 @@ impl Registration for AgentRegistration {
         };
         let mut registered_discovery_handlers = self.registered_discovery_handlers.lock().unwrap();
         // Check if any DiscoveryHandlers have been registered under this name
-        if let Some(register_request_map) = registered_discovery_handlers.get_mut(&dh_name) {
+        if let Some(register_request_map) = registered_discovery_handlers.get_mut(dh_name.clone()) {
```
bfjelds, line 139: why do we need to clone now?

bfjelds:
there is a lot of UniCase usage ... i see a bunch of people doing this in github:

```
    let mut map: HashMap<UniCase<String>, String> = HashMap::new();
    map.insert("asdfASDF".into(), "1".into());
```

not much usage of case_insensitive_hashmap.  are we using case_insensitive_hashmap to avoid changing our code for each insert/get (i.e. the above call to `into`)?

rodz:
> there is a lot of UniCase usage ... i see a bunch of people doing this in github:
> 
> ```
>     let mut map: HashMap<UniCase<String>, String> = HashMap::new();
>     map.insert("asdfASDF".into(), "1".into());
> ```
> 
> not much usage of case_insensitive_hashmap. are we using case_insensitive_hashmap to avoid changing our code for each insert/get (i.e. the above call to `into`)?

That was the original idea but in fact some changes on our inserts had to be done for compatibility. Will try out Kates suggestions to make it more clean.

kate-goldenring:
> Will try out Kates suggestions to make it more clean.

@rodz I think @bfjelds suggestion of making the key `UniCase<String>` is easier than creating a wrapper

rodz:
Updated the PR with a mix of your suggestions. A wrapper around UniCase was created that allowed us to keep accessing the map the same way as previously and still having the case insensitive capabilities.

kate-goldenring:
> Updated the PR with a mix of your suggestions. A wrapper around UniCase was created that allowed us to keep accessing the map the same way as previously and still having the case insensitive capabilities.

I personally think just using `into()` on all `HashMap<UniCase<String>, String>` accesses/modifications is a better solution than the wrapper. While the wrapper results in less changes right now, if someone wants to use other HashMap methods (such as `entry`) then they'd have to add it to the wrapper struct. Plus the compiler will complain if into is not used.

jiria:
Question, why do we not want to enforce consistent naming? Do other K8s resources also allow for case insensitiveness?

rodz:
> Question, why do we not want to enforce consistent naming? Do other K8s resources also allow for case insensitiveness?

Thanks for the question. We are doing it to simplify the scenario with more the one discovery handler for the same protocol are been used. This is about the discovery handler name it passes to the agent while registering so I dont think its much related to the k8s resources itself but rather to a situation where the agent may interact with different DHs depending on the node, for example a wasi krustlet and normal kubelet node.

rodz:
> > Updated the PR with a mix of your suggestions. A wrapper around UniCase was created that allowed us to keep accessing the map the same way as previously and still having the case insensitive capabilities.
> 
> I personally think just using `into()` on all `HashMap<UniCase<String>, String>` accesses/modifications is a better solution than the wrapper. While the wrapper results in less changes right now, if someone wants to use other HashMap methods (such as `entry`) then they'd have to add it to the wrapper struct. Plus the compiler will complain if into is not used.

Good point Kate. I agree with your concerns but having to add the .intro() or other ways to transform to UniCase may bring more friction during development then adding the necessary functions to the struct. Would love to hear more opinions from the rest of the team.

jiria:
> > Question, why do we not want to enforce consistent naming? Do other K8s resources also allow for case insensitiveness?
> 
> Thanks for the question. We are doing it to simplify the scenario with more the one discovery handler for the same protocol are been used. This is about the discovery handler name it passes to the agent while registering so I dont think its much related to the k8s resources itself but rather to a situation where the agent may interact with different DHs depending on the node, for example a wasi krustlet and normal kubelet node.

I might be missing some context, but I am still not sure why we need to add case insensitivity handling just because we have multiple DHs or multiple types of nodes. In either case, would it not be enough just to use case sensitive names?

kate-goldenring:
> > > Question, why do we not want to enforce consistent naming? Do other K8s resources also allow for case insensitiveness?
> > 
> > 
> > Thanks for the question. We are doing it to simplify the scenario with more the one discovery handler for the same protocol are been used. This is about the discovery handler name it passes to the agent while registering so I dont think its much related to the k8s resources itself but rather to a situation where the agent may interact with different DHs depending on the node, for example a wasi krustlet and normal kubelet node.
> 
> I might be missing some context, but I am still not sure why we need to add case insensitivity handling just because we have multiple DHs or multiple types of nodes. In either case, would it not be enough just to use case sensitive names?

Like @jiria, I didn't see the motivation for this change being to handle multiple DH or different types of nodes. Mainly, i see it as a way to improve user experience. This issue https://github.com/deislabs/akri/issues/344 rods ran into would have been prevented. Put the other way, what is the advantage of case sensitive fields in the Akri Configuration? Why should the casing of the name field in the `discoveryDetails` section of a Configuration exactly match the name under which a discovery handler registers?

jiria:
> > > > Question, why do we not want to enforce consistent naming? Do other K8s resources also allow for case insensitiveness?
> > > 
> > > 
> > > Thanks for the question. We are doing it to simplify the scenario with more the one discovery handler for the same protocol are been used. This is about the discovery handler name it passes to the agent while registering so I dont think its much related to the k8s resources itself but rather to a situation where the agent may interact with different DHs depending on the node, for example a wasi krustlet and normal kubelet node.
> > 
> > 
> > I might be missing some context, but I am still not sure why we need to add case insensitivity handling just because we have multiple DHs or multiple types of nodes. In either case, would it not be enough just to use case sensitive names?
> 
> Like @jiria, I didn't see the motivation for this change being to handle multiple DH or different types of nodes. Mainly, i see it as a way to improve user experience. This issue #344 rods ran into would have been prevented. Put the other way, what is the advantage of case sensitive fields in the Akri Configuration? Why should the casing of the name field in the `discoveryDetails` section of a Configuration exactly match the name under which a discovery handler registers?

Was the design not to have the name under discoveryDetails match the registered discovery handlers?

If it was supposed to match, then #344 is not an issue imho. I would rather not add more code, especially not so supported crates, to handle cases like that. As K8s itself is not case insensitive when it comes to resources, I dont see a reason why we should be, especially if it adds more complexity and possibly unsupported code into the project.

Happy to be overruled. And I might be also still missing something.

bfjelds:
i'm not sold on case-insensitivity either.  though my concern is more about it being confusing ... how do we describe where we allow case insensitivity?  case will matter in the Configuration pod spec and service specs.  case will matter in all of the Configuration labels.  with this change, we will ignore case for one field's value.

and then, it will probably mostly matter to the contents of discoveryDetails (for example, if someone uses discoveryDetails.Descriptions for DebugEcho, will it work?).

rodz:
Will be closing this Pull Request based on the discussion. Thanks so much for the feedback and comments.

Talking about the more descriptive error messages I looked more into it, and I think it doesn't fit into Akri now since it's not really an error as if you deploy a Discovery Handler with a different name it will act the same as the others, will be registered and keep waiting for a configuration specifying this protocol to start the discovery process, so there is no errors involved. And if you run Akri with trace log level the Agent will print the current registered DHs list as soon as the discovery process for some protocol start, so the developer can notice the error there if any typos happen.

PR #351, opened 2021-07-19
ONVIF Discovery Handler Optimizations

### LINKED ISSUE #249: ONVIF Camera discovery issue
**Describe the bug**
Unable to discover the ONVIF cameras following the provided installation instructions in the repository.

**Output of `kubectl get pods,akrii,akric -o wide`**
pi@raspberrypi:~ $ kubectl get pods,akrii,akric -o wide
NAME                                              READY   STATUS             RESTARTS   AGE   IP               NODE          NOMINATED NODE   READINESS GATES
pod/akri-controller-deployment-56748f985f-dmmz2   1/1     Running            0          45m   10.42.0.11       raspberrypi   <none>           <none>
pod/akri-video-streaming-app-fc8d97f69-k5q65      1/1     Running            0          31m   10.42.0.12       raspberrypi   <none>           <none>
pod/akri-agent-daemonset-fz7rl                    0/1     CrashLoopBackOff   10         29m   192.168.10.100   raspberrypi   <none>           <none>

NAME                               CAPACITY   AGE
configuration.akri.sh/akri-onvif   2          111m

**Kubernetes Version: [e.g. Native Kubernetes 1.19, MicroK8s 1.19, Minikube 1.19, K3s]**
K3s 1.18.9+k3s1

**To Reproduce**
1. Create cluster using k3s
2. Install Akri with the Helm command
3. Try to discover 2 ONVIF cameras connected to the same local network 
4. Discovery done, but connection failed (log attached)

**Expected behavior**
Cameras connection successful and video streaming works fine

**applicable logs attached**
[akri.log](https://github.com/deislabs/akri/files/6004969/akri.log)

kate-goldenring:
This PR optimizes our ONVIF discovery handler by:
1. Applying scope filtering based on discovery response removing the need for an additional `GetScopes` call to the cameras during `apply_filters`
1. Testing the responsiveness of cameras during discovery via a call to the publicly accessible `GetSystemDateAndTime`. This prevents further filtering calls from being made to unresponsive cameras and ensures that camera is responsive.
1. After initial discovery, maintaining a list of the previously visible cameras and only apply filtering on new cameras that are seen. Previously, every 10 seconds, we were making calls to get the cameras' ip and mac addresses which not only is computationally expensive but also could overload the cameras.

Via some local testing, it appears that the cpu requests and limits can be reduced as a result of the optimizations.  Previously discovery was getting stuck on one thread if a connection could not be made to a camera.

During this PR, also learned that:
- IPv6 cameras with local-link addresses throw an `"Invalid argument"` error due to an interface not being specified. A later PR should address that. For now, this PR fixed the previous behavior of panicking upon this error: closes #249.
- Authenticated cameras do not support `GetNetworkInterfaces` calls, so they are filtered out due to no response. A later PR should add support for obtaining the mac address from authenticated cameras, possibly via arp.
-

**If applicable**:


discovery-handlers/onvif/src/discovery_impl.rs
```diff
@@ -195,156 +196,170 @@ pub mod util {
...
+    async fn get_qualifying_uris_from_discovery_response(
+        discovery_response: &str,
+        scopes: Option<&FilterList>,
+        onvif_query: &impl OnvifQuery,
+    ) -> Vec<String> {
+        let filtered_uris =
+            get_scope_filtered_uris_from_discovery_response(discovery_response, scopes);
+        let mut responsive_uris = Vec::new();
+        for uri in filtered_uris {
+            match onvif_query.is_device_responding(&uri).await {
```
bfjelds, line 215: maybe for the next iteration, i wonder if we couldn't put each of these is_device_responding into a thread and then wait for all the threads to finish.  it seems like if there were 10 cameras and each wasn't responding, we'd have a 10 second loop here (sequentially waiting 1 second for each to timeout).  maybe putting each in its own thread would allow us to wait closer to 1 second total for all the queries?

discovery-handlers/onvif/src/discovery_impl.rs
```diff
@@ -195,156 +196,170 @@ pub mod util {
...
+            );
         }
     }
 
-    pub async fn simple_onvif_discover(timeout: Duration) -> Result<Vec<String>, anyhow::Error> {
-        let (mut discovery_timeout_tx, mut discovery_timeout_rx) = mpsc::channel(2);
-        let (mut discovery_cancel_tx, mut discovery_cancel_rx) = mpsc::channel(2);
-        let shared_devices = Arc::new(Mutex::new(Vec::new()));
-
+    pub async fn get_socket() -> Result<UdpSocket, anyhow::Error> {
```
bfjelds, line 274: `get_discovery_response_socket`?

discovery-handlers/onvif/src/discovery_handler.rs
```diff
@@ -72,42 +72,48 @@ impl DiscoveryHandler for DiscoveryHandlerImpl {
...
+                    if !prev_cameras.contains(c) {
+                        if let Some(device) =
+                            apply_filters(&discovery_handler_config, &c, &onvif_query).await
+                        {
+                            changed_camera_list = true;
+                            filtered_camera_devices.insert(c.clone(), device);
+                        }
+                    }
+                }
+                prev_cameras = latest_cameras;
```
bfjelds, line 111: nit: can `prev_cameras = latest_cameras` be moved into the `if changed_camera_list` scope?

discovery-handlers/onvif/src/discovery_handler.rs
```diff
@@ -72,42 +72,48 @@ impl DiscoveryHandler for DiscoveryHandlerImpl {
...
                     }
                 });
-                if changed_camera_list || matching_camera_count != cameras.len() {
-                    trace!("discover - sending updated device list");
-                    cameras = filtered_onvif_cameras.clone();
+
+                // Insert newly discovered camera that are not filtered out
+                for c in &latest_cameras {
+                    if !prev_cameras.contains(c) {
+                        if let Some(device) =
```
bfjelds, line 103: if we spawned each of the apply_filters calls, would they be more parallel?

bfjelds (approved):

discovery-handlers/onvif/src/discovery_impl.rs
kate-goldenring: Thats a great point. I made changes to `simple_onvif_discover` to first get all filtered uris and then join all the futures that make calls to devices: c55d42feebfea32

discovery-handlers/onvif/src/discovery_handler.rs
kate-goldenring: Fixed in c55d42feebfea32

kate-goldenring:
@bfjelds I made some changes to make the device calls more concurrent, further improving performance. A follow up to this PR should address the verbose logging. trace and debug logging from the ONVIF discover handler dependencies are appearing in the Discovery Handler pod logs.

discovery-handlers/onvif/src/discovery_handler.rs
```diff
@@ -72,42 +72,51 @@ impl DiscoveryHandler for DiscoveryHandlerImpl {
...
                 .await
                 .unwrap();
-                trace!("discover - filtered:{:?}", &filtered_onvif_cameras);
-                let mut changed_camera_list = false;
-                let mut matching_camera_count = 0;
-                filtered_onvif_cameras.iter().for_each(|camera| {
-                    if !cameras.contains(camera) {
+                trace!("discover - discovered:{:?}", &latest_cameras);
+                // Remove cameras that have gone offline
+                previous_cameras.iter().for_each(|c| {
```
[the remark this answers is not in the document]
kate-goldenring: The retry timeout is effectively 10 seconds since that is the discovery interval. If they were to come back online on the next discovery check they will once again be reported to the agent

discovery-handlers/onvif/src/discovery_handler.rs
```diff
@@ -295,18 +258,16 @@ mod tests {
     #[tokio::test]
     async fn test_apply_filters_no_filters() {
         let mock_uri = "device_uri";
+        let mock_ip = "mock.ip";
```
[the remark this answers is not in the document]
kate-goldenring: i think we were aiming at mimicking somewhat the format of an ip address vs a mac address, with `.` and `:` separators, respectively.

discovery-handlers/onvif/src/discovery_handler.rs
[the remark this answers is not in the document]
jiria: So if on second X camera A is present, on second X+9 it disappears, on X+10 another scan does not find it, so it removes it from Akri instance, on X+11 camera comes back (so camera had intermittent connectivity for 2 seconds), we still end up evicting pods that were using the camera or is my understanding wrong?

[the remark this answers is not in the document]
kate-goldenring: Your scenario does/would not occur. The discovery handler logic is different from the agent logic. The Discovery Handler reports to the Agent whenever there are changes in device visibility, checking for changes every 10 seconds. The Agent then decides what to do with that information. Currently, for shared devices, it gives a 5 minute timeout before deleting instances, triggering the controller to bring down pods.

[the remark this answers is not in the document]
kate-goldenring: > Does this logic need some retry timeout before removing the cameras or is that handled elsewhere?

To summarize: the removing of instances and brokers is handled elsewhere, in the Agent and Controller respectively after a 5 minute period of the Discovery Handler not reporting the device.

Issue #324, opened 2021-04-26
Increase dependency versions where not covered by auto-update


kate-goldenring:
Akri's dependencies should be updated. While our auto-update workflow follows semver for updating patches and minor versions where possible, the following crate versions need to be updated:
- [ ] udev to 0.6
- [x] async-stream in Agent
- [x] tonic, tokio, warp (0.3 requires tokio 1.0), futures, prost, hyper which are [interdependent and non-trivial to bump](https://github.com/deislabs/akri/issues/223)
- [x] actix in Webhook 
- [x] yaserde in ONVIF discovery handler

Akri's Python and .NET sample brokers and apps should also be inspected for version updates.

kate-goldenring:
In the udev Discovery Handler, a `Enumerator` trait was created in order to mock the `udev` crate's [`Enumerator` struct](https://docs.rs/udev/0.6.2/udev/struct.Enumerator.html#method.scan_devices). Updating `udev` to 0.6.0 requires changing the `scan_devices` [function definition in the Trait](https://sourcegraph.com/github.com/deislabs/akri/-/blob/discovery-handlers/udev/src/wrappers.rs#L101) to match the [new definition](https://docs.rs/udev/0.6.2/udev/struct.Enumerator.html#method.scan_devices).
FROM
`fn scan_devices(&mut self) -> std::io::Result<udev::Devices>;`
TO
`fn scan_devices(&mut self) -> Result<List<'_, Enumerator, Device>>;`
which requires reconciling lifetimes

kate-goldenring:
We cannot update to the latest versions of kube-rs and k8s-openapi because `kube-rs` requires that all fields of a CRD Spec derive `JsonSchema`; however, `k8s-openapi` structs do not. Consequently, any CRD specs that contain K8s objects (such as `PodSpec` and `ServiceSpec` structs in ours) do no compile. There is an issue up on k8s-openapi to support `JsonSchema` that will need to be resolved before we can proceed with updating our dependencies.  This fix https://github.com/Arnavion/k8s-openapi/pull/104 works as I am able to compile everything by patching `k8s-openapi` with the PR branch.

kate-goldenring:
Closing due to all dependencies being updated besides `udev` which now has its own issue: https://github.com/deislabs/akri/issues/364

PR #361, opened 2021-08-12
Update Akri Kubernetes and Runtime Dependencies

### LINKED ISSUE #223: Update key dependencies: kube-rs, tonic, and tokio
We are currently using an outdated version of `tokio`, `kube-rs`, and `tonic`, and they are all interdependent because the latter two use `tokio`, and there cannot be two versions of `tokio` running in an application or tokio (a runtime for async apps) will error. 
To **get everything to `tokio` 1.0**, `kube-rs` will need to be updated to [v0.47.0+](https://github.com/clux/kube-rs/blob/master/CHANGELOG.md#0470--2021-01-06) and `reqwest` to v0.11.0. 

**This task also provides an opportunity to do pruning of dependencies.** Async in Rust was fairly new when Akri was first being developed, so many crates were interwoven to accomplish tasks. Now, `tokio` may be able to accomplish what we were using `std:::futures` for in some scenarios. 

**I took a look at some of the work that will need to be done by the lucky updater:**
Some changes that will need to be made due to upgrading **tokio**:
- `tokio_core` is now a part of `tokio`, so the new/equivalent functions/structs will need to be used from `tokio`. Our [signal module](https://sourcegraph.com/github.com/deislabs/akri/-/blob/shared/src/os/signal.rs) depends on this. This module may be able to be removed/ changed. Only the udev broker depends on it and its functionality may be uneccesary.
- `time::delay` -> `time::sleep`
- ONVIF [discovery implementation](https://sourcegraph.com/github.com/deislabs/akri/-/blob/agent/src/protocols/onvif/discovery_impl.rs#L229) needs to use `oneshot` instead of `mpsc` channels.

Some changes that will need to be made due to upgrading **kube-rs**:
- ApiClient -> Client
- RawApi -> Api
- Void -> NotUsed
- Considerable changes to their custom resources interface. Looks like they added a `kube_derive` crate that can be utilized from [kube::api::Resource::dynamic](https://github.com/clux/kube-rs/blob/master/CHANGELOG.md#0360--2020-07-19). Looks like they [only support custom resources for K8s >= v1.17](https://github.com/clux/kube-rs/blob/master/CHANGELOG.md#0260--2020-02-25), so this update may limit us from supporting K8s 1.16. See the `crd` prefixed [examples](https://github.com/clux/kube-rs/tree/master/examples).

Some changes that will need to be made due to upgrading **tonic**:
- `prost` (a rust protobuf implementation) will need to be updated, so the code autogenerated from our proto files ([pluginapi.proto](https://sourcegraph.com/github.com/deislabs/akri/-/blob/agent/proto/pluginapi.proto) will change. Then, our [device plugin service implementation](https://sourcegraph.com/github.com/deislabs/akri/-/blob/agent/src/util/device_plugin_service.rs#L126) will need slight modifications to match these changes. 

Other dependencies that will likely need to be updated:
- `hyper`
- `h2` (will need rebase our patch branch)
- `tower`
- `futures` crates (e.g. stream is now a submodule of futures)
- others probably

Some dependencies will probably be able to be removed such as:
- `tokio_core` (now embedded in tokio)
- `futures-old` (only need one futures crate now)

> Note: This task of upgrading dependencies will cause a lot of code changes. It is best to not due this while another large feature change is underway, namely, modularizing discovery protocols https://github.com/deislabs/akri/issues/198.

### LINKED ISSUE #349: RUSTSEC-2021-0073: Conversion from `prost_types::Timestamp` to `SystemTime` can cause an overflow and panic
> Conversion from `prost_types::Timestamp` to `SystemTime` can cause an overflow and panic

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `prost-types`                      |
| Version             | `0.6.1`                   |
| URL                 | [https://github.com/tokio-rs/prost/issues/438](https://github.com/tokio-rs/prost/issues/438) |
| Date                | 2021-07-08                         |
| Patched versions    | `>=0.8.0`                  |

Affected versions of this crate contained a bug in which untrusted input could cause an overflow and panic when converting a `Timestamp` to `SystemTime`.

It is recommended to upgrade to `prost-types` v0.8 and switch the usage of `From&lt;Timestamp&gt; for SystemTime` to `TryFrom&lt;Timestamp&gt; for SystemTime`.

See [#438] for more information.

[#438]: https://github.com/tokio-rs/prost/issues/438

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0073.html) for additional details.

### LINKED ISSUE #358: RUSTSEC-2021-0079: Integer overflow in `hyper`'s parsing of the `Transfer-Encoding` header leads to data loss
> Integer overflow in `hyper`&#39;s parsing of the `Transfer-Encoding` header leads to data loss

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `hyper`                      |
| Version             | `0.13.10`                   |
| URL                 | [https://github.com/hyperium/hyper/security/advisories/GHSA-5h46-h7hh-c6x9](https://github.com/hyperium/hyper/security/advisories/GHSA-5h46-h7hh-c6x9) |
| Date                | 2021-07-07                         |
| Patched versions    | `>=0.14.10`                  |

When decoding chunk sizes that are too large, `hyper`&#39;s code would encounter an integer overflow. Depending on the situation,
this could lead to data loss from an incorrect total size, or in rarer cases, a request smuggling attack.

To be vulnerable, you must be using `hyper` for any HTTP/1 purpose, including as a client or server, and consumers must send
requests or responses that specify a chunk size greater than 18 exabytes. For a possible request smuggling attack to be possible,
any upstream proxies must accept a chunk size greater than 64 bits.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0079.html) for additional details.

### LINKED ISSUE #359: RUSTSEC-2021-0078: Lenient `hyper` header parsing of `Content-Length` could allow request smuggling
> Lenient `hyper` header parsing of `Content-Length` could allow request smuggling

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `hyper`                      |
| Version             | `0.13.10`                   |
| URL                 | [https://github.com/hyperium/hyper/security/advisories/GHSA-f3pg-qwvg-p99c](https://github.com/hyperium/hyper/security/advisories/GHSA-f3pg-qwvg-p99c) |
| Date                | 2021-07-07                         |
| Patched versions    | `>=0.14.10`                  |

`hyper`&#39;s HTTP header parser accepted, according to RFC 7230, illegal contents inside `Content-Length` headers.
Due to this, upstream HTTP proxies that ignore the the header may still forward them along if it chooses to ignore the error.

To be vulnerable, `hyper` must be used as an HTTP/1 server and using an HTTP proxy upstream that ignores the header&#39;s contents
but still forwards it. Due to all the factors that must line up, an attack exploiting this vulnerablity is unlikely.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0078.html) for additional details.

kate-goldenring:
This PR updates our major dependencies. Most importantly, it updates `tokio` to `1.0` and to the latest `kube-rs` and `k8s-openapi` versions. 

- Resolves all of ref #324 excepts for `udev` and any dependencies in the Webhook (See next points)
- The Configuration validating webhook tests failed after updating to latest `kube-rs` and `k8s-openapi`. ~It may be due to the need to possibly regenerate the Configuration OpenAPI API that @DazWilkin generated and hosts [here](https://github.com/DazWilkin/openapi-admission-v1). We could also switch to using the `kube-rs` Admission Controller type now that [they support one](https://github.com/kube-rs/kube-rs/blob/master/examples/admission_controller.rs). For now, the Webhook dependencies are left un-updated and point at the `akri-shared` library from the latest release.~ It was due to tests expecting the `metadata.generation` to be f64; however, the [Kubernetes API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#objectmeta-v1-meta) defines it as an integer, since this field is [incremented every time a CustomResource is modified](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource).
- ~Updates `version.sh` to not update or check the version Akri packages that have a git source (aka akri-shared in the Admission Controller).~
- Fixes `cargo clippy` warnings that had been hanging around. Some still are not fixed, such as functions with too many parameters
- Points to updated h2 patch
- `udev` crate needs more work to update

closes #223

The following will be closed once the Webhook dependencies are also updated.
closes #349
closes #358
closes #359

**If applicable**:


agent/src/util/config_action.rs
```diff
@@ -90,36 +90,35 @@ async fn watch_for_config_changes(
...
-                discovery_handler_map.clone(),
-                new_discovery_handler_sender,
-            )
-            .await?
-        }
+    let resource = Api::<Configuration>::all(kube_interface.get_kube_client());
+    let mut stream = resource
+        .watch(
+            &ListParams::default(),
+            akri_shared::akri::API_VERSION_NUMBER,
```
bfjelds, line 97: curiousity: why move `akri_shared::akri::API_VERSION_NUMBER` out of the use block above?

kate-goldenring: No particular reason. I created `API_VERSION_NUMBER` in this PR and just thought to import it in line since it is only used once here

shared/src/k8s/mod.rs
```diff
@@ -95,133 +88,78 @@ impl OwnershipInfo {
...
-        &self,
-        pod_to_create: &Pod,
-        namespace: &str,
-    ) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>>;
-    async fn remove_pod(
-        &self,
-        pod_to_remove: &str,
-        namespace: &str,
-    ) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>>;
+    async fn find_pods_with_label(&self, selector: &str) -> Result<ObjectList<Pod>, anyhow::Error>;
```
bfjelds, line 95: soooo much love for this simplification

shared/src/akri/configuration.rs
```diff
@@ -36,9 +34,12 @@ pub struct DiscoveryHandlerInfo {
 /// capabilities.  For any specific capability found that is described by this
 /// configuration, an Instance
 /// is created.
-#[derive(Serialize, Deserialize, Clone, Debug)]
+#[derive(CustomResource, Serialize, Deserialize, Clone, Debug, JsonSchema)]
+// group = API_NAMESPACE and version = API_VERSION
+#[kube(group = "akri.sh", version = "v0", kind = "Configuration", namespaced)]
```
bfjelds, line 39: can these use our constants?  or do constants not work in the attributes?

shared/src/akri/configuration.rs
```diff
@@ -4,24 +4,22 @@
...
+    api::{Api, ListParams, ObjectList},
+    client::Client,
 };
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
-pub type KubeAkriConfig = Object<Configuration, Void>;
-pub type KubeAkriConfigList = ObjectList<Object<Configuration, Void>>;
+pub type KubeAkriConfigList = ObjectList<Configuration>;
```
bfjelds, line 18: should this become ConfigurationList?  (it seems like KubeAkriConfig became Configuration)

shared/src/k8s/mod.rs
kate-goldenring: I agree! I think we can put up an issue to change all `Result<(), Box<dyn std::error::Error + Send + Sync + 'static>>` into `anyhow::Result<()>`. I didn't update it everywhere, just where error bubbling up needed to be changed anyways

shared/src/akri/configuration.rs
kate-goldenring: They weren't working which is why I commented it. I'll look into it a little more, though.

bfjelds: looks like no: https://github.com/rust-lang/rust/issues/52393

shared/src/akri/configuration.rs
```diff
@@ -134,53 +122,48 @@ pub async fn get_configurations(
...
 /// ```no_run
 /// use akri_shared::akri::configuration;
-/// use kube::client::APIClient;
+/// use kube::client::Client;
 /// use kube::config;
 ///
 /// # #[tokio::main]
 /// # async fn main() {
-/// let api_client = APIClient::new(config::incluster_config().unwrap());
-/// let dcc = configuration::find_configuration(
```
bfjelds, line 143: lol ... dcc ... devcapconfig :)

agent/src/util/config_action.rs
```diff
@@ -326,8 +323,8 @@ mod config_action_tests {
         let _ = env_logger::builder().is_test(true).try_init();
         let path_to_config = "../test/yaml/config-a.yaml";
         let config_yaml = fs::read_to_string(path_to_config).expect("Unable to read file");
-        let config: KubeAkriConfig = serde_yaml::from_str(&config_yaml).unwrap();
-        let config_name = config.metadata.name.clone();
+        let config: Configuration = serde_yaml::from_str(&config_yaml).unwrap();
+        let config_name = config.metadata.name.clone().unwrap();
```
bfjelds, line 327: why do we need to unwrap the name?

bfjelds: nevermind ... i see it in k8s-openapi: `pub name: Option<String>` ... strange to imagine a resource without a name.

agent/src/util/device_plugin_service.rs
```diff
@@ -115,7 +116,9 @@ impl DevicePlugin for DevicePluginService {
         Ok(Response::new(resp))
     }
 
-    type ListAndWatchStream = mpsc::Receiver<Result<ListAndWatchResponse, Status>>;
+    type ListAndWatchStream =
+        Pin<Box<dyn Stream<Item = Result<ListAndWatchResponse, Status>> + Send + Sync + 'static>>;
+    // type ListAndWatchStream = mpsc::Receiver<Result<ListAndWatchResponse, Status>>;
```
bfjelds, line 121: remove commented line.

( ick ... mpsc::Receiver looks so much more understandable :( )

controller/src/util/node_watcher.rs
```diff
@@ -408,12 +410,16 @@ mod tests {
         let _ = env_logger::builder().is_test(true).try_init();
 
         let node_json = file::read_file_to_string("../test/json/node-b-not-ready.json");
-        let node: NodeObject = serde_json::from_str(&node_json).unwrap();
+        let node: Node = serde_json::from_str(&node_json).unwrap();
         let mut node_watcher = NodeWatcher::new();
+        // Add Node to known_nodes with original Running state
+        node_watcher
```
bfjelds, line 416: why do we need to add a known_node now but didn't before?

controller/src/util/pod_watcher.rs
```diff
@@ -1291,7 +1306,7 @@ mod tests {
             false,
         );
         mock.expect_create_service()
-            .returning(move |_, _| Err(None.ok_or("failure")?));
+            .returning(move |_, _| Err(anyhow::anyhow!("Failure")));
```
bfjelds, line 1309: such an improvement ... i always felt like None.ok_or(...) was cheating a little.

discovery-handlers/onvif/src/discovery_impl.rs
```diff
@@ -183,7 +183,7 @@ pub mod util {
         scopes: Option<&FilterList>,
     ) -> Vec<String> {
         let response_envelope =
-            yaserde::de::from_str::<to_deserialize::Envelope>(&discovery_response);
```
bfjelds, line 186: are the removal of all of these borrowing `&`s coming from clippy?  we were really overusing `&`

shared/src/akri/instance.rs
```diff
@@ -186,48 +168,34 @@ pub async fn find_instance(
...
+    let instances_client: Api<Instance> = Api::namespaced(kube_client.clone(), namespace);
+
+    let mut instance = Instance::new(name, instance_to_create.clone());
+    instance.metadata = ObjectMeta {
+        name: Some(name.to_string()),
+        owner_references: Some(vec![OwnerReference {
+            api_version: format!("{}/{}", API_NAMESPACE, API_VERSION),
+            kind: "Configuration".to_string(),
+            controller: Some(true),
+            block_owner_deletion: Some(true),
```
bfjelds, line 188: do we want to set block_owner_deletion to true?  i'm not sure there is a reason to track ownership if deletion doesn't happen when akri is being uninstalled.

bfjelds: hmm ... nevermind ... i see that you are just matching existing functionality.  i need to refresh on owner refs :)

bfjelds: (had it backwards ... this setting isn't blocking the owner from deleting ... heh ... the other way around)

shared/src/k8s/mod.rs
```diff
@@ -639,22 +535,24 @@ pub async fn try_delete_instance(
...
+                                log::trace!(
+                                    "try_delete_instance - discovered Instance {} already deleted",
+                                    instance_name
+                                );
+                                break;
+                            }
+                            log::error!("try_delete_instance - when looking up Instance {}, got kube API error: {:?}", instance_name, ae);
+                        }
                     }
+                    // Err(e) => {
```
bfjelds, line 553: remove commented code

shared/src/k8s/mod.rs
bfjelds: probably a good 'first issue' type of thing too

bfjelds:
there is sooo much code change in this PR ... thanks for sorting through all the ramifications of updating these!

agent/src/util/config_action.rs
kate-goldenring: Right? `k8s_openapi` updated to put `Option` around almost everything. I dont see a case when a `Configuration` `Pod` etc wouldn't have a name.

controller/src/util/node_watcher.rs
kate-goldenring: Good point. this was left over from when i was trying to use the `watcher` instead of `Api::watch`

discovery-handlers/onvif/src/discovery_impl.rs
kate-goldenring: Yeah, especially in our tests. I didn't even address them all. I ran the following to auto-fix some:
```
rustup override set nightly
cargo clippy --fix -Z unstable-options --allow-dirty
```
Note: it requires nightly to run, so unset the nightly version like so: `rustup override unset`

shared/src/akri/configuration.rs
kate-goldenring: Good point. Fixed in 67595e71132f

agent/src/util/device_plugin_service.rs
kate-goldenring: I agree :( Fixed in 67595e71132f

agent/src/util/discovery_operator.rs
```diff
@@ -125,9 +125,9 @@ impl DiscoveryOperator {
                             self.config.spec.discovery_handler.name
                         );
                         match discovery_handler.discover(discover_request).await {
-                            Ok(device_update_receiver) => {
-                                Some(StreamType::Embedded(device_update_receiver.into_inner()))
-                            }
+                            Ok(device_update_receiver) => Some(StreamType::Embedded(
+                                device_update_receiver.into_inner().into_inner(),
```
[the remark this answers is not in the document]
kate-goldenring: Added in 67595e71132f

controller/src/util/node_watcher.rs
kate-goldenring: removed in 67595e71132f

shared/src/akri/configuration.rs
kate-goldenring: What a gem. I excavated it 67595e71132f :)

agent/src/util/device_plugin_service.rs
bfjelds: it looks like someone got ReceiverStream to compile without the Box/Pin stuff ... might simplify the code?

https://github.com/massbitprotocol/test-ci-for-massbitprotocol/blob/80698ad273eb56d8e168b7c34671292e6e9e7615/chain-reader/chain-reader/src/grpc_stream.rs

agent/Cargo.toml
```diff
@@ -14,46 +14,48 @@ akri-opcua = { path = "../discovery-handlers/opcua", optional = true }
...
+env_logger = "0.9.0"
 futures = { version = "0.3.1", package = "futures" }
 futures-core = "0.3"
 futures-util = "0.3"
-hyper = "0.13.10"
-h2 = "=0.2.6"
-kube = { version = "0.23.0", features = ["openapi"] }
-k8s-openapi = { version = "0.6.0", features = ["v1_16"] }
+hyper = "0.14.2"
+h2 = "0.3.0"
```
romoh, line 27: Lock h2 version so that it won't update when doing cardo update (ie: "=0.3.0")

agent/src/util/device_plugin_service.rs
```diff
@@ -1288,7 +1292,7 @@ mod device_plugin_service_tests {
             .withf(move |name: &str, namespace: &str| {
                 namespace == instance_namespace && name == instance_name
             })
-            .returning(move |_, _| Err(get_kube_not_found_error()));
+            .returning(move |_, _| Err(get_kube_not_found_error().into()));
```
romoh, line 1295: curious, what does this do?

version.txt
```diff
@@ -1 +1 @@
-0.6.12
+0.6.13
```
romoh, line 1: Shouldn't that be a breaking change?

romoh (changes requested):
With the exception of locking the h2 version, the change looks good.

discovery-handlers/onvif/src/discovery_impl.rs
```diff
@@ -372,7 +372,7 @@ pub mod util {
             time::{Duration, SystemTime},
         };
```
bfjelds, line 374: should try_recv_string use try_recv_from now?

bfjelds:
might not be for this PR, but when i run the tests on my work machine i get failures in test_timeout_for_simple_onvif_discover.  it seems to be that i'm finding 4 cameras but get_responsive_uris is only returning for 2 of them, so simple_onvif_discover is taking longer than the expected time.  maybe we should mock OnvifQuery in the test?

version.txt
kate-goldenring: What do you see it breaking? It doesn't modify our CRDs.

agent/src/util/device_plugin_service.rs
kate-goldenring: `find_instance` now returns `anyhow::Error` which implements `From`, converting anything that implements `std::Error` such as `kube::Error` into an `anyhow::Error`. `into` does that conversion

discovery-handlers/onvif/src/discovery_impl.rs
kate-goldenring: Thanks for pointing out that note. I think using the async `recv_from` may be better than `try_recv_from`. It seems preferable to await a response that to poll and loop continuously. If that makes sense to you too, I'll remove the comment

kate-goldenring:
> might not be for this PR, but when i run the tests on my work machine i get failures in test_timeout_for_simple_onvif_discover. it seems to be that i'm finding 4 cameras but get_responsive_uris is only returning for 2 of them, so simple_onvif_discover is taking longer than the expected time. maybe we should mock OnvifQuery in the test?

Interesting. I tested it locally too using `RUST_LOG=trace cargo test -- test_timeout_for_simple_onvif_discover --nocapture` and am just under the time. Like you said, its probably taking too long because the `get_reponsive_uri` call lasts long enough so the next time check is over X seconds. We should add a timeout on that call like we do with `try_recv_string`. That could be a "good first issue" issue.

agent/Cargo.toml
kate-goldenring: Thank you! fixed in d4bab9a

agent/src/util/device_plugin_service.rs
kate-goldenring: good point. fixed in 5dc53e223be4e6a29a6f4310625872d5cbc8215d

kate-goldenring:
I ran `RUST_LOG=trace cargo tarpaulin -v --all-features --out Xml` locally on the same version `0.16.0` and it succeeds. It seems we are running into this bug https://github.com/xd009642/tarpaulin/issues/463, but there doesn't seem to be a clear solution. I am going to try to update the version and see what happens

bfjelds:
> Interesting. I tested it locally too using `RUST_LOG=trace cargo test -- test_timeout_for_simple_onvif_discover --nocapture` and am just under the time. Like you said, its probably taking too long because the `get_reponsive_uri` call lasts long enough so the next time check is over X seconds. We should add a timeout on that call like we do with `try_recv_string`. That could be a "good first issue" issue.

there is something wrong with the test.  no matter what i set the timeout to (it is currently set to 5 seconds, i upped it to 5 minutes), the test takes the same amount of time and at least one of the cameras that is discovered never gets a response from is_device_responding.

bfjelds:
> Interesting. I tested it locally too using `RUST_LOG=trace cargo test -- test_timeout_for_simple_onvif_discover --nocapture` and am just under the time. Like you said, its probably taking too long because the `get_reponsive_uri` call lasts long enough so the next time check is over X seconds. We should add a timeout on that call like we do with `try_recv_string`. That could be a "good first issue" issue.

i think the test has lost its meaning a bit.  previously, the intent was to see if the udp discovery stopped ~2 seconds when the timeout was set to 2.

now, with the timeout set to 2 seconds, for me, the discovery takes ~3.1 seconds (but for you it takes less than 2.2 seconds).  i can make the test pass if i change what `wait_for_call_millis` is set to: `let wait_for_call_millis = 4000;`  i'm not sure that is a good solution though because it seems to depend on how many cameras are actually discovered.

in any case, maybe this is a test improvement that can be made later (or if tests fail for other people).

kate-goldenring:
> Out tarpaulin code coverage tests are failing with:
> 
> ```
>    Compiling tower-http v0.1.1
> error: Broken pipe (os error 32)
> warning: build failed, waiting for other jobs to finish...
> thread 'main' panicked at 'already borrowed: BorrowMutError', src/tools/cargo/src/cargo/util/config/mod.rs:302:20
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> Aug 17 23:13:30.930 ERROR cargo_tarpaulin: Failed to compile tests! Error: the trait bound `ResBody: http_body::Body` is not satisfied
> Error: "Failed to compile tests! Error: the trait bound `ResBody: http_body::Body` is not satisfied"
> ```
> 
> However, when I ran `RUST_LOG=trace cargo tarpaulin -v --all-features --out Xml` locally with the same tarpaulin version `0.16.0`, it succeeds. It seems we are running into this bug [xd009642/tarpaulin#463](https://github.com/xd009642/tarpaulin/issues/463), but there doesn't seem to be a clear solution. I am going to try to update the version and see what happens

increasing the tarpaulin version to the latest seems to have done the trick

bfjelds (approved):

version.txt
romoh: If someone is building their own discovery handler in some repo, doesn't this change require them to update their code as well? Or is the assumption that others doing so will be locked to a specific version?

kate-goldenring: If the discovery handler is running in its own pod (not a part of Akri's repository), then it can run whatever versions it wants.

romoh (approved):

Issue #372, opened 2021-09-02
Instance + Broker Pod disappears


ammmze:
**Describe the bug**
I'm having an issue where my Instance and it is associated Broker pod seem to disappear and won't come back until I either restart my agent and udev discovery pods, or it seems I can simply modify my `Configuration` resource which triggers it to come back.

**Output of `kubectl get pods,akrii,akric -o wide`**
```
NAME                                                 READY   STATUS    RESTARTS   AGE     IP             NODE                 NOMINATED NODE   READINESS GATES
pod/aeotec-gen-5-79e1a1-pod                          1/1     Running   0          9m21s   10.244.0.8     talos-192-168-32-2   <none>           <none>
pod/akri-agent-daemonset-cv95v                       1/1     Running   0          23d     10.244.1.5     talos-192-168-32-3   <none>           <none>
pod/akri-agent-daemonset-kpqzq                       1/1     Running   0          13h     10.244.0.224   talos-192-168-32-2   <none>           <none>
pod/akri-agent-daemonset-pwdj6                       1/1     Running   0          25d     10.244.2.8     talos-192-168-32-4   <none>           <none>
pod/akri-controller-deployment-5496f7c977-p75v6      1/1     Running   0          17h     10.244.2.85    talos-192-168-32-4   <none>           <none>
pod/akri-udev-discovery-daemonset-6kw69              1/1     Running   0          23d     10.244.1.3     talos-192-168-32-3   <none>           <none>
pod/akri-udev-discovery-daemonset-gkhzv              1/1     Running   0          25d     10.244.2.7     talos-192-168-32-4   <none>           <none>
pod/akri-udev-discovery-daemonset-np5dt              1/1     Running   0          13h     10.244.0.223   talos-192-168-32-2   <none>           <none>
pod/chrony-5jqss                                     1/1     Running   0          17d     10.244.0.253   talos-192-168-32-2   <none>           <none>
pod/chrony-5vhn5                                     1/1     Running   0          17d     10.244.2.2     talos-192-168-32-4   <none>           <none>
pod/chrony-94xlv                                     1/1     Running   0          17d     10.244.1.6     talos-192-168-32-3   <none>           <none>
pod/intel-gpu-plugin-2wkdw                           1/1     Running   0          46h     10.244.2.3     talos-192-168-32-4   <none>           <none>
pod/intel-gpu-plugin-rrj4w                           1/1     Running   4          75d     10.244.0.162   talos-192-168-32-2   <none>           <none>
[...]

PR #373, opened 2021-09-04
fix: prevent re-creation instances when only Configuration metadata or status changes

ammmze:
Adds a check when a Modified event comes in for a Configuration resource so that we only perform the instance recreation IF the resource actually changed. This allows things like the metadata and status to be updated, but not trigger recreation. There are tools out there like [Flux](https://fluxcd.io/) which will update the metadata every time it runs reconciliation with a new git commit (regardless of whether the resource spec has actually changed). This in turn previously would cause the instance(s) to be recreated every time Flux reconciliation occurs with a new git commit. 

**What this PR does NOT address**:
It was also determined in #372 that there is likely a race condition that occurs when multiple instances are being recreated at nearly the same time. This in turn causes some of the instances to get lost. While this still needs to be investigated, the changes made in this PR, should make that race condition less likely to present itself...at least in my scenario.

In addition, while I was testing this in my cluster, I had found some oddness with instances and/or broker pods simply not getting created until I restart the controller, agent, and udev discovery pods. I'm not sure what actually got it to create it. Sometimes restarting the controller would get it to go, other times i had to restart all of them. I'm not sure the cause of this yet, but it does not appear to be related to these changes. I swapped the agent back to 0.6.13 and saw the same behavior. Prior to any of these changes I was running 0.6.5, which I don't recall seeing this behavior. I had to upgrade everything to 0.6.13 to be able to test my agent changes.

I'll have to spend some time getting a ubuntu vm setup to be able to run tests and docs. When I try to run these on my local machine, they error out because the `libudev` package is only available for Linux.

**If applicable**:


ammmze:
@kate-goldenring Here's my initial go at this! It looks like I'll need to get a VM setup to be able to run the tests and stuff. Because of that I haven't even attempted to write any tests yet. But feel free to look at what i've got thus far and give some pointers to things that should be done differently. 

I have successfully tested it in my cluster though...

#### logs after flux does reconciling, and updates metadata
```
akri-agent-daemonset-7r65w akri-agent [2021-09-04T05:41:07Z TRACE agent::util::config_action] handle_config - something happened to a configuration
akri-agent-daemonset-7r65w akri-agent [2021-09-04T05:41:07Z INFO  agent::util::config_action] handle_config - modified Configuration Some("aeotec-gen-5")
akri-agent-daemonset-7r65w akri-agent [2021-09-04T05:41:07Z TRACE agent::util::config_action] should_recreate_config - checking if config aeotec-gen-5 needs to be recreated
akri-agent-daemonset-7r65w akri-agent [2021-09-04T05:41:07Z TRACE agent::util::config_action] handle_config - config Some("aeotec-gen-5") has not changed. ignoring config modified event.
```

#### logs after making changes to the Configuration spec
```
akri-agent-daemonset-7r65w akri-agent [2021-09-04T05:43:09Z TRACE agent::util::config_action] handle_config - something happened to a configuration
akri-agent-daemonset-7r65w akri-agent [2021-09-04T05:43:09Z INFO  agent::util::config_action] handle_config - modified Configuration Some("aeotec-gen-5")
akri-agent-daemonset-7r65w akri-agent [2021-09-04T05:43:09Z TRACE agent::util::config_action] should_recreate_config - checking if config aeotec-gen-5 needs to be recreated
akri-agent-daemonset-7r65w akri-agent [2021-09-04T05:43:09Z TRACE agent::util::config_action] handle_config_delete - for config aeotec-gen-5 telling do_periodic_discovery to end
```

agent/src/util/config_action.rs
```diff
@@ -154,6 +158,16 @@ async fn handle_config(
...
+            let do_recreate = should_recreate_config(&config, config_map.clone())
+                .await
+                .unwrap();
+            if !do_recreate {
+                trace!(
+                    "handle_config - config {:?} has not changed. ignoring config modified event.",
+                    config.metadata.name,
+                );
+                return Ok(());
+            }
```
ammmze, line 170: Just realized, I should probably move this above the info log and maybe change my trace log to an info log?

ammmze: Actually, I'll keep my log message a trace, but I think it should be before the info log. The only reason I was considering it an info was because i thought the "something happened to a configuration" log was an info, and I didn't want a user to be left with a message saying that something happened, but then potentially no log saying what happened. The gist is, this message about ignoring the event should be at the same log level (or more verbose) as the "something happened to a configuration" log message.

agent/src/util/config_action.rs
```diff
@@ -276,6 +291,24 @@ async fn handle_config_delete(
     Ok(())
 }
 
+/// Checks to see if the configuration needs to be recreated.
+/// At present, this just checks to see if the `.metadata.generation` has changed.
+/// The `.metadata.generation` value is incremented for all changes, except for changes to `.metadata` or `.status`.
+async fn should_recreate_config(config: &Configuration, config_map: ConfigMap) -> Result<bool, ()> {
+    let name = config.metadata.name.clone().unwrap();
+    let last_generation = config_map.lock().await.get(&name).unwrap().last_generation;
```
ammmze, line 299: There is probably some sort of error handling that should happen around this where we get the config from the map. Still trying to wrap my head around how that works in Rust.

agent/src/util/config_action.rs
```diff
@@ -276,6 +291,24 @@ async fn handle_config_delete(
     Ok(())
 }
 
+/// Checks to see if the configuration needs to be recreated.
+/// At present, this just checks to see if the `.metadata.generation` has changed.
+/// The `.metadata.generation` value is incremented for all changes, except for changes to `.metadata` or `.status`.
+async fn should_recreate_config(config: &Configuration, config_map: ConfigMap) -> Result<bool, ()> {
```
ammmze, line 297: I chose to put this all in its own method so that in the future, if we decided to do something more precise than just looking at the generation, we do that more easily here so the event handler doesn't become too bloated.

kate-goldenring:
`libudev` needs to be installed for the `udev` discovery handler. It's dependency can be installed with
``` 
apt update
apt install -y libudev-dev
```
You may also need `libssl-dev pkg-config libv4l-dev`
We have a script `build/setup.sh` for installing these and Rust.

agent/src/util/config_action.rs
kate-goldenring: It would be best to bubble up this error instead of potentially panicking with an `unwrap`. Maybe return type `Result<bool, anyhow::Error>`  and do something like 
`let last_generation = config_map.lock().await.get(&name).ok_or(anyhow::anyhow!("Configuration {} not found in ConfigMap", name))?.last_generation;`

agent/src/util/config_action.rs
```diff
@@ -341,6 +374,7 @@ mod config_action_tests {
                 stop_discovery_sender,
                 instance_map: instance_map.clone(),
                 finished_discovery_receiver,
+                last_generation: config.metadata.generation,
```
kate-goldenring, line 377: Where is `last_generation` being used in this function? Does it need to be passed in?

agent/src/util/config_action.rs
```diff
@@ -150,6 +154,16 @@ async fn handle_config(
         }
         // If a config is updated, delete all associated instances and device plugins and then recreate them to reflect updated config
         WatchEvent::Modified(config) => {
+            let do_recreate = should_recreate_config(&config, config_map.clone())
```
kate-goldenring, line 157: use `?` operator instead of `unwrap` and panicking here

agent/src/util/config_action.rs
kate-goldenring: Awesome!

agent/src/util/config_action.rs
ammmze: It's not currently being used. I only added it because the rust analyzer gives me the following if I don't add it. Alternatively, I'm sure there's probably a way we can make it optional in the struct?

```
missing structure fields:
- last_generation
```

kate-goldenring: Ooops! The GitHub UI confused me and made me think this was part of a function definition. What you have is correct. Thanks

ammmze: 👍🏻  I've done that way too many times 😀

ammmze:
> > I'll have to spend some time getting a ubuntu vm setup to be able to run tests and docs. When I try to run these on my local machine, they error out because the libudev package is only available for Linux.
> 
> `libudev` needs to be installed for the `udev` discovery handler. It's dependency can be installed with
> 
> ```
> apt update
> apt install -y libudev-dev
> ```
> 
> You may also need `libssl-dev pkg-config libv4l-dev`
> We have a script `build/setup.sh` for installing these and Rust.

Good to know! I'll use these when I go to setup my VM. Actually, I wanna try doing it in the vscode devcontainer thing first and see what happens. Not sure if we've attempted the vscode dev containers before.

agent/src/util/config_action.rs
```diff
@@ -276,6 +291,24 @@ async fn handle_config_delete(
     Ok(())
 }
 
+/// Checks to see if the configuration needs to be recreated.
+/// At present, this just checks to see if the `.metadata.generation` has changed.
+/// The `.metadata.generation` value is incremented for all changes, except for changes to `.metadata` or `.status`.
+async fn should_recreate_config(config: &Configuration, config_map: ConfigMap) -> Result<bool, ()> {
+    let name = config.metadata.name.clone().unwrap();
+    let last_generation = config_map.lock().await.get(&name).unwrap().last_generation;
+    trace!(
```
romoh, line 300: I wonder if it is better to log this along with the result of the function once at the end of the function or maybe in the caller after line 159

ammmze: Yea...TBH we may not even need it. The logging we apply have up in the caller method is probably sufficient. The main reason I put it here, is that it kind of felt like we in general had at least trace logging in most methods to trace that we went into a method.

agent/src/util/config_action.rs
kate-goldenring: That methodology makes sense. We need to do an overhaul of our logging at some point so I understand if the current strategy seems inconsistent. What you have right now is great!

agent/src/util/config_action.rs
ammmze: Looks like clippy doesn't love the ok_or usage...
```
warning: use of `ok_or` followed by a function call
   --> agent/src/util/config_action.rs:304:10
    |
304 |           .ok_or(anyhow::anyhow!(
    |  __________^
305 | |             "Configuration {} not found in ConfigMap",
306 | |             name
307 | |         ))?
    | |__________^
    |
    = note: `#[warn(clippy::or_fun_call)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
help: try this
    |
304 |         .ok_or_else(|| anyhow::anyhow!(
305 |             "Configuration {} not found in ConfigMap",
306 |             name
307 |         ))?
    |
```

Gonna try its suggestion.

ammmze: Looks like it seems happier about that.

agent/src/util/config_action.rs
```diff
@@ -548,4 +584,64 @@ mod config_action_tests {
...
+
+        let config_info = ConfigInfo {
+            instance_map: Arc::new(Mutex::new(HashMap::new())),
+            stop_discovery_sender: stop_discovery_sender.clone(),
+            finished_discovery_receiver,
+            last_generation: Some(1),
+        };
+        let config_name = config.metadata.name.clone().unwrap();
+        let config_map: ConfigMap = Arc::new(Mutex::new(HashMap::new()));
+        config_map.lock().await.insert(config_name, config_info);
```
kate-goldenring, line 607: Nit: Can the shared portion of these two tests be moved to a helper function that both tests call to reuse the code?

ammmze: Yep ... working on that now. Not sure if its just me, but it takes a while to run the tests, build, etc. I suspect it has do to with the fact that I'm using a vscode devcontainer, which runs everything in docker BUT the source files are mounted into the container. And container mounts on Docker Desktop on Mac (and presumably windows) are notoriously slow. Ultimately it's not docker's fault...its the VM's shared folder implementation that is slow.

kate-goldenring: Cargo can certainly take a fair bit of time sometimes as the dependencies directory can really blow up. Sometimes it helps to just run and build the pertinent parts: `cargo build -p agent` or `cargo test <test name>`.

agent/src/util/config_action.rs
```diff
@@ -276,6 +289,28 @@ async fn handle_config_delete(
...
+async fn should_recreate_config(
+    config: &Configuration,
+    config_map: ConfigMap,
+) -> Result<bool, anyhow::Error> {
+    let name = config.metadata.name.clone().unwrap();
+    let last_generation = config_map
+        .lock()
+        .await
+        .get(&name)
+        .ok_or_else(|| anyhow::anyhow!("Configuration {} not found in ConfigMap", name))?
```
bfjelds, line 304: nit curiousity: if you use &name here, do you need the clone on line 299?

agent/src/util/config_action.rs
```diff
@@ -276,6 +289,28 @@ async fn handle_config_delete(
...
+        .last_generation;
+
+    if config.metadata.generation == last_generation {
```
bfjelds, line 307: is there a chance that we'd get updates out of order?  i'm wondering if we wouldn't want <= concept rather than ==?

.gitignore
```diff
@@ -2,3 +2,4 @@
 **/obj
 **/bin
 **/cobertura.xml
+.idea
```
bfjelds, line 5: what is an .idea file?

ammmze: It's the IDE workspace directory created by the IDE by JetBrains ([IntelliJ IDEA](https://www.jetbrains.com/idea/)). I was using that at first before swapping to a vscode dev container so I could run the tests and stuff.

agent/src/util/config_action.rs
ammmze: Good question! I'm not sure TBH. I can make that change though.

agent/src/util/config_action.rs
ammmze: I get this when I attempt to do that:

```
cannot move out of `config.metadata.name` which is behind a shared reference
move occurs because `config.metadata.name` has type `std::option::Option<std::string::String>`, which does not implement the `Copy` traitrustcE0507
config_action.rs(299, 16): consider borrowing the `Option`'s content: `config.metadata.name.as_ref()`
```

```rust
let name = config.metadata.name.unwrap();
let last_generation = config_map
    .lock()
    .await
    .get(&name)
    .ok_or_else(|| anyhow::anyhow!("Configuration {} not found in ConfigMap", &name))?
    .last_generation;
```

ammmze: Looks like I could do this though...

```rust
let name = config.metadata.name.as_ref().unwrap();
let last_generation = config_map
    .lock()
    .await
    .get(name)
    .ok_or_else(|| anyhow::anyhow!("Configuration {} not found in ConfigMap", &name))?
    .last_generation;
```

.gitignore
ammmze: I can remove it if we don't want it. I just wanted to prevent my IDE files from getting committed.

agent/src/util/config_action.rs
ammmze: Gotcha...is there a way to have the `cargo test ...` command only build the agent stuff? It seems like it spends most of the time building everything else.

ammmze: ah...nvm..its same as build. https://doc.rust-lang.org/cargo/commands/cargo-test.html#package-selection

kate-goldenring: ideally the order wouldnt matter!

kate-goldenring (approved):
Thanks for adding these changes and some great tests!

agent/src/util/config_action.rs
ammmze: FYI ... I've made this change so we'll only consider it modified if the generation is greater than the last generation we had seen.

.gitignore
bfjelds: Was just curious...seems fine to leave in :)

bfjelds (approved):

PR #378, opened 2021-09-09
Use `kube-rs` resource `watcher` instead of `Api::watch`

### LINKED ISSUE #374: Use resource `watcher` instead of `Api::watch`
Akri's Agent and Controller watch Pods, Nodes, Instances, and Configuration resources from the Kubernetes API Server. Currently, these watches are done using [the `kube-rs` `Api::watch` function](https://docs.rs/kube/0.60.0/kube/struct.Api.html#method.watch). Now that we have updated to the latest `tokio` versions in #361, we can take advantage of [`kube-rs`'s `watcher` function](https://docs.rs/kube-runtime/0.60.0/kube_runtime/watcher/fn.watcher.html), which, compared to Api::watch, will automatically try to recover the stream upon errors."

Change these to use `watcher`:
[Instance watcher](https://github.com/deislabs/akri/blob/main/controller/src/util/instance_action.rs#L81)
[Node watcher](https://github.com/deislabs/akri/blob/main/controller/src/util/node_watcher.rs)
[Pod watcher](https://github.com/deislabs/akri/blob/main/controller/src/util/pod_watcher.rs)
[Configuration watcher](https://github.com/deislabs/akri/blob/main/agent/src/util/config_action.rs)

**Additional context**
This may help solve https://github.com/deislabs/akri/issues/222 if it can handle disconnects with API server

kate-goldenring:
closes #374 

We updated to using the `Api::watch` object in `kube-rs` `v0.59.0` in https://github.com/deislabs/akri/pull/361, expecting it to work similarly as the informer we were previously using to watch for actions on Pods, Nodes, Instances, Configurations. However, we have experienced different behavior, resulting in ref #372. and using `Api::watch` instead of `kube::runtime::watcher`. I was able to reproduce unexpected behavior locally when using the current implementation. If there are pre-existing configurations, `handle_config_add` is called twice, once from [where we handle pre-existing configs](https://github.com/deislabs/akri/blob/main/agent/src/util/config_action.rs#L50) and the other from the `Api::watch`. `Api::watch` appears to pull previous events before the Agent starts, which is fairly unpredictable. Using `watcher` should resolve this.

~Note: currently not handling restart events other than logging them. This is due to:~
~1. The fact that we handled `WatchEvent::Error` previously just by logging~
~2. A restart event occurs on start up (aka is a start event too). We'd need to know it is the first restart event and handle that differently than others. We could pass a `first_event` `bool` in and error so long as it is not the first event.~

In 19a44f9778208e, made it so the watchers throw an error if they receive a restarted event that is not the initial start event. This will cause the controller or agent containers to error and restart and properly reconcile.

**If applicable**:


agent/src/util/config_action.rs
```diff
@@ -228,7 +227,7 @@ async fn handle_config_delete(
     kube_interface: &impl KubeInterface,
     config: &Configuration,
     config_map: ConfigMap,
-) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+) -> anyhow::Result<()> {
```
bfjelds, line 230: yay!

controller/src/util/instance_action.rs
```diff
@@ -71,68 +72,68 @@ async fn internal_handle_existing_instances(
...
+                info!("handle_instance - watcher started");
+            } else {
+                return Err(anyhow::anyhow!(
+                    "Instance watcher restarted - throwing error"
+                ));
+            }
         }
-        WatchEvent::Bookmark(_) => Ok(()),
     }
+    if *first_event {
```
bfjelds, line 133: nit: not sure what is more expensive, branching logic or setting logic, but it might be simpler to just always set `*first_event=false` here?

bfjelds (approved):

agent/src/util/config_action.rs
```diff
@@ -138,42 +151,28 @@ async fn handle_config(
...
         }
-        WatchEvent::Error(ref e) => {
-            error!("handle_config - error for Configuration: {}", e);
-            Ok(())
+        Event::Restarted(_configs) => {
+            if *first_event {
+                info!("handle_config - watcher started");
+            } else {
+                return Err(anyhow::anyhow!(
+                    "Configuration watcher restarted - throwing error"
```
bfjelds, line 167: maybe say "Configuration watcher restarted - throwing error to restart agent"

controller/src/util/instance_action.rs
kate-goldenring: Good questions. From a quick search i cant find an exploration someone has done on that -- might be fun to test. In the meantime, I think you're right that it makes more sense to just set it.

PR #385, opened 2021-09-23
Make Discovery Handlers check channel health each discovery loop

kate-goldenring:
Akri's Discovery Handlers were designed to only send updates to the Agent if new devices are discovered. This means that if the Agent restarts, the Discovery Handler may not know to register with it again, since it is only aware of Agent restarts when a send errors.  This creates scenarios where an operator may do a `helm upgrade` to modify a configuration, say to filter for certain ONVIF cameras and the change may not be communicated to the Discovery Handler due to the following: The Agent will see the Configuration has changed and close its connection with the discovery handler. The Discovery Handler will not know the connection is closed since devices have not gone offline, so it will not re-register and therefore receive the new filtering information.

To fix this, the Discovery Handlers should check if the channel with the Agent is open at the beginning of each discovery loop. If not, it should re-register with the Agent.

**If applicable**:


discovery-handlers/opcua/src/discovery_handler.rs
```diff
@@ -89,6 +89,15 @@ impl DiscoveryHandler for DiscoveryHandlerImpl {
             let discovery_method = discovery_handler_config.opcua_discovery_method.clone();
             let application_names = discovery_handler_config.application_names.clone();
             loop {
+                // Before each iteration, check if receiver has dropped
```
bfjelds, line 92: does udev need this block of code too?

kate-goldenring: definitely! Thanks. added

bfjelds (approved):

PR #382, opened 2021-09-21
ONVIF Discovery Handler - Handle Multicast Response Duplicates

kate-goldenring:
**What this PR does / why we need it**:
closes #365 

**Special notes for your reviewer**:
Seems like our multicast occasionally returns the same service url twice. Simply deduping the returned ONVIF service URLs seems like a good strategy.

**If applicable**:
- [ ] this PR has an associated PR with documentation in [akri-docs](https://github.com/deislabs/akri-docs)
- [ ] this PR contains unit tests
- [x] added code adheres to standard Rust formatting (`cargo fmt`)
- [x] code builds properly (`cargo build`)
- [x] code is free of common mistakes (`cargo clippy`)
- [x] all Akri tests succeed (`cargo test`)
- [ ] inline documentation builds (`cargo doc`)
- [x] version has been updated appropriately (`./version.sh`)

PR #393, opened 2021-09-30
Fix how duplicates are handled in ONVIF [SAME VERSION]

kate-goldenring:
There is an ONVIF discovery bug caused by this PR https://github.com/deislabs/akri/pull/382, which means that ONVIF discovery doesn't work for Akri versions 0.6.17-0.6.19. Any thoughts on how this should be handled? This is further compounded by the fact that I discovered this after creating the release, so currently latest points to 0.6.19 which means `helm install akri akri-helm-charts/akri` installs the faulty version. 

My proposed fix: push this fix as version 0.6.19 and create new release and in the release CHANGELOG (see #394) put that ONVIF discovery does not work in versions 0.6.17 and 0.6.18 .

Thoughts?

The bug was caused by the fact that we were applying `to_lowercase()` on xml strings to be passed to later parts of discovery and leaving cameras un-discovered. To fix this, this PR removes duplicates on only the URIs returned from `get_scope_filtered_uris_from_discovery_response` is more ideal.
Example of issue: original string:
```
<?xml version="1.0" encoding="UTF-8"?>
    <SOAP-ENV:Envelope xmlns:SOAP-ENV="http://www.w3.org/2003/05/soap-envelope" xmlns:SOAP-ENC="http://www.w3.org/2003/05/soap-encoding" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:wsa="http://schemas.xmlsoap.org/ws/2004/08/addressing" xmlns:wsdd="http://schemas.xmlsoap.org/ws/2005/04/discovery" xmlns:tdn="http://www.onvif.org/ver10/network/wsdl"><SOAP-ENV:Header><wsa:MessageID>urn:uuid:b7b46c8b-16e3-4b77-920e-edd1374a3fe6</wsa:MessageID><wsa:RelatesTo>uuid:b800c7f9-24b5-4817-b87d-446e8fb76e49</wsa:RelatesTo><wsa:ReplyTo SOAP-ENV:mustUnderstand="true"><wsa:Address>http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous</wsa:Address></wsa:ReplyTo><wsa:To SOAP-ENV:mustUnderstand="true">http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous</wsa:To><wsa:Action SOAP-ENV:mustUnderstand="true">http://schemas.xmlsoap.org/ws/2005/04/discovery/ProbeMatches</wsa:Action><wsdd:AppSequence InstanceId="0" MessageNumber="36"></wsdd:AppSequence></SOAP-ENV:Header><SOAP-ENV:Body><wsdd:ProbeMatches><wsdd:ProbeMatch><wsa:EndpointReference><wsa:Address>urn:uuid:fb03dab7-1787-4e12-ab8b-4567327b23c6</wsa:Address></wsa:EndpointReference><wsdd:Types>tdn:NetworkVideoTransmitter</wsdd:Types><wsdd:Scopes>onvif://www.onvif.org/name/Unknown onvif://www.onvif.org/Profile/Streaming</wsdd:Scopes><wsdd:XAddrs>http://10.0.0.4:1000/onvif/device_service</wsdd:XAddrs><wsdd:MetadataVersion>0</wsdd:MetadataVersion></wsdd:ProbeMatch></wsdd:ProbeMatches></SOAP-ENV:Body></SOAP-ENV:Envelope>
```
And after `to_lowercase()`:
```
<?xml version="1.0" encoding="utf-8"?>
    <soap-env:envelope xmlns:soap-env="http://www.w3.org/2003/05/soap-envelope" xmlns:soap-enc="http://www.w3.org/2003/05/soap-encoding" xmlns:xsi="http://www.w3.org/2001/xmlschema-instance" xmlns:xsd="http://www.w3.org/2001/xmlschema" xmlns:wsa="http://schemas.xmlsoap.org/ws/2004/08/addressing" xmlns:wsdd="http://schemas.xmlsoap.org/ws/2005/04/discovery" xmlns:tdn="http://www.onvif.org/ver10/network/wsdl"><soap-env:header><wsa:messageid>urn:uuid:b7b46c8b-16e3-4b77-920e-edd1374a3fe6</wsa:messageid><wsa:relatesto>uuid:b800c7f9-24b5-4817-b87d-446e8fb76e49</wsa:relatesto><wsa:replyto soap-env:mustunderstand="true"><wsa:address>http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous</wsa:address></wsa:replyto><wsa:to soap-env:mustunderstand="true">http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous</wsa:to><wsa:action soap-env:mustunderstand="true">http://schemas.xmlsoap.org/ws/2005/04/discovery/probematches</wsa:action><wsdd:appsequence instanceid="0" messagenumber="36"></wsdd:appsequence></soap-env:header><soap-env:body><wsdd:probematches><wsdd:probematch><wsa:endpointreference><wsa:address>urn:uuid:fb03dab7-1787-4e12-ab8b-4567327b23c6</wsa:address></wsa:endpointreference><wsdd:types>tdn:networkvideotransmitter</wsdd:types><wsdd:scopes>onvif://www.onvif.org/name/unknown onvif://www.onvif.org/profile/streaming</wsdd:scopes><wsdd:xaddrs>http://10.0.0.4:1000/onvif/device_service</wsdd:xaddrs><wsdd:metadataversion>0</wsdd:metadataversion></wsdd:probematch></wsdd:probematches></soap-env:body></soap-env:envelope>
```
**If applicable**:

### DIFF
--- a/discovery-handlers/onvif/src/discovery_impl.rs
+++ b/discovery-handlers/onvif/src/discovery_impl.rs
@@ -134,6 +134,7 @@ pub mod util {
     use super::{common, probe_types, to_deserialize, to_serialize};
     use akri_discovery_utils::filtering::{FilterList, FilterType};
     use log::{error, info, trace};
+    use std::collections::HashSet;
     use std::net::{IpAddr, Ipv4Addr, SocketAddr};
     use tokio::{
         io::ErrorKind,
@@ -202,7 +203,10 @@ pub mod util {
             .collect::<Vec<String>>()
     }
 
-    async fn get_responsive_uris(uris: Vec<String>, onvif_query: &impl OnvifQuery) -> Vec<String> {
+    async fn get_responsive_uris(
+        uris: HashSet<String>,
+        onvif_query: &impl OnvifQuery,
+    ) -> Vec<String> {
         let futures: Vec<_> = uris
             .iter()
             .map(|uri| onvif_query.is_device_responding(uri))
@@ -310,7 +314,7 @@ pub mod util {
         scopes_filters: Option<&FilterList>,
         timeout: Duration,
     ) -> Result<Vec<String>, anyhow::Error> {
-        let mut broadcast_responses = std::collections::HashSet::new();
+        let mut broadcast_responses = Vec::new();
 
         let start = Instant::now();
         loop {
@@ -323,7 +327,7 @@ pub mod util {
 
             match try_recv_string(socket, time_left).await {
                 Ok(s) => {
-                    broadcast_responses.insert(s.to_lowercase());
+                    broadcast_responses.push(s);
                 }
                 Err(e) => match e.kind() {
                     ErrorKind::WouldBlock | ErrorKind::TimedOut => {
@@ -341,7 +345,7 @@ pub mod util {
             "simple_onvif_discover - uris discovered by udp broadcast {:?}",
             broadcast_responses
         );
-        let mut filtered_uris = Vec::new();
+        let mut filtered_uris = std::collections::HashSet::new();
         broadcast_responses.into_iter().for_each(|r| {
             filtered_uris.extend(get_scope_filtered_uris_from_discovery_response(
                 &r,


discovery-handlers/onvif/src/discovery_impl.rs
```diff
@@ -341,7 +345,7 @@ pub mod util {
             "simple_onvif_discover - uris discovered by udp broadcast {:?}",
             broadcast_responses
         );
-        let mut filtered_uris = Vec::new();
+        let mut filtered_uris = std::collections::HashSet::new();
         broadcast_responses.into_iter().for_each(|r| {
             filtered_uris.extend(get_scope_filtered_uris_from_discovery_response(
```
bfjelds, line 350: this seem like a fine solution.  would it make sense to pass filtered_uris into get_scope_filtered_uris_from_discovery_response so that you only execute_filter once for duplicate responses?

bfjelds:
> My proposed fix: push this fix as version 0.6.19 and create new release and in the release CHANGELOG (see #394) put that ONVIF discovery does not work in versions 0.6.17 and 0.6.18 .

this seems fine ... i'm not super worried about versions 6.17 and 6.18 not working.  bugs happen.

discovery-handlers/onvif/src/discovery_impl.rs
kate-goldenring: `filtered_uris` are created by `get_scope_filtered_uris_from_discovery_response` which gets the uris from each discovery response.

romoh:
Is there a way to `yank` a release similar to what [cargo supports](https://doc.rust-lang.org/cargo/reference/publishing.html#cargo-yank) ?
It is reasonable to create a new release with the fix and modify the bad release to add a warning (or label) mentioning the broken scenario.

kate-goldenring:
> Is there a way to `yank` a release similar to what [cargo supports](https://doc.rust-lang.org/cargo/reference/publishing.html#cargo-yank) ? It is reasonable to create a new release with the fix and modify the bad release to add a warning (or label) mentioning the broken scenario.

I think that would definitely be the approach if other commits (changing versions) were added between the release and discovering the bug. Fortunately, since it was discovered quickly, we can effectively overwrite the release

bfjelds (approved):

discovery-handlers/onvif/src/discovery_impl.rs
```diff
@@ -310,7 +314,7 @@ pub mod util {
         scopes_filters: Option<&FilterList>,
         timeout: Duration,
     ) -> Result<Vec<String>, anyhow::Error> {
-        let mut broadcast_responses = std::collections::HashSet::new();
+        let mut broadcast_responses = Vec::new();
```
romoh, line 317: That can be a HashSet too, but agree it doesn't provide much value at this scope.

romoh (approved):

romoh:
> > Is there a way to `yank` a release similar to what [cargo supports](https://doc.rust-lang.org/cargo/reference/publishing.html#cargo-yank) ? It is reasonable to create a new release with the fix and modify the bad release to add a warning (or label) mentioning the broken scenario.
> 
> I think that would definitely be the approach if other commits (changing versions) were added between the release and discovering the bug. Fortunately, since it was discovered quickly, we can effectively overwrite the release
> 
> Not to mention, i already deleted the release. Its that the containers and helm package still exist

Sounds good. yanked releases happen. Glad you found it early on.
On the other side, I'm wondering if we need to add more testing to catch similar issues in the future.

kate-goldenring:
> > > Is there a way to `yank` a release similar to what [cargo supports](https://doc.rust-lang.org/cargo/reference/publishing.html#cargo-yank) ? It is reasonable to create a new release with the fix and modify the bad release to add a warning (or label) mentioning the broken scenario.
> > 
> > 
> > I think that would definitely be the approach if other commits (changing versions) were added between the release and discovering the bug. Fortunately, since it was discovered quickly, we can effectively overwrite the release
> > Not to mention, i already deleted the release. Its that the containers and helm package still exist
> 
> Sounds good. yanked releases happen. Glad you found it early on. On the other side, I'm wondering if we need to add more testing to catch similar issues in the future.

Yeah i wonder if we want to try to set up e2e testing for onvif maybe that only runs on release or can be triggered by a `test-onvif` label

PR #399, opened 2021-10-18
Update code of conduct and add governance document

kate-goldenring:
The PR updates our code of conduct to directly adopt the [Contributor Covenant](https://www.contributor-covenant.org/) Code of Conduct, which both Microsoft and CNCF adopt. It should be adopted directly rather than just linking to (as previously) so as to add direct contacts for CoC violations.

This also adds a governance document that talks about the roles and responsibilities within Akri. Including one is not only important for role clarity in the community but also for our [transition to the CNCF](https://github.com/cncf/sandbox/issues/229)

Context and instructions for open governance: https://github.com/opengovernance/opengovernance.dev

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -117,7 +117,7 @@ the community.
 
 ## Attribution
 
-This Code of Conduct is adapted from the [Contributor Covenant][homepage],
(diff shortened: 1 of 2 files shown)


bfjelds (approved):

kate-goldenring:
@edrickwong how does this look to you?

PR #416, opened 2021-11-01
Helm Agent nodeSelector

### LINKED ISSUE #414: NodeSelector for Agent in Helm
Not all of my nodes need the agent on it. By default every node gets the agent on it. 

I would like to select only the nodes which have the devices that the Akri agent will actually use. 

Simply add `nodeSelector` to the agent section in the helm `values.yaml`. Then apply them to the `nodeSelector` in the template. 

I can patch the chart in the after render handler but it's not ideal.

kferrone:
Adds the ability to select the desired nodes for the agent to run on. This is needed because not every node in a cluster will be an edge node nor have the desired devices. For example maybe only one node has a camera attached. Maybe you have a policy to not run the agent on the control-plane nodes. 

I used these commands to verify it works:

```
helm template foo deployment/helm/ --debug --set agent.nodeSelector.foo=bar --set agent.linuxOnly=false
```
By default linuxOnly is true
```
helm template foo deployment/helm/ --debug --set agent.nodeSelector.foo=bar
```
Try with no nodeSelector, simply creates an empty object which is indeed valid yaml
```
helm template foo deployment/helm/ --debug --set agent.linuxOnly=false
```

**If applicable**:

closes #414


bfjelds:
Very cool!

bfjelds:
bump the version by running `version.sh -u -p`

kate-goldenring:
This is great! Can you also add nodeSelectors to the discovery handlers in case folks are not using the `agent-full`?

kferrone:
@bfjelds I can't run the `version.sh -u -p` command?
It is complaining about `\r` all over the place. Any ideas why this happens? Seems to be Windows vs Linux issues.

kferrone:
@kate-goldenring 
Sure I can add a few more places. I also noticed in some places there is an `s` at the end like `nodeSelectors`. Would it be ok to change to just `nodeSelector`. 

I also was thinking a global/top level `nodeSelector` would be useful.

kate-goldenring:
> @kate-goldenring Sure I can add a few more places. I also noticed in some places there is an `s` at the end like `nodeSelectors`. Would it be ok to change to just `nodeSelector`.
> 
> I also was thinking a global/top level `nodeSelector` would be useful.

yeah it looks like in `values.yaml` the convention is the plural since the value takes 1+ entries. I htink it would make sense to stick to what it is (with `s`) but if singular makes more sense to you then changing it is fine. Keeping it consistent seems most important

kate-goldenring:
> @bfjelds I can't run the `version.sh -u -p` command? It is complaining about `\r` all over the place. Any ideas why this happens? Seems to be Windows vs Linux issues.

Are you running Akri on Windows? Maybe jump into a WSL2 instance if it is giving you trouble. We haven't enabled windows support yet.

kferrone:
I was in wsl when I got the error. It did work with GIT Bash

kferrone:
> This is great! Can you also add nodeSelectors to the discovery handlers in case folks are not using the agent-full?

@kate-goldenring I thought I understood what you meant here. I'm not quite sure what doesn't have them. Seems like all of the discovery-handlers have nodeSelectors. Two of them actually. 
See line 15 and 68: https://github.com/project-akri/akri/blob/main/deployment/helm/templates/udev-discovery-handler.yaml

I'll clean that up to only have one one each in the same manner. 

I also changed what I did to use plural to be consistent.

kferrone:
On another note the DCO issue is really fighting with me. Every time I try and do the rebase there is just a bunch of conflicts.

deployment/helm/templates/agent.yaml
```diff
@@ -17,10 +17,7 @@ spec:
       hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
       {{- end }}
-      {{- if .Values.agent.linuxOnly }}
-      nodeSelector:
-        "kubernetes.io/os": linux
-      {{- end }}
+      nodeSelector: {{ include "akri.agent.nodeSelectors" . | nindent 8 }}
```
kate-goldenring, line 20: I think we still want to maintain the linux-only node selector switch. Can you mimic the [format of the controller deplpoment](https://github.com/project-akri/akri/blob/main/deployment/helm/templates/controller.yaml), adding the linux only selector if specified and then the other selectors?

kate-goldenring:
> > This is great! Can you also add nodeSelectors to the discovery handlers in case folks are not using the agent-full?
> 
> @kate-goldenring I thought I understood what you meant here. I'm not quite sure what doesn't have them. Seems like all of the discovery-handlers have nodeSelectors. Two of them actually. See line 15 and 68: https://github.com/project-akri/akri/blob/main/deployment/helm/templates/udev-discovery-handler.yaml
> 
> I'll clean that up to only have one one each in the same manner.
> 
> I also changed what I did to use plural to be consistent.

Good find. There should definitely only be one. Can you mimic how the [controller adds the both](https://github.com/project-akri/akri/blob/main/deployment/helm/templates/controller.yaml#L65) the linux only flag and the other selectors? It looks like for the discovery handlers, we do not have a linux only value but just make that a constant requirement. I think that is fine for now but definitely inconsistent with the  agent and controller.

kate-goldenring:
> On another note the DCO issue is really fighting with me. Every time I try and do the rebase there is just a bunch of conflicts.

Yeah its a new addition with our joining the CNCF and definitely troublesome at times. It looks like you have 10 commits and some are signed and others are not. 
Do the following commands work?:
```
git rebase HEAD~10 --signoff
git push --force-with-lease origin main
```

kferrone:
Oh no ten of them! I definitely did that wrong then, there should only be five. I'm can just ditch this branch and redo the right way.

kate-goldenring:
> Oh no ten of them! I definitely did that wrong then, there should only be five. I had some annoying conflicts running those commands. Seems easier to just ditch this branch and redo the right way in a new PR.

@kferrone i believe you could retain the same PR. GitHub will pull from whatever is at `kferrone:main` so if you want you could create a new branch locally (say `branch1`) that is configured as you'd like and then `git branch -m branch1 main` then checkout main and force push it

kate-goldenring:
> > Oh no ten of them! I definitely did that wrong then, there should only be five. I had some annoying conflicts running those commands. Seems easier to just ditch this branch and redo the right way in a new PR.
> 
> @kferrone i believe you could retain the same PR. GitHub will pull from whatever is at `kferrone:main` so if you want you could create a new branch locally (say `branch1`) that is configured as you'd like and then `git branch -m branch1 main` then checkout main and force push it

@kferrone, i cloned your branch to play around with it. I think the following should work:
```
git rebase HEAD~6 --signoff -Xtheirs
# when prompted
git commit --allow-empty
```
Then you'll want to fetch the latest from akri main to rebase 
```
git remote add upstream git@github.com:project-akri/akri.git
git fetch upstream
git merge upstream/main --signoff
```
Then update the version again
```
./version.sh -u -p
```

romoh:
@kferrone, are you still working on this?

PR #418, opened 2021-11-12
Install crictl in Agent containers

### LINKED ISSUE #6: Handle crictl query error in slot reconciliation
Currently, we are doing a best effort approach to slot reconciliation, which ensures that device usage on an Instance reflects the real state of which pods are using an instance.
To get the real usage, we are using crictl to query the container runtime in search of active Containers that have our slot Annotations that device-plugin adds to a pod upon an Allocate call from kubelet.
If this crictl query fails (possibly due to crictl not being mounted correctly), we are doing an early return on slot reconciliation. This should be handled more specifically in the future.

Alternatively, if Kubernetes adds Deallocate to Device-Plugin, slot reconciliation might not be needed. This is the PR for Deallocate: kubernetes/kubernetes#91190

kate-goldenring:
Instead of requiring users to install crictl locally, this adds crictl directly to the Agent containers, removing the need to mount it from the host in the charts/templates. 

Many users can forget to specify the path of crictl. fixes #6 

The size of crictl locally is 28.5 MBs:
```
kagold@nuc-ubtunu1:~$ stat /usr/local/bin/crictl
  File: /usr/local/bin/crictl
  Size: 28488749        Blocks: 55648      IO Block: 4096   regular file
Device: fd00h/64768d    Inode: 9440263     Links: 1
Access: (0755/-rwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2021-11-12 00:47:30.575087778 +0000
Modify: 2019-12-16 03:11:36.000000000 +0000
Change: 2021-10-04 21:26:57.523199337 +0000
 Birth: -
```
Similarly, adding crictl to the agent increases its size from 128MB to 161MB.

Associated [documentation](https://docs.akri.sh/user-guide/cluster-setup) should be updated to match this PR.

We could also remove the need to set the container runtime path `agent.host.dockerShimSock` and instead have users specify their distro with something like `kubernetesDistro.microk8s=true`. Then the appropriate socket path will be set. 

We also may want to consider using a more recent version of crictl and add updating it to our release cycle. 

**If applicable**:


build/containers/Dockerfile.agent
```diff
@@ -9,7 +9,16 @@ RUN echo "Using Rust binaries from ${CROSS_BUILD_TARGET}/${BUILD_TYPE}"
 # Link the container to the Akri repository
 LABEL org.opencontainers.image.source https://github.com/project-akri/akri
 
-RUN apt-get update && apt-get install -y --no-install-recommends libssl-dev openssl && apt-get clean
+RUN apt-get update && apt-get install -y --no-install-recommends libssl-dev openssl curl ca-certificates && apt-get clean
+
 COPY ./target/${CROSS_BUILD_TARGET}/${BUILD_TYPE}/agent /agent
+
+# Install crictl
+RUN VERSION="v1.17.0" && \
```
bfjelds, line 17: `VERSION` is set, can it be used in the subsequent commands (curl, tar, rm)?

bfjelds:
is this a patch-level version change or a minor-level change?

bfjelds (approved):

build/containers/Dockerfile.agent
kate-goldenring: Sorry that was a remnant of me trying to do just that. I couldnt get the variable substitution to work like it would in normal bash. I kept getting:
```
/bin/sh: 1: Bad substitution
The command '/bin/sh -c curl -L https://github.com/kubernetes-sigs/cri-tools/releases/download/$VERSION/crictl-${$VERSION}-linux-amd64.tar.gz --output crictl-${$VERSION}-linux-amd64.tar.gz &&    tar zxvf crictl-$VERSION-linux-amd64.tar.gz -C /usr/local/bin &&     rm -f crictl-$VERSION-linux-amd64.tar.gz &&     apt-get remove -y curl ca-certificates && apt-get clean' returned a non-zero code: 2
```
Any idea how to make it work? I also tried with using an `ENV VERSION="v1.17.0"` on the previous line

kate-goldenring: fixed in 879ef84ae58f182de0b3a07c635cf1fee2843e6d key was no brackets or parentheses 🤷

kate-goldenring:
@bfjelds what do you think about:
> We could also remove the need to set the container runtime path agent.host.dockerShimSock and instead have users specify their distro with something like kubernetesDistro.microk8s=true. Then the appropriate socket path will be set.

I was thinking that it would make sense to do this in a subsequent PR so we only have to change our docs once. With just this PR, `AKRI_HELM_CRICTL_CONFIGURATION`  should be changed to something like `CRI_SOCKET` but if we enable setting a K8s distro, it could be an added line of `--set kubernetesDistro=<microk8s|k3s|k8s>` or `kubernetesDistro.<distro>=true`

build/containers/Dockerfile.agent-full
```diff
@@ -9,7 +9,14 @@ RUN echo "Using Rust binaries from ${CROSS_BUILD_TARGET}/${BUILD_TYPE}"
 # Link the container to the Akri repository
 LABEL org.opencontainers.image.source https://github.com/project-akri/akri
 
-RUN apt-get update && apt-get install -y --no-install-recommends libssl-dev openssl && apt-get clean
+RUN apt-get update && apt-get install -y --no-install-recommends libssl-dev openssl curl ca-certificates && apt-get clean
+
+# Install crictl
+RUN curl -L https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.17.0/crictl-v1.17.0-linux-amd64.tar.gz --output crictl-v1.17.0-linux-amd64.tar.gz &&\
```
bfjelds, line 15: use same `VERSION=... && curl $VERSION` style here too?

bfjelds:
> @bfjelds what do you think about:
> 
> > We could also remove the need to set the container runtime path agent.host.dockerShimSock and instead have users specify their distro with something like kubernetesDistro.microk8s=true. Then the appropriate socket path will be set.
> 
> I was thinking that it would make sense to do this in a subsequent PR so we only have to change our docs once. With just this PR, `AKRI_HELM_CRICTL_CONFIGURATION` should be changed to something like `CRI_SOCKET` but if we enable setting a K8s distro, it could be an added line of `--set kubernetesDistro=<microk8s|k3s|k8s>` or `kubernetesDistro.<distro>=true`

how confident are we that we can get the right path based on distro?  if we are really sure, then it sounds like a good idea.  i suppose if someone was using some nonstandard-path, they wouldn't be able to use the helm charts directly anymore, but maybe that never happens?

kate-goldenring:
> > @bfjelds what do you think about:
> > > We could also remove the need to set the container runtime path agent.host.dockerShimSock and instead have users specify their distro with something like kubernetesDistro.microk8s=true. Then the appropriate socket path will be set.
> > 
> > 
> > I was thinking that it would make sense to do this in a subsequent PR so we only have to change our docs once. With just this PR, `AKRI_HELM_CRICTL_CONFIGURATION` should be changed to something like `CRI_SOCKET` but if we enable setting a K8s distro, it could be an added line of `--set kubernetesDistro=<microk8s|k3s|k8s>` or `kubernetesDistro.<distro>=true`
> 
> how confident are we that we can get the right path based on distro? if we are really sure, then it sounds like a good idea. i suppose if someone was using some nonstandard-path, they wouldn't be able to use the helm charts directly anymore, but maybe that never happens?

I think we could still enable someone to override the default. I'll put in a PR of what im thinking after this

build/containers/Dockerfile.agent-full
kate-goldenring: good catch thanks! fixed 1d7e52d97e50f67d174d704e1a769e0d14381e63

build/containers/Dockerfile.agent
jiria, line 17: Why did you pick 1.17? We had problems using older crictl against newer clusters.

build/containers/Dockerfile.agent
```diff
@@ -9,7 +9,16 @@ RUN echo "Using Rust binaries from ${CROSS_BUILD_TARGET}/${BUILD_TYPE}"
...
+    curl -L https://github.com/kubernetes-sigs/cri-tools/releases/download/$VERSION/crictl-$VERSION-linux-amd64.tar.gz --output crictl-$VERSION-linux-amd64.tar.gz && \
+    tar zxvf crictl-$VERSION-linux-amd64.tar.gz -C /usr/local/bin && \
+    rm -f crictl-$VERSION-linux-amd64.tar.gz && \
+    apt-get remove -y curl ca-certificates && apt-get clean
```
jiria, line 21: Can we merge the two RUNs? Otherwise curl/ca-certs will be part of the akri layers and contribute to the overall size.

build/containers/Dockerfile.agent
kate-goldenring: This is the version we have been using in our e2e tests and [documentation](https://docs.akri.sh/user-guide/cluster-setup#configure-crictl). 

>We also may want to consider using a more recent version of crictl and add updating it to our release cycle.

I think our next release we can consider moving to a newer version

build/containers/Dockerfile.agent
kate-goldenring: I condensed them in 75f8467b8652aff8c72dc222ed6fd3c161832eea but it didn't decrease the size at all interestingly.  Still 161 MB

jiria: Interesting. Do we know ca-certs and curl were not in the initial image to begin with or installed as dependencies of the other packages we are installing?

kate-goldenring: I am pretty sure they werent there because it would not install until i added them

jiria (approved):

PR #427, opened 2021-11-23
Set K8s distribution with Helm to simplify choosing container runtime socket

kate-goldenring:
Now that we have removed the need to specify a crictl path https://github.com/project-akri/akri/pull/418, all that needs to be set when installing Akri is the container runtime socket, which is set with `--set agent.host.containerRuntimeSock="/some/sock.sock`. This varies by distribution. To simplify installation and [Kubernetes cluster setup](https://docs.akri.sh/user-guide/cluster-setup), this proposes providing a `kubernetesDistro` value. Simply set `kubernetesDistro.k3s=true`, `kubernetesDistro.k8s=true`, or `kubernetesDistro.microk8s=true`. The template will always preference `agent.host.containerRuntimeSock` if set. Chart throws the following error if no distribution nor socket are set:
`Please set container runtime socket by either selecting the appropriate K8s distro `kubernetesDistro.<k8s|k3s|microk8s>=true` or setting `agent.host.containerRuntimeSock=/container/runtime.sock`. See https://docs.akri.sh/user-guide/cluster-setup for more information."`

**Alternatives**:
If adding `kuberntesDistro`  is more confusing than beneficial, just retain the `agent.host.containerRuntimeSock` value and set it per usual in the [docs](https://docs.akri.sh/user-guide/cluster-setup#configure-crictl), just renaming to be associated with container runtime. These docs will need to be modified to describe setting a distro during installation anyways. This PR could switch focus to changing the name of the `AKRI_HELM_CRICTL_CONFIGURATION` environment variable.

**[Docs to modify](https://docs.akri.sh/user-guide/cluster-setup#configure-crictl)**

**If applicable**:


deployment/helm/values.yaml
```diff
@@ -16,6 +16,13 @@ useDevelopmentContainers: true
 # This can be set from the helm command line using `--set imagePullSecrets[0].name="mysecret"`
 imagePullSecrets: []
 
+# kubernetesDistro describes the Kubernetes distro Akri is running on. It is used to conditionally set
+# distribution specific values such as container runtime socket.
+kubernetesDistro:
+  microk8s: false
```
bfjelds, line 22: i think i'd lean towards using kubernetesDistro==k8s | kubernetesDistro==microk8s | kubernetesDistro==k3s ... rather than kubernetesDistro.microk8s==true, etc.  

seems like maybe a cleaner way to make sure one and only one distro is ever specified.

bfjelds (approved):

romoh (approved):
Should we default to some value? like if not set then k8s is picked.

deployment/helm/values.yaml
kate-goldenring: I was worried about spelling but this makes sense. Changed it all!

kate-goldenring:
> Should we default to some value? like if not set then k8s is picked.

My thought is no. It seems better to error and let them decide their distro. Otherwise slot reconciliation may not work. i know this is different from what we were doing but it would reduce the number of issues folks have with slot reconciliation errors.

kate-goldenring:
Associated documentation: https://github.com/project-akri/akri-docs/pull/19

bfjelds (approved):

deployment/helm/templates/agent.yaml
```diff
@@ -103,7 +103,17 @@ spec:
...
         hostPath:
-          path: "{{ .Values.agent.host.dockerShimSock }}"
+          {{- if ne "" .Values.agent.host.containerRuntimeSocket }}
+          path: {{.Values.agent.host.containerRuntimeSocket }}
+          {{- else if eq .Values.kubernetesDistro "microk8s" }}
+          path: "/var/snap/microk8s/common/run/containerd.sock"
+          {{- else if eq .Values.kubernetesDistro "k3s" }}
+          path: "/run/k3s/containerd/containerd.sock"
+          {{- else if eq .Values.kubernetesDistro "k8s" }}
+          path: "/var/run/dockershim.sock"
```
jiria, line 113: What if folks use containerd instead of dockershim?

jiria:
Please update the PR message with the updated design.

deployment/helm/templates/agent.yaml
kate-goldenring: then they can set `agent.host.containerRuntimeSocket` to whatever they want

kate-goldenring: The change in documentation should clarify that https://github.com/project-akri/akri-docs/pull/19

jiria: Would it make sense to then use k8s-dockershim to reduce confusion? Otherwise folks with containerd might be surprised why it is not working. Just FYI, AKS has transitioned from dockershim to containerd, the future lies with containerd, not dockershim.

kate-goldenring: in that case i think it would be better to do the alternative i describe in the PR description. Namely, we would provide no `kubernetesDistro=$DISTRO_NAME` and instead recommend how to set `agent.host.containerRuntimeSocket` based on the distro. I think that would be better than providing multiple options per distro (k8s-containerd`, `k8s-dockershim`, etc). @bfjelds and @romoh on thoughts?

kate-goldenring: @jiria @romoh @bfjelds bump here. Would be nice to get the documentation in one this once we decide on what to do here

bfjelds: i would say have distro=k8s|k3s|etc ... then use whatever the distro default is (dockershim or containerd) ... and document that anything that doesn't follow the defaults can be supported iwth containerRuntimeSocket.  though i also have no problem with k8s-dockershim|k8s-containerd|etc.

kate-goldenring: I am going with brian's last comment and sticking with what we have here. If folks complain about wanting our k8s default changed, we can change it. And I'll make sure the documentation points out how to set it to containerd.

kate-goldenring:
@romoh @bfjelds instead of erroring out, I set k8s' docker as the default. I could not throw a warning with helm, but i commented in the chart manifest that the default is being used: 161cdfd

deployment/helm/templates/agent.yaml
jiria: Sounds good to me.

jiria (approved):

PR #437, opened 2021-12-10
Support for Deploying Terminating Pods as Brokers

kate-goldenring:
This adds support for deploying Jobs to devices discovered by the Akri Controller. Previously, Akri only supported deploying Pods that never terminated. Adding Jobs will enable more scenarios with Akri such as device management scenarios. It follows the design in this proposal https://github.com/project-akri/akri-docs/pull/17. Some specific changes to call out:

- The Configuration CRD: (**Breaking change**) The `brokerPodSpec` section of the Configuration changed to a `brokerType` enum with a `brokerPodSpec` or `brokerJobSpec` options.
- The Controller: Logic to deploy a Pod or Job depending on the `brokerType` of the Configuration. ~Logic to handle the `brokerType` of a Configuration being updated -- will bring down and redeploy workloads and services.~
- The Agent: Logic to not bring down a Configuration if the brokerType changed.
- Helm Charts: In rbac.yaml, extends Controller's ClusterRole to enable it to create, delete, modify Jobs. Enables providing a JobSpec in Configurations, where `completions`, `parallelism`, and `backoffLimit` are configurable. All other types of JobSpec customizations will need to be done by creation a Configuration yaml and modifying it. Instructions for this can be added to our docs, [just as they are for PodSpecs](https://docs.akri.sh/user-guide/customizing-an-akri-installation#modifying-the-brokerpodspec). Removed default image for debugEcho -- [docs](https://docs.akri.sh/development/debugging) will need to be updated for this.

**Docs**
I put in a PR with docs updating Pod deployment in tandem with this one: https://github.com/project-akri/akri-docs/pull/21
It also contains documentation on Jobs.

**Tests**
This PR includes unit tests. Should we extend out e2e workflows to test Jobs? Maybe we just test a subset of our platform/k8s distro matrix for jobs to prevent too many runs. If so, can that be tabled to a later pr so as to not overload this one?

I went ahead and built and published some containers for folks to try out this implementation. You can run this on this [Kubernetes playground](https://www.katacoda.com/courses/kubernetes/playground) if you do not have a cluster:
1. Clone and checkout this branch. Then package the helm chart locally:
   ```
   git clone git@github.com:kate-goldenring/akri.git --branch strategy-jobs akri-jobs
   helm package akri-jobs/deployment/helm
   ```
2. Install Akri with a debugEcho Configuration that will deploy a Job to each (`foo0` and `foo1`) discovered device. The Job will echo "Hello World" and then sleep for 5 seconds before terminating.
   ```
   helm install akri akri-0.8.0.tgz \
    --set agent.allowDebugEcho=true \
    --set debugEcho.discovery.enabled=true \
    --set debugEcho.discovery.image.tag=latest \
    --set controller.image.repository=ghcr.io/kate-goldenring/job-brokers/controller \
    --set controller.image.tag=v0.8.0-20220112_182137-amd64 \
    --set agent.image.repository=ghcr.io/kate-goldenring/job-brokers/agent \
    --set agent.image.tag=v0.8.0-20220107_184416-amd64 \
   --set debugEcho.configuration.brokerJob.image.repository=busybox \
   --set debugEcho.configuration.brokerJob.command[0]="sh" \
   --set debugEcho.configuration.brokerJob.command[1]="-c" \
   --set debugEcho.configuration.brokerJob.command[2]="echo 'Hello World'" \
   --set debugEcho.configuration.brokerJob.command[3]="sleep 5" \
   --set debugEcho.configuration.enabled=true
   ```
 
3. Say you are feeling more exuberant and want the Job to echo "Hello Amazing World", you can change the `brokerSpec`. Upgrade the installation to do so and watch Akri delete all the resources (instances, jobs) and recreate them, deploying the new job (Do `kubectl logs` on the Pods to verify).

    ```
    helm upgrade akri akri-0.8.0.tgz \
    --set agent.allowDebugEcho=true \
    --set debugEcho.discovery.enabled=true \
    --set debugEcho.discovery.image.tag=latest \
    --set controller.image.repository=ghcr.io/kate-goldenring/job-brokers/controller \
    --set controller.image.tag=v0.8.0-20220112_182137-amd64 \
    --set agent.image.repository=ghcr.io/kate-goldenring/job-brokers/agent \
    --set agent.image.tag=v0.8.0-20220107_184416-amd64 \
   --set debugEcho.configuration.brokerJob.image.repository=busybox \
   --set debugEcho.configuration.brokerJob.command[0]="sh" \
   --set debugEcho.configuration.brokerJob.command[1]="-c" \
   --set debugEcho.configuration.brokerJob.command[2]="echo 'Hello Amazing World'" \
   --set debugEcho.configuration.brokerJob.command[3]="sleep 5" \
   --set debugEcho.configuration.enabled=true
   ```

4. Test out some of the Job settings and modify the Job to run the Job twice per device (`completions=2`) in parallel (`parallelism=2`). These Helm settings simply modify equivalently named parts of the [Kubernetes JobSpec](https://kubernetes.io/docs/concepts/workloads/controllers/job/)
   
   ```
   helm upgrade akri akri-0.8.0.tgz \
    --set agent.allowDebugEcho=true \
    --set debugEcho.discovery.enabled=true \
    --set debugEcho.discovery.image.tag=latest \
    --set controller.image.repository=ghcr.io/kate-goldenring/job-brokers/controller \
    --set controller.image.tag=v0.8.0-20220112_182137-amd64 \
    --set agent.image.repository=ghcr.io/kate-goldenring/job-brokers/agent \
    --set agent.image.tag=v0.8.0-20220107_184416-amd64 \
   --set debugEcho.configuration.brokerJob.image.repository=busybox \
   --set debugEcho.configuration.brokerJob.command[0]="sh" \
   --set debugEcho.configuration.brokerJob.command[1]="-c" \
   --set debugEcho.configuration.brokerJob.command[2]="echo 'Hello Amazing World'" \
   --set debugEcho.configuration.brokerJob.command[3]="sleep 5" \
   --set debugEcho.configuration.brokerJob.parallelism=2 \
   --set debugEcho.configuration.brokerJob.completions=2 \
   --set debugEcho.configuration.enabled=true
   ```
5. Pods are still deployed the same per usual and they can also be updated.
   ```
   helm install akri akri-0.8.0.tgz \
    --set agent.allowDebugEcho=true \
    --set debugEcho.discovery.enabled=true \
    --set debugEcho.discovery.image.tag=latest \
    --set controller.image.repository=ghcr.io/kate-goldenring/job-brokers/controller \
    --set controller.image.tag=v0.8.0-20220112_182137-amd64 \
    --set agent.image.repository=ghcr.io/kate-goldenring/job-brokers/agent \
    --set agent.image.tag=v0.8.0-20220107_184416-amd64 \
   --set debugEcho.configuration.brokerPod.image.repository=nginx \
   --set debugEcho.configuration.enabled=true
   ```

**Note on future extension**
Many deployment scenarios can be fulfilled with Jobs just by adjusting the `parallelism` (number of the same Pod that should run concurrently) and `completions` (number of times the Pods of the Job need to successfully complete). However, an additional scenario could be enabled by adding support in the controller for adding `podAntiAffinity` (by instance name). 
For example, you could specify that you want 3 Jobs running concurrently (`parallelism=3` and `completions=3`) but that you want at most 1 on each node. Limiting one to run on each node can be done by setting `podAntiAffinity` in the JobSpec like so:
```yaml
  spec:
    template:
      metadata:
        labels:
          akri.sh/instance: akri-instance-ha5h
      spec:
        containers:
        - name: pi
          image: perl
          command: ["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]
        restartPolicy: Never
        affinity:
          podAntiAffinity:
            requiredDuringSchedulingIgnoredDuringExecution:
            - labelSelector:
                matchExpressions:
                - key: akri.sh/instance
                  operator: In
                  values:
                  - akri-instance-ha5h
              topologyKey: "kubernetes.io/hostname"
```
This would need to be done by the controller rather than in the Helm charts as it is likely desirable to set `podAntiAffinity` per device/instance rather than per Configuration.

**If applicable**:


shared/src/akri/configuration.rs
```diff
@@ -28,6 +29,21 @@ pub struct DiscoveryHandlerInfo {
...
+/// See this clippy warning for more details:
+/// https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
+#[derive(Serialize, Deserialize, PartialEq, Clone, Debug, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum BrokerSpec {
+    // PodSpec for Pod that should be scheduled to all nodes that can
+    // access any capability described by this Configuration
+    BrokerPodSpec(Box<PodSpec>),
+    // JobSpec for Job that should be deployed to each capability described by this Configuration
+    BrokerJobSpec(Box<JobSpec>),
```
bfjelds, line 44: nit: no need to prefix BrokerSpec objects with 'Broker' ... maybe `JobSpec` and `PodSpec` are enough?

agent/src/util/slot_reconciliation.rs
```diff
@@ -143,7 +143,11 @@ impl DevicePluginSlotReconciler {
                 .as_ref()
                 .unwrap_or(&Vec::new())
                 .iter()
-                .any(|condition| condition.type_ == "ContainersReady" && condition.status != "True")
+                .any(|condition| {
+                    condition.type_ == "ContainersReady"
+                        && condition.status != "True"
+                        && condition.reason != Some("PodCompleted".to_string())
```
bfjelds, line 149: is there any way this `condition.reason != Some("PodCompleted".to_string()` could be applied to a broker PodSpec (rather than a JobSpec)?

job.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: batch/v1
```
bfjelds, line 1: is this file needed?

shared/src/akri/configuration.rs
kate-goldenring: I added the prefix in case we run into issue with templating for OpenAPI or other specs.

agent/src/util/slot_reconciliation.rs
kate-goldenring: Yes. Which should be fine. This code is simply telling slot reconciliation that there are no pending Pods (Pods that are neither ready nor completed).

job.yaml
kate-goldenring: no thanks will remove

jiria:
nit on description: Previously, Akri only supported deploying Jobs that never terminated.

should this say pods?

jiria:
Agreed we should add e2e tests for jobs and can be done separately.

jiria:
When a job spec is changed, do you kill old jobs and schedule new jobs?

jiria:
Would be good to figure out how to break these big prs into several smaller ones. No action needed for this one.

kate-goldenring:
> Would be good to figure out how to break these big prs into several smaller ones. No action needed for this one.

One way would be to make the Configuration CRD change in a first one. And then add the implementation in another.

agent/src/util/config_action.rs
```diff
@@ -382,6 +387,71 @@ mod config_action_tests {
         .is_err());
     }
 
+    // Tests that a Configuration is not recreated if ONLY the BrokerSpec of the ConfigurationSpec has increased.
```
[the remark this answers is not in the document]
kate-goldenring: sorry should be "changed" can modify

kate-goldenring:
> When a job spec is changed, do you kill old jobs and schedule new jobs?
> 
> edit: I think I found the answer to this is yes. Do we need to rethink this? What is a firmware update of a device is in progress and you kill the job? You might brick the device that is being flashed.

Good point. One thing we could do is never delete a job -- let them stick around as "Completed" (which may be useful anyways for logs purposes) -- and jsut deploy the new job and add an anti-affinity that will not run it so long as another one is running. Not sure if one exists because podantiaffinity may just result in the new job being scheduled to a different node.

controller/src/util/config_action.rs
```diff
@@ -0,0 +1,429 @@
...
+                    "Configuration watcher restarted - throwing error to restart agent"
+                ));
+            }
+        }
+    }
+    *first_event = false;
+    Ok(())
+}
+
+/// This function is called when the Configuration::BrokerSpec was modified where
```
bfjelds, line 174: nit: not sure i understand this sentence ... should "where BrokerSpec::BrokerPodSpec" be removed?

kate-goldenring: its trying to say that its called when the BrokerSpec is type BrokerPodSpec. i can reword

shared/src/akri/configuration_state.rs
```diff
@@ -0,0 +1,141 @@
...
+    match config
+        .metadata
+        .generation
+        .unwrap()
+        .cmp(&config_state.last_generation.unwrap())
+    {
+        Ordering::Less => {
+            error!(
+                "should_recreate_config - configuration generation somehow went backwards {:?} < {:?}",
+                config.metadata.generation, config.metadata.generation
```
bfjelds, line 31: nit: for this error message, should one of these be config_state.last_generation?

webhooks/validating/configuration/src/main.rs
```diff
@@ -299,23 +299,25 @@ mod tests {
...
-                    "brokerPodSpec": {
-                        "containers": [
-                            {
-                                "image": "image",
-                                "name": "name",
-                                "resources": {
-                                    "limits": {
-                                        "{{PLACEHOLDER}}": "1"
+                    "brokerSpec": {
+                        "brokerPodSpec": {
```
bfjelds, line 303: should there be tests for brokerJobSpec?

bfjelds: and a failure test if brokerPodSpec and brokerJobSpec are defined (assuming that isn't allowed)?

controller/src/util/pod_watcher.rs
```diff
@@ -317,16 +331,18 @@ impl BrokerPodWatcher {
...
 
-        // Make sure instance has required Pods
-        if let Ok(instance) = kube_interface.find_instance(&instance_id, namespace).await {
-            super::instance_action::handle_instance_change(
-                &instance,
-                &super::instance_action::InstanceAction::Update,
-                kube_interface,
-            )
-            .await?;
+        // Only redeploy Pods that are managed by the Akri Controller (have Instance ownerReference)
```
bfjelds, line 334: i'm a little confused here ... what happens here if !instance_owned?  does get_instance_and_configuration_from_pod return successfully if the pod is not instance_owned?  what does it return in that case?

shared/src/akri/configuration.rs
```diff
@@ -258,23 +278,25 @@ mod crd_serialization_tests {
...
                 },
-                "brokerPodSpec": {
-                    "containers": [
-                        {
-                            "image": "nginx:latest",
-                            "name": "usb-camera-broker",
-                            "resources": {
-                                "limits": {
-                                    "{{PLACEHOLDER}}": "1"
+                "brokerSpec": {
```
bfjelds, line 281: should there be some tests with brokerJobSpec?

controller/src/util/config_action.rs
```diff
@@ -0,0 +1,429 @@
...
+            match get_entry(config.metadata.name.as_ref().unwrap(), config_map.clone()).await {
+                Some(prev_config_state) => match &config.spec.broker_spec {
+                    None => {
+                        config_map.lock().await.insert(
+                            config.metadata.name.as_ref().unwrap().to_string(),
+                            prev_config_state,
+                        );
+                    }
+                    Some(broker) => {
+                        if !should_recreate_config(&config, &prev_config_state) {
```
bfjelds, line 113: sorry, it has been a while since i've looked at this code ... should_recreate_config is confusing to me.

here we are only acting if it returns false.  it seems like we used to do nothing if it returned false.

the name is also confusing to me.  are we actually recreating a Configuration?  or are we just executing code **as though** the Configuration is deleted and recreated?

kate-goldenring:
> > When a job spec is changed, do you kill old jobs and schedule new jobs?
> > edit: I think I found the answer to this is yes. Do we need to rethink this? What is a firmware update of a device is in progress and you kill the job? You might brick the device that is being flashed.
> 
> Good point. One thing we could do is never delete a job -- let them stick around as "Completed" (which may be useful anyways for logs purposes) -- and jsut deploy the new job and add an anti-affinity that will not run it so long as another Job is running.

@jiria anti-affinity cannot be set in a JobSpec; however, we could set it in the podspecs as illustrated in the "Note on future extension" section of the PR description above. However, that would restrict any two pods of a job from running on a node, which seems okay, but there may be a scenario (doesnt seem likely though) where someone may want two identical pods 
 of a job (parallelism=2+) to run on the same node at once

webhooks/validating/configuration/src/main.rs
kate-goldenring: For 1: Yes there should. I was going to save that for a later PR. Something that i should have documented in the PR description.
For 2: The configuration definition should disallow the definition of both brokerPodSpec and brokerJobSpec as the `brokerSpec` property [is an enum](https://github.com/kate-goldenring/akri/blob/strategy-jobs/shared/src/akri/configuration.rs#L39).

controller/src/util/pod_watcher.rs
kate-goldenring: Yes it returns successfully. For terminating pods, this call to handle_instance_change makes sure that pods are re-deployed if terminated. For jobs, that check isnt necessary and would hit [this logging line](https://github.com/kate-goldenring/akri/blob/strategy-jobs/controller/src/util/instance_action.rs#L478) and do nothing. This basically stops unnecessary code from running. It assumes that if an instance resource is being use by an owner other than Akri's controller (ie a Job) then we don't have further action to do when a pod stops running.

shared/src/akri/configuration.rs
kate-goldenring: Sure i can add one, but what would this check beyond verifying that serde works properly?

controller/src/util/config_action.rs
kate-goldenring: i agree the name is unhelpful. It works with the agent's context but not the controller. It may be best to rename it something like `broker_changed` or more specifically `only_broker_changed`. This check is done by the agent upon a config change. If true, it does not recreate the configuration rather lets the change stand (no further action). The controller does this check, and if true, it redeploys the brokers to reflect the new config. This is our first step at more explicit configuration modification handling. For this PR, we could leave this out and stick to the old way of deleting and reapplying a configuration, but it means instances will be brought down and redeployed. We need to figure out how to handle this gracefully and this was my first stab at it. it does mean though that if the controller goes down for some reason, a configuration may be applied that does not reflect the system (since the controller was not there to redeploy the brokers)

deployment/helm/templates/custom-configuration.yaml
```diff
@@ -7,26 +7,61 @@ spec:
...
-  {{- if .Values.custom.configuration.brokerPod.image.repository }}
-  brokerPodSpec:
-    containers:
-    - name: {{ .Values.custom.configuration.name }}-broker
-      image: {{ printf "%s:%s" .Values.custom.configuration.brokerPod.image.repository .Values.custom.configuration.brokerPod.image.tag | quote }}
-      {{- with .Values.custom.configuration.pullPolicy }}
-      imagePullPolicy: {{ . }}
+  {{- if or .Values.custom.configuration.brokerPod.image.repository .Values.custom.configuration.brokerJob.image.repository }}
+  {{- /* Only add brokerSpec if a broker image is provided */}}
+  brokerSpec:
```
romoh, line 12: so current template doesn't prevent a user from specifying both a podspec and a jobspec, which I believe is not supported. correct?
We may want to add that to the validating webhook.

controller/src/util/instance_action.rs
```diff
@@ -398,17 +371,127 @@ pub async fn handle_instance_change(
...
             anyhow::anyhow!("Namespace not found for instance: {}", &instance_name)
         })?;
+    let configuration = match kube_interface
+        .find_configuration(&instance.spec.configuration_name, instance_namespace)
+        .await
+    {
+        Ok(config) => config,
+        _ => {
+            if action != &InstanceAction::Remove {
+                // In this scenario, a configuration has been deleted without a Akri Agent deleting the associated Instances.
```
romoh, line 386: nit nit: remove "a"

shared/src/k8s/job.rs
```diff
@@ -0,0 +1,571 @@
...
+///
+/// Example:
+///
+/// ```no_run
+/// use akri_shared::k8s::{
+///     OwnershipInfo,
+///     OwnershipType,
+///     job
+/// };
+/// use akri_shared::akri::instance::{Instance, InstanceSpec};
```
romoh, line 83: Is the sample code in the comment left on purpose?

shared/src/k8s/job.rs
```diff
@@ -0,0 +1,571 @@
...
+    ownership: OwnershipInfo,
+    resource_limit_name: &str,
+    job_spec: &JobSpec,
+    app_name: &str,
+) -> anyhow::Result<Job> {
+    trace!("create_new_job_from_spec enter");
+    // TODO: Consider optionally enabling podAntiAffinity in this function
+    // (using an instance name label) to ensure only one Job runs on each Node per instance.
+    let instance_name = instance.metadata.name.as_ref().unwrap();
+    let mut labels: BTreeMap<String, String> = BTreeMap::new();
```
romoh, line 121: Curious, why are we using a BTree and not a HashMap?

deployment/helm/templates/custom-configuration.yaml
kate-goldenring: I believe it does prevent it. The configuration definition should disallow the definition of both brokerPodSpec and brokerJobSpec as the brokerSpec property [is an enum](https://github.com/kate-goldenring/akri/blob/strategy-jobs/shared/src/akri/configuration.rs#L39).

shared/src/k8s/job.rs
kate-goldenring: Yes, I provided examples for all functions in jobs.rs just like we do with pods.rs and instance.rs, etc

shared/src/k8s/job.rs
kate-goldenring: Because that that is the [type for `labels` defined by k8s_openapi](https://docs.rs/k8s-openapi/latest/k8s_openapi/apimachinery/pkg/apis/meta/v1/struct.ObjectMeta.html#structfield.labels) (Rust crate for Kubernetes types)

controller/src/util/config_action.rs
kate-goldenring: @bfjelds @jiria @romoh on this note, i am thinking that at least for this PR, we should remove support for handling broker changes and stick to our old approach of the agent delete and reapplying modified configurations. We can then spec out and propose separately how we want to handle changes and cooperate between the agent and controller

shared/src/akri/configuration.rs
bfjelds: i would phrase it differently ... it checks that we've asked serde to do what we think we did

controller/src/util/pod_watcher.rs
bfjelds: ok, so BrokerJobSpec pods aren't owned ... but have akri bits in the set of labels.  that wasn't clear to me before.

webhooks/validating/configuration/src/main.rs
bfjelds: my understanding is that the configuration definition will simply fail to parse ... in the same way as if a user created a yaml where a string was passed to Capacity.  something like this is what the webhook exists to detect, right?  and we should have a test that validates that.

kate-goldenring: agreed. on both. just wanted to clarify that even without the webhook, kubernetes API would catch 2 but we need the webhook to catch 1. Is it okay to save this for a later PR?

controller/src/util/pod_watcher.rs
kate-goldenring: Yes all pods are owned by the Job. But we own the job

shared/src/akri/configuration.rs
kate-goldenring: okay can add!

webhooks/validating/configuration/src/main.rs
bfjelds: i don't have any idea if the kubernetes API would catch it ... or how it would handle it.  would it ignore one of them?  would an error happen?  i don't know ... part of that is why i would prefer a test exist.

controller/src/util/pod_watcher.rs
bfjelds: "we own the job" ... does that mean the instance owns the job?  and the job owns the pod?

kate-goldenring: yes exactly. sorry for not being more specific

bfjelds: i think i might change some of the variable names and functions to try to clarify that somehow.

maybe `fn instance_owned -> bool` can be `fn get_broker_pod_type_from_ownership -> BrokerPodType` where BrokerPodType is an enum: [BrokerPod, BrokerJob, Unknown]

kate-goldenring:
@romoh @jiria @bfjelds to reduce the size of this PR, I have removed the Configuration watcher from the Controller -- motivated by [this note](https://github.com/project-akri/akri/pull/437#discussion_r780387772). This not only was the section that was the most confusing (Controller and Agent coordinating configuraiton recreation, handling letting jobs finish before deleting them, managing configuration state) but it is also hard to justify. It was added only to enable modifying a brokerSpec without bringing down and recreating instances. Since instances are still directly tied to a Configuration, this seems unneccesary. I have updated the PR description -- the examples point to newly built containers. There should now be much less to review. I will update the [proposal](https://github.com/project-akri/akri-docs/pull/17) as well. Gracefully handling Configuration changes is a discussion that we can leave for another proposal and PR.

deployment/helm/templates/custom-configuration.yaml
kate-goldenring: oh i see what you're saying. The helm templates could construct a config with both a PodSpec and JobSpec. It would fail to be applied to the cluster because it wouldnt match the CRD definition, though

shared/src/akri/configuration.rs
kate-goldenring: added in 6e1e8817ba6a34b99f3

controller/src/util/config_action.rs
jiria: All in favor of splitting into smaller prs.

Another thought on how to handle controller restarts, we could cache the config that was used to create a broker as part of broker annotation. So then controller could check if the broker is based on the latest configuration or not.

jiria:
> > > When a job spec is changed, do you kill old jobs and schedule new jobs?
> > > edit: I think I found the answer to this is yes. Do we need to rethink this? What is a firmware update of a device is in progress and you kill the job? You might brick the device that is being flashed.
> > 
> > 
> > Good point. One thing we could do is never delete a job -- let them stick around as "Completed" (which may be useful anyways for logs purposes) -- and jsut deploy the new job and add an anti-affinity that will not run it so long as another Job is running.
> 
> @jiria anti-affinity cannot be set in a JobSpec; however, we could set it in the podspecs as illustrated in the "Note on future extension" section of the PR description above. However, that would restrict any two pods of a job from running on a node, which seems okay, but there may be a scenario (doesnt seem likely though) where someone may want two identical pods of a job (parallelism=2+) to run on the same node at once

Why do we need the anti-affinity in this case?

Re cleanup, imho some cleanup might be useful, but perhaps not killing the running jobs.

Also, from another pov, we might not want to have two firmware updates running at the same time. Is there a good way for the two jobs to signal each other?

kate-goldenring:
> > > > When a job spec is changed, do you kill old jobs and schedule new jobs?
> > > > edit: I think I found the answer to this is yes. Do we need to rethink this? What is a firmware update of a device is in progress and you kill the job? You might brick the device that is being flashed.
> > > 
> > > 
> > > Good point. One thing we could do is never delete a job -- let them stick around as "Completed" (which may be useful anyways for logs purposes) -- and jsut deploy the new job and add an anti-affinity that will not run it so long as another Job is running.
> > 
> > 
> > @jiria anti-affinity cannot be set in a JobSpec; however, we could set it in the podspecs as illustrated in the "Note on future extension" section of the PR description above. However, that would restrict any two pods of a job from running on a node, which seems okay, but there may be a scenario (doesnt seem likely though) where someone may want two identical pods of a job (parallelism=2+) to run on the same node at once
> 
> Why do we need the anti-affinity in this case?
> 
> Re cleanup, imho some cleanup might be useful, but perhaps not killing the running jobs.
> 
> Also, from another pov, we might not want to have two firmware updates running at the same time. Is there a good way for the two jobs to signal each other?

@jiria per [this comment and the changes associated with it](https://github.com/project-akri/akri/pull/437#issuecomment-1007720309), I removed all gracefully handling of Configuration changes from this PR. As this thread shows, we have plenty to discuss here. For now, they will be deleted immediately upon Configuration changes just like pods are. We dont do anything graceful from them currently either. See [this diagram](https://github.com/project-akri/akri-docs/blob/af874807f55368285f66e90fc9519d409b115d69/media/controller-with-jobs-flow.png) from the proposal PR for a visual. As to your questions, "anti-affinity" was the signalling i was hoping for. aka i was looking for  a way for a job to know another job is running and therefore not run. Anti-affinity cannot do that for jobs. I agree that we should think this through for firmware update settings and make a plan.

deployment/helm/templates/custom-configuration.yaml
kate-goldenring: added tests https://github.com/project-akri/akri/pull/437/commits/198ca4f365232f9006b364c4f49d0716587345d7

controller/src/util/config_action.rs
kate-goldenring: Good point. That can also done simply via the Job name. I am using the Configuration generation in the Job name. However, someone could feasibly delete a configuration and reapply a new one while the Controller is down and then the Controller would wrongly think it has the right state.

controller/src/util/pod_watcher.rs
kate-goldenring: I see your point. One thing that linguistically confuses me is how Jobs are a management object that deploy Pods. So in the Pod Watcher, we are reacting to Pods being created by the Job. Is it fair to call that Pod a BrokerJob?

kate-goldenring: Maybe just `get_broker_type_from_ownership` (removing pod) would help

bfjelds: `get_broker_type_from_ownership` works for me

bfjelds: for me, this will help clarify the if statements ... the intent/meaning of `if type == BrokerPod {...}` is much clearer to me than `if instance_owned {...}`

webhooks/validating/configuration/src/main.rs
```diff
@@ -521,6 +401,113 @@ mod tests {
...
+        ]
+    }
+    "#;
+
+    // Valid JSON but invalid akri.sh/v0/Configuration when inserted into
+    // brokerSpec of ADMISSION_REVIEW OR EXTENDED_ADMISSION_REVIEW constants.
+    // Misplaced `resources`
+    //   Valid: .request.object.spec.brokerSpec.brokerPodSpec.containers[*].resources
+    // Invalid: .request.object.spec.brokerSpec.brokerPodSpec.resources
+    const INVALID_BROKER_POD_SPEC: &str = r#"
```
bfjelds, line 430: i think it woud be good to have an invalid test where both job and pod specs are defined.

kate-goldenring: I had the wrong thinking around this. Currently, our CRD does not specify `oneOf` with regards to `brokerSpec`. It seems that we cannot use the  [`oneOf`](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema) structural schema specifier since we need to set `x-kubernetes-preserve-unknown-fields: true`. We should do more investigation here to see if we can limit in our Configuration CRD specifying both types of brokers -- that way the error isnt caught by our application and they error out. For now, here is the test that asserts that serde will catch this and error out the Agent: 20422dc

kate-goldenring:
@romoh @jiria this should be ready for a final look

bfjelds (approved):

PR #442, opened 2022-01-23
fix all clippy errors and update versions

### LINKED ISSUE #383: Fix `cargo clippy` warnings
[`cargo clippy`](https://github.com/rust-lang/rust-clippy) is a linter that suggest Rust code improvements. We currently check all PRs for clippy errors before merging in our [Check Rust workflow](https://github.com/deislabs/akri/blob/4ad5743182e9d3af53828a298ccbe0e05810585c/.github/workflows/check-rust.yml#L48); however, this does not seem to prevent merging code that adds more clippy warnings.

All clippy warnings should be addressed:
`cargo clippy --all`

Furthermore, clippy errors in tests should be addressed
`cargo clippy --all-targets --all-features -- -D warnings`

Some of these may be able to be addressed via `cargo clippy --fix`

Also, our [Check Rust workflow](https://github.com/deislabs/akri/blob/4ad5743182e9d3af53828a298ccbe0e05810585c/.github/workflows/check-rust.yml) could be updated to run `cargo clippy --all-targets --all-features -- -D warnings` instead of `cargo clippy --all` so as to always check test code and disallow merging code that creates clippy warnings

Ragnyll:
closes #383 
This fixes all current clippy lint errors and adds ci checks for clippy.

#383 mentions ["adding a clippy.toml file" for addressing `function has too many arguments`](https://github.com/project-akri/akri/issues/383#issuecomment-932543354) but i elected not to add that in interest of `#[allow(clippy::too_many_arguments)]` instead. This makes those functions with too many args the exception rather than a rule. If these need to be refactored into some kind of struct later than it may be worth logging an issue specifically for that. 

**If applicable**:


agent/src/util/device_plugin_builder.rs
```diff
@@ -140,8 +140,8 @@ impl DevicePluginBuilderInterface for DevicePluginBuilder {
                     UnixListener::bind(task_socket_path).expect("Failed to bind to socket path");
 
                 async_stream::stream! {
-                    while let item = uds.accept().map_ok(|(st, _)| unix_stream::UnixStream(st)).await {
```
Ragnyll, line 143: i was a little confused handling these. The way this is written it looks like the loop is trying to break the while when there's no more connections to accept ( i think). but according to clippy this is an irrefutable statement that never breaks.
```
warning: irrefutable `while let` pattern
   --> discovery-utils/src/discovery/mod.rs:231:27
    |
231 |                     while let item = uds.accept().map_ok(|(st, _)| unix_stream::UnixStream(st)).await {
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(irrefutable_let_patterns)]` on by default
    = note: this pattern will always match, so the loop will never exit
    = help: consider instead using a `loop { ... }` with a `let` inside it
```
but matching this and yielding the unix stream doesnt work because this fn returns a Result. 

So ultimately i am a little concerned if this is the desired functionality. I trusted clippy and thought id make a note of this.

Ragnyll: i dont think this causes an infinite loop though. Otherwise any test that was using the [test version of this](https://github.com/project-akri/akri/pull/442/files#diff-87f938518167c58192faa87e9f179c26b5da52546181289fdbd13db1cd36f819R274) would have failed.

kate-goldenring:
@Ragnyll, thanks for picking this up! You're approach of using `#[allow(clippy::too_many_arguments)]` flags instead of a `clippy.toml` file is super sound -- it also makes it clear exactly which functions need to be updated in the future. I'll take a look at your comments tomorrow but this is looking great.

agent/src/util/device_plugin_builder.rs
kate-goldenring: This loop looks fine -- it almost matches the [tonic example](https://github.com/hyperium/tonic/blob/v0.6.2/examples/src/uds/server.rs#L54). The item from the stream could potentially be stored in an `item` variable [like in the example](https://github.com/hyperium/tonic/blob/v0.6.2/examples/src/uds/server.rs#L54) for better readability.

agent/src/util/device_plugin_service.rs
```diff
@@ -894,6 +894,7 @@ mod device_plugin_service_tests {
             .collect();
         assert_eq!(devices.len(), capacity);
         // Can't use map on Device type
+        #[allow(clippy::needless_collect)]
```
kate-goldenring, line 897: Might be optimal to refactor instead of adding flags where possible/not extensive.
I believe the following will work without the flag:
```
        let mut device_ids = devices.into_iter().map(|device| device.id);
        for device in expected_device_ids {
            assert!(device_ids.any(|d| d == device));
        }
```

kate-goldenring (changes requested):
Just one nit to avoid the `needless_collect` flag and this looks good to go. Thanks for all this work!

agent/src/util/device_plugin_builder.rs
Ragnyll: Gotcha, i see what you mean by the tonic example too.
refactored all these to be 
```
let item = ...
yield item;
```
in f333e781c98158c8a462f341b8d002e8de76a37c

agent/src/util/device_plugin_service.rs
Ragnyll: sorry @kate-goldenring thats I dont think gonna work. The `any` consumes the iterator (both into_iter and iter which i find odd, because `iter()` was supposed to be non consuming I thought). Here's an example if you wanna see what I mean: [gist](https://gist.github.com/Ragnyll/f77d24acb98a647b7d64c92c9b10e9d0) (this made me think i was going crazy for a sec).  So if the two sets of devices are not in the same order it'll eat up the non matching ones before it starts matching.

anyways reconstructing the iterator for each any in the loop works, and clippy doesnt complain: 8f48d04017fb60e55c4d90804072d8b533565666

if you dont like that i'd say just go back to the version with the `allow needless collect`,  this new way is kinda weird, so i'm not sure which is better.

kate-goldenring: Really cool investigation -- this could be a blog post if not reason to put up a bug against the `allow needless collect` flag on clippy. Like you said, using the flag is neater more efficient code but what you have now is fine.

kate-goldenring (approved):
Once all checks pass, we can merge!

Ragnyll:
Huzzah! It's merged. Thanks @kate-goldenring . I'm taking a look around clippy to see if it's worth logging an issue for needless collect thing. There's a ton for that lint already that may fit, but I'll see. 

Thanks again!

Issue #453, opened 2022-02-23
Update to latest Kubernetes client (`kube-rs`) and types (`k8s-openapi`) versions


kate-goldenring:
Akri should be updated to use the latest versions of the Kubernetes client ([`kube-rs`](https://github.com/kube-rs/kube-rs)) and types (`k8s-openapi`) crates. The `k8s-openapi` version is dependent on the `kube-rs` version as seen in the [`kube-rs` README](https://github.com/kube-rs/kube-rs).

Ragnyll:
Ay, mind if I take this issue? Id like to take a crack at this as I'm familiarizing myself with the project more

kate-goldenring:
> Ay, mind if I take this issue? Id like to take a crack at this as I'm familiarizing myself with the project more

Hey @Ragnyll that would be great! I just assigned you to it. Let me know if you run into any issues.

Ragnyll:
Opened https://github.com/project-akri/akri/pull/458#issue-1160890021

PR #458, opened 2022-03-07
#453 update all kube versions

Ragnyll:
#453 updates kube-rs

I updated the feature on k8s_openapi from v1_16 to v1_22 . This is what is noted in the [installation guide for kube-rs](https://github.com/kube-rs/kube-rs#installation). I looked a bit at the [k8s_openapi docs](https://docs.rs/k8s-openapi/latest/k8s_openapi/index.html#) and this seems to indicate the version of kubernetes it supports. I'm not sure if we should keep this v1_16 or what the policy on this version update is or should be. I suppose that may be worth talking about since I'm still pretty early in my education on Kubernetes.

As far as this update goes, I think the simple version bumps are fine. I dont see any deprecation notices in the builds.

**If applicable**:


agent/Cargo.toml
```diff
@@ -61,10 +61,10 @@ tempfile = "3.1.0"
...
-# Or set the features when building the agent with  
+# Or set the features when building the agent with
 # `cargo build --manifest-path agent/Cargo.toml --features "agent-full onvif-feat opcua-feat udev-feat"`
 default = []
 onvif-feat = [ "akri-onvif"]
 opcua-feat = ["akri-opcua"]
 udev-feat = ["akri-udev"]
-agent-full = []
\ No newline at end of file
+agent-full = []
```
romoh, line 70: keep agent-full out by default?

kate-goldenring: i believe that this is just a spacing fix. it is still as it was set

agent/Cargo.toml
```diff
@@ -25,9 +25,9 @@ futures-core = "0.3"
 futures-util = "0.3"
 hyper = "0.14.2"
 h2 = "=0.3.3"
-kube = { version = "0.59.0",  features = ["derive"] }
-kube-runtime = "0.59.0"
-k8s-openapi = { version = "0.13.0", default-features = false, features = ["schemars", "v1_16"] }
+kube = { version = "0.69.1",  features = ["derive"] }
+kube-runtime = "0.69.1"
+k8s-openapi = { version = "0.14.0", default-features = false, features = ["schemars", "v1_22"] }
```
kate-goldenring, line 30: Looks like each [feature corresponds to a supported K8s version](https://github.com/Arnavion/k8s-openapi/blob/master/Cargo.toml#L54). I wonder if we want to list all from 1_16 to 1_22? curious how that changes the size of the crate. The e2e tests did pass on all the versions curiously.

kate-goldenring (commented):
@Ragnyll, thanks for this! This looks pretty good to go -- all the tests passed. It might be nice to get an understanding of how the `k8s_openapi` k8s features work before merging. This might result in just a comment saying why we only have `1_22` there, adding all features there, etc

agent/Cargo.toml
kate-goldenring: Should we do `1_23` instead of `1_22`?

kate-goldenring: It looks like based on how they are pulling K8s [swaggers here](https://github.com/Arnavion/k8s-openapi/blob/563883804ca06448f24890ebf47b4f7efa491caf/k8s-openapi-codegen/src/supported_version.rs#L60) that whatever version feature is chosen is the K8s client specification that is used (kubernetes source APIs can be found [here](https://github.com/kubernetes/kubernetes/tree/v1.23.4/api/openapi-spec)). I think it is safe to go with `1_22` or better yet `1_23` but i would be curious how to determine when the API we are using is not backwards compatible.

Ragnyll: If I follow that kubernetes api doc you linked through a few more hops I eventually end up at the Open API Spec. The [versioning clause states](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#versions):

> The OpenAPI Specification is versioned using a major.minor.patch versioning scheme.

which would make me think that so long as it doesnt jump a major version we would be fine. However the spec goes on to say:

> Occasionally, non-backwards compatible changes may be made in minor versions of the OAS where impact is believed to be low relative to the benefit provided.

which is a little spooky. I'm asking the maintainer of the crate for their thoughts [here](https://github.com/Arnavion/k8s-openapi/issues/113) . However I'd imagine the responsibility eventually falls to the end client code (so Akri) to handle the regression tests from changes in the OAPI. In that regard, if Akri's regression tests succeed then I imagine we would be fine.

Ragnyll: in any case though i've bumped the version to to v1_23. The local build stuff passed locally, so we'll see what the CI checks say.

agent/Cargo.toml
romoh: ahh got it

agent/Cargo.toml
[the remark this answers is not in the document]
kate-goldenring: breaking change for Akri? IMO no. As long as users are able to use akri with the same installations and custom resources, then i'd say its not breaking.

kate-goldenring: Thanks for inquiring about this! I think we are all set to merge.

kate-goldenring (approved):

agent/Cargo.toml
Ragnyll: Cool, I agree. And to repost what is probably the relevant situation in that conversation here:

> "Situation 3: Kubernetes releases v1.61.0. There will be a new major release of k8s-openapi, ie v0.101.0, with support for this new v1.61 line. This is because adding support for new Kubernetes version lines is semver-incompatible"...

So unless we have a major version change for k8s-openapi its fine. 

Sorry for the delay on this response. It's been an interesting week for me.

PR #481, opened 2022-06-07
Fix Kubernetes setup in end-to-end tests

kate-goldenring:
Signed-off-by: Kate Goldenring <kate.goldenring@microsoft.com>

For some reason, in the past we have not needed to install a networking addon (flannel calico etc) during Kubernetes setup in our e2e tests. Now, the node is getting stuck in a `NotReady` state, making our e2e tests fail before they can even be run. Seems like adding flannel solves the issue [a netowrking addon is needed to start coredns](https://v1-20.docs.kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#pod-network).

closes #481

### DIFF
--- a/.github/workflows/run-test-cases.yml
+++ b/.github/workflows/run-test-cases.yml
@@ -233,6 +233,7 @@ jobs:
           sudo mkdir -p $HOME/.kube
           sudo cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
           sudo chown $(id -u):$(id -g) $HOME/.kube/config
+          kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml
           kubectl taint nodes --all node-role.kubernetes.io/master-
           echo '--set kubernetesDistro=k8s' > /tmp/k8s_distro_to_test.txt
           echo 'kubectl' > /tmp/runtime_cmd_to_test.txt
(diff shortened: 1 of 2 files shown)


.github/workflows/run-test-cases.yml
```diff
@@ -229,10 +229,11 @@ jobs:
           kubeadm version && echo "kubeadm return code: $?" || echo "kubeadm return code: $?"
           kubelet --version && echo "kubelet return code: $?" || echo "kubelet return code: $?"
           sudo swapoff -a              
-          sudo kubeadm init
+          sudo kubeadm init --pod-network-cidr=10.244.0.0/16
```
romoh, line 232: any reason why not rely on the default?

romoh (approved):

.github/workflows/run-test-cases.yml
kate-goldenring: K8s [docs](https://v1-20.docs.kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#pod-network) say "Take care that your Pod network must not overlap with any of the host networks: you are likely to see problems if there is any overlap. (If you find a collision between your network plugin's preferred Pod network and some of your host networks, you should think of a suitable CIDR block to use instead, then use that during kubeadm init with --pod-network-cidr and as a replacement in your network plugin's YAML)."

Just using the default without a `pod-network-cidr` caused a `error: /run/flannel/subnet.env: no such file or directory` error in kubelet during Pod bringup. This [issue suggests](https://github.com/kubernetes/kubernetes/issues/70202#issuecomment-485145431) adding the CIDR network resolves it. We shall see.

PR #482, opened 2022-06-15
rust toolchain version update

romoh:
Updates the toolchain version. This is required to ingest new updated dependencies that requires a more recent version of rustc.


bfjelds (approved):

discovery-handler-modules/debug-echo-discovery-handler/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "debug-echo-discovery-handler"
-version = "0.8.11"
+version = "0.8.12"
 authors = ["Kate Goldenring <kate.goldenring@microsoft.com>"]
 edition = "2018"
 rust-version = "1.58.1"
```
kate-goldenring, line 6: The rust versions listed in all the cargo.tomls should be increased, as well.

kate-goldenring (changes requested):
Thanks for pushing this up! Just a couple more spots we need the version increased: cargo.tomls and the badge in the base readme. After this merges, we need to also update the docs to point to htis latest version too: https://github.com/project-akri/akri-docs/pull/22

agent/src/util/device_plugin_service.rs
```diff
@@ -746,8 +746,8 @@ pub async fn terminate_device_plugin_service(
 /// This creates an Instance's unique name
 pub fn get_device_instance_name(id: &str, config_name: &str) -> String {
     format!("{}-{}", config_name, &id)
-        .replace(".", "-")
-        .replace("/", "-")
+        .replace('.', "-")
+        .replace('/', "-")
```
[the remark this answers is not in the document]
romoh: yep, I changed it based on the clippy errors.

discovery-handler-modules/debug-echo-discovery-handler/Cargo.toml
romoh: Good catch

kate-goldenring:
Looks like the tarpaulin version may need to be increased or parameterized to use the correct rust version: https://github.com/xd009642/tarpaulin

PR #520, opened 2022-09-23
Implement external device inventory integration

leolumicrosoft:
It is related to proposal in akri-doc repository PR#45: integrate external device inventory proposal draft. The intention is to be able to query additional device information from external device inventory system and send to broker pod together with current basic device discovery information. Ultimately in the future this feature can be used to enable Akri work with a full functional device management platform for devices provisioning. It targets using Akri in a hybrid IoT solution with K8s edge cluster.

**Demo external device inventory for test**
To assist test and development of this feature, I have prepared a very simple demo device inventory prototype. You can directly deploy its image to cluster. 
```
kubectl create deployment deviceinventory --image=ghcr.io/leolumicrosoft/testakriintegratedevicesinventory:1.2
kubectl expose deployment/deviceinventory --port 3000 --type="NodePort"
#Find out the nodePort
kubectl get service/deviceinventory --output=jsonpath='{.spec.ports[?(@.port==3000)].nodePort}' && echo
```
After the demo device inventory is up, it will enable a REST endpoint for the Akri agent to query device info. For the convenience, it also provides a web page. You can use the web page to easily add/delete/update device records, and also monitor any newly discovered device by Akri (if the Akri config asks for querying additional device information).
Assume you will access the frontend web page from your own computer, you will need to create a SSH reverse tunnel from your remote K8s cluster node.
```
ssh -p [SSH port] -i [private SSH key] [username]@[node ip] -L 3000:localhost:[exposed nodePort]
eg.
ssh -p 8094 -i C:\Users\ll\.ssh\mykey.pem azureuser@20.192.200.46 -L 3000:localhost:31049
```
After that, you can open the web page in your browser with url: http://localhost:3000/, below is a sample screenshot

image

You can use the demo device inventory to add/update device record with any metadata in "query device result". Later after you prepare akri config with this demo devices inventory endpoint, you will see this information finally going into created Akri instance and further injecting into broker pods as an environment variable "EXTINFO".

>PS: The source code of this demo device inventory is in https://github.com/leolumicrosoft/testAkriIntegrateDevicesInventory 

**Test discovery handler config**:
After the demo external device inventory is setup, you are ready to add device query setting into Akri config file. You will add a new field "queryDeviceHttp" to the discoveryDetails part. Its value is http://localhost:[nodePort_number]/queryDevice if you are running akri in local environment. 
>PS: If you will build akri component as container and run within k8s cluster, you may directly use the service name "http://deviceinventory:3000/queryDevice" instead of "http://localhost:[nodePort_number]/queryDevice" for Akri agent accessing the external device inventory demo.

Below are two examples for DebugEcho config and udev config.
image

image

**If applicable**:
- [x] this PR has an associated PR with documentation in [akri-docs](https://github.com/project-akri/akri-docs)


kate-goldenring:
@leolumicrosoft thanks for this! I havent got the chance to look at this, but would you be interested in demoing your work at our community meeting October 4th at 9AM PT: https://hackmd.io/@akri/S1GKJidJd

discovery-handlers/debug-echo/src/discovery_handler.rs
```diff
@@ -76,53 +90,56 @@ impl DiscoveryHandler for DiscoveryHandlerImpl {
...
+                    let device_query_requests = descriptions.iter()
+                    .map(|description| {
+                        let mut properties = HashMap::new();
+                        properties.insert(
+                            super::DEBUG_ECHO_DESCRIPTION_LABEL.to_string(),
+                            String::from(description),
+                        );
+                        let query_body = QueryDevicePostBody{
+                            id:String::from(description),
+                            protocol: "DebugEcho".to_string()
```
bfjelds, line 117: maybe use DISCOVERY_HANDLER_NAME here?

discovery-handlers/onvif/src/discovery_handler.rs
```diff
@@ -520,6 +558,7 @@ mod tests {
             }),
             scopes: None,
             discovery_timeout_seconds: 1,
+            query_device_http: None,
```
bfjelds, line 561: we probably need a test somewhere where query_device_http==!None

discovery-utils/src/registration_client.rs
```diff
@@ -48,3 +56,70 @@ pub async fn register_discovery_handler_again(
...
+        for device_query in devices_query.clone() {
+            fetch_devices_tasks.push(tokio::spawn(query_device_info(device_query,query_http.clone())));    
+        }
+        let result= futures::future::try_join_all(fetch_devices_tasks).await;
+        if let Ok(query_devices_result) = result{
+            return query_devices_result;
+        }
+    }
+    
+    //if there is no need to query additional device information or the previous device query is not completely successful, simply assemeble Devices and return
```
bfjelds, line 113: `is not complteted successful` ... why no error proprogation here?  should we at least log the failure?

discovery-handlers/debug-echo/src/discovery_handler.rs
leolumicrosoft: you are right that I shall use the constants. I will change.

discovery-handlers/onvif/src/discovery_handler.rs
leolumicrosoft: Will that be to create a mockup REST device inventory and check the query result of a test device?

discovery-utils/src/registration_client.rs
leolumicrosoft: Http query should be logged. I will add it into query_device_info(registration_client.rs) .
All discovered device will be used to query info. There is chance some queries may fail. In current implementation, query_device_info will be invoked by each device. It will simply give "Device" result if a device query fails and wont' propagate to fail the future try_join_all call. Let's say 10 devices are queried and one fail, we will still get 9 correct assembled "Device" and the 10th "Device" with empty external information.

agent/src/util/discovery_operator.rs
```diff
@@ -208,6 +209,9 @@ impl DiscoveryOperator {
         let discovery_operator = Arc::new(self.clone());
         let stop_discovery_receiver: &mut tokio::sync::broadcast::Receiver<()> =
             &mut dh_details.close_discovery_handler_connection.subscribe();
+
+        let mut saved_instance_name_extinfo_map: HashMap<String, String> = HashMap::new();
```
kate-goldenring, line 213: nit: can this (`saved_instance_name_extinfo_map`) have a more concise name?

agent/src/util/discovery_operator.rs
```diff
@@ -218,6 +222,25 @@ impl DiscoveryOperator {
                 result = stream.get_message() => {
                     let response = result?.ok_or_else(|| anyhow::anyhow!("Received response type None. Should not happen."))?;
                     trace!("internal_do_discover - got discovery results {:?}", response.devices);
+
+                    //if device external information update is detected, simply remove the old akri instance
```
kate-goldenring, line 226: Is the commented code needed? Also, should have space between forward slash and comment text and start with capitalized letter. `cargo fmt` may help with some of this but i am not sure.

agent/src/util/discovery_operator.rs
```diff
@@ -478,11 +528,11 @@ pub mod start_discovery {
         kube_interface: Arc<dyn k8s::KubeInterface>,
     ) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
         let config = discovery_operator.get_config();
+        let config_name = config.metadata.name.clone().unwrap();
         info!(
-            "start_discovery - entered for {} discovery handler",
-            config.spec.discovery_handler.name
+            "start_discovery - entered for {} discovery handler with akric {}",
```
kate-goldenring, line 533: nit: can directly add value: `... discovery handler with akric {config_name}`

discovery-handlers/debug-echo/src/discovery_handler.rs
```diff
@@ -23,12 +27,23 @@ pub const DEBUG_ECHO_AVAILABILITY_CHECK_PATH: &str = "/tmp/debug-echo-availabili
...
+
 /// DebugEchoDiscoveryDetails describes the necessary information needed to discover and filter debug echo devices.
 /// Specifically, it contains a list (`descriptions`) of fake devices to be discovered.
 /// This information is expected to be serialized in the discovery details map sent during Discover requests.
 #[derive(Serialize, Deserialize, Clone, Debug)]
 #[serde(rename_all = "camelCase")]
 pub struct DebugEchoDiscoveryDetails {
+    //if there is query_device_http in the discovery detail, discovery handler will call agent QueryDeviceInfo rpc function for each discovered new device
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub query_device_http: Option<String>,
```
kate-goldenring, line 45: if every discovery handle is going to contain `query_device_http`, can this be moved under `DiscoveryHandler` rather than as part of the `discovery_details` which are unique to each discovery handler?

discovery-handlers/debug-echo/src/discovery_handler.rs
```diff
@@ -23,12 +27,23 @@ pub const DEBUG_ECHO_AVAILABILITY_CHECK_PATH: &str = "/tmp/debug-echo-availabili
...
 /// DebugEchoDiscoveryDetails describes the necessary information needed to discover and filter debug echo devices.
 /// Specifically, it contains a list (`descriptions`) of fake devices to be discovered.
 /// This information is expected to be serialized in the discovery details map sent during Discover requests.
 #[derive(Serialize, Deserialize, Clone, Debug)]
 #[serde(rename_all = "camelCase")]
 pub struct DebugEchoDiscoveryDetails {
+    //if there is query_device_http in the discovery detail, discovery handler will call agent QueryDeviceInfo rpc function for each discovered new device
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub query_device_http: Option<String>,    
+    //for debug echo discovery handler, each string in description will be treated as a discovered device ID
```
kate-goldenring, line 46: nit: `// For`

discovery-handlers/debug-echo/src/discovery_handler.rs
```diff
@@ -76,53 +90,56 @@ impl DiscoveryHandler for DiscoveryHandlerImpl {
                     break;
                 }
 
+                //Standardize the procedures as same as the method used in OPC protocol discovery handler
```
kate-goldenring, line 93: should this comment and standardization move to the `akri-shared` crate?

discovery-utils/proto/discovery.proto
```diff
@@ -29,6 +32,18 @@ message RegisterDiscoveryHandlerRequest {
 message Empty {
 }
 
+message QueryDeviceInfoRequest{
+    //the basic device information that discovery handler can get is used in this request, such as device ipaddress, mac address etc. It will be a JSON string and it will be directly send to external device inventory system REST API. It is external device inventory system's resonsibility to understand this query_device_payload JSON string and respond.
+    string query_device_payload = 1;
```
kate-goldenring, line 37: can these names be simplified to remove redundant `query_device`?

discovery-handlers/opcua/src/discovery_handler.rs
```diff
@@ -121,14 +138,19 @@ impl DiscoveryHandler for DiscoveryHandlerImpl {
...
                             properties,
-                            mounts: Vec::default(),
-                            device_specs: Vec::default(),
+                            query_device_payload:Some(serde_json::to_string(&query_body).unwrap_or(String::from("{}"))),
+                            mounts:Vec::default(),
                         }
                     })
-                    .collect::<Vec<Device>>();
+                    .collect::<Vec<DeviceQueryInput>>();
+                let discovered_devices= query_devices(device_query_requests,query_device_http.clone()).await;
```
kate-goldenring, line 153: I may be missing something, but I dont see where we are checking if a `query_device_http` was provided before calling `query_devices`

agent/src/util/discovery_operator.rs
kate-goldenring, line 226: Why not update the instance instead? If we remove it, then it will also bring down associated broekrs

discovery-utils/src/discovery/v0.rs
```diff
@@ -30,6 +30,19 @@ pub mod register_discovery_handler_request {
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Empty {}
 #[derive(Clone, PartialEq, ::prost::Message)]
+pub struct QueryDeviceInfoRequest {
+    ///the basic device information that discovery handler can get is used in this request, such as device ipaddress, mac address etc. It will be a JSON string and it will be directly send to external device inventory system REST API. It is external device inventory system's resonsibility to understand this query_device_payload JSON string and respond.
```
kate-goldenring, line 34: Where are we defining the expected JSON string per protocol? Do we expect this to change?

kate-goldenring (commented):
Just some preliminary questions, comments, and nits.

kate-goldenring:
Do we anticipate this being the default way that discovery handlers get credentials for a device? I noticed user/pass being a part of one of the payloads in the UI but don't see where discovery handlers are using the user/pass. Also, would this be how X.509 certs are passed? would it be better to have a local cert store that the agent updates and Discovery Handlers reference rather then sending them in an HTTP response body?

agent/src/util/discovery_operator.rs
leolumicrosoft: The comment is added for the newly added code. When external device information is updated in device inventory, the Akrii should be recreated to trigger new broker pod with latest external information. I will change the format.

discovery-handlers/debug-echo/src/discovery_handler.rs
leolumicrosoft: I definitely agree. As this is the first PR I work on Akri, I tried to not change too much, for example, not changing the yaml structure and HELM. I plan this as a feature implementation with minimum change to other part.

discovery-handlers/debug-echo/src/discovery_handler.rs
leolumicrosoft: Noted, thanks.

discovery-utils/proto/discovery.proto
leolumicrosoft: Ok.

discovery-handlers/opcua/src/discovery_handler.rs
leolumicrosoft: The http endpoint check is actually inside query_devices. You can call query_devices from discovery handler without http query. In that case, query_devices simply assemble Devices without insert external properties, the block from L102 will be skipped.
https://github.com/project-akri/akri/blob/8dc09fdc891659cc1d0ff60da9630ebb5aa1d92b/discovery-utils/src/registration_client.rs#L100-L102

agent/src/util/discovery_operator.rs
leolumicrosoft: I did not know we can update instance. Let me know more in the community meeting.

discovery-utils/src/discovery/v0.rs
leolumicrosoft: query_device_payload JSON strcture is defined in each discovery handler and it will be different for different protocol. for the current supported protocols, it is defined in each discovery_handler.rs. for example in udev protocol,
https://github.com/project-akri/akri/blob/8dc09fdc891659cc1d0ff60da9630ebb5aa1d92b/discovery-handlers/udev/src/discovery_handler.rs#L26-L31

leolumicrosoft:
> Do we anticipate this being the default way that discovery handlers get credentials for a device? I noticed user/pass being a part of one of the payloads in the UI but don't see where discovery handlers are using the user/pass. Also, would this be how X.509 certs are passed? would it be better to have a local cert store that the agent updates and Discovery Handlers reference rather then sending them in an HTTP response body?

This is a first step for now as http response is simple and generic, both for username/password, token or X509 cert/key pair case. I view it as the first step to support protected edge devices. 
There is definitely further safer implementation to expand. For example, there could be private certificate authority inside the cluster network (typically it is customer's IT/OT environment), Akri agent may dynamically generate public/private key pair and raise certificate signing request to the certificate authority to get a signed cert, and then being authorized to access edge devices, such as PLC machines' OPC interface. 
So, I do not view this HTTP REST feature implementation as the only way but it can be a first step to support accessing protected devices. Beyond that, it is also a good implementation to provide other device meta data, such as the hierarchy information of PLC machine, which factory/level/workflow stage it belongs to, for example. Or other important information , such as OPC query interval

leolumicrosoft:
close by mistake

leolumicrosoft:
I will love to. Is it possible to reschedule it to Oct 4th 5pm so I wont join in the late night?

PR #540, opened 2022-11-17
ListAndWatch only sends device if the list has changed

### LINKED ISSUE #379: `ListAndWatch` should only send `Devices` when the list has changed
Akri makes a device plugin for each device it discovers, so kubelet can advertise it as a node-level resource. Currently, the Agent in `device_plugin_service.rs` in its implementation of `list_and_watch` has each device plugin send the list of devices on a loop with interval of `LIST_AND_WATCH_SLEEP_SECS`. An additional step should be added before sending the devices to the kubelet [here](https://github.com/deislabs/akri/blob/main/agent/src/util/device_plugin_service.rs#L187) that checks if the [devices](https://github.com/deislabs/akri/blob/main/agent/src/util/device_plugin_service.rs#L183) have changed from the previous list of devices. The device plugin manager (implemented by kubelet) assumes -- see [code commented description of ListAndWatch](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#device-plugin-implementation) -- that it will only receive updates when state has changed. We are currently not following that paradigm. 

**A potential solution**
Keep a `Vector` of the previously returned devices and compare it to the ones to be sent. `Device` implements `PartialEq` so a `==` comparison can be made. If there are no changes, then do not send the devices that loop iteration.

harrison-tin:
Signed-off-by: Harrison Tin <harrisontin@microsoft.com>

As suggested [here](https://github.com/project-akri/akri/issues/379#:~:text=A%20potential%20solution), the change keeps a vector of previously returned virtual devices and only send updates to kubelet if the list has changed.
It closes #379
Test: passed all unit tests, and manually tested it with some debug messages.

**If applicable**:


agent/src/util/device_plugin_service.rs
```diff
@@ -179,23 +180,26 @@ impl DevicePlugin for DevicePluginService {
...
-                        "list_and_watch - for Instance {} kubelet no longer receiving with error {}",
-                        dps.instance_name,
-                        e
-                    );
-                    // This means kubelet is down/has been restarted. Remove instance from instance map so
-                    // do_periodic_discovery will create a new device plugin service for this instance.
-                    dps.instance_map.lock().await.remove(&dps.instance_name);
-                    dps.server_ender_sender.clone().send(()).await.unwrap();
-                    keep_looping = false;
+                if prev_virtual_devices != virtual_devices {
```
kate-goldenring, line 183: This assumes that the discovery handler always returns the devices in the same order. For example, the following assertion fails:
```sh
    let vec1 = vec![1,2,3];
    let vec2 = vec![2,1,3];
    assert!(vec1 == vec2);
```
You may consider using HashSets.

kate-goldenring (changes requested):
Thanks for looking into this!

kate-goldenring (approved):
This is great! Thank you for this change that gets us more aligned with the device plugin interface specification of only sending updated device lists

adithyaj:
@harrison-tin I think with the latest merge there's a clippy error [that popped up](https://github.com/project-akri/akri/actions/runs/3568943314/jobs/5998368891)

kate-goldenring:
@harrison-tin @adithyaj are we good to merge this?

harrison-tin:
> @harrison-tin @adithyaj are we good to merge this?

yep LGTM. I can go ahead and merge it.

PR #541, opened 2022-11-21
change tokio::sync::Mutex to tokio::sync::RwLock and use read()/write…

### LINKED ISSUE #376: Use `tokio::sync::RwLock` instead of `tokio::sync::Mutex`
Throughout Akri's code, we are often using a `Mutex` where a [`RwLock`](https://docs.rs/tokio/1.11.0/tokio/sync/struct.RwLock.html) would be preferred, as it provides multiple read handles while only one write handle. For example, our maps of data, such as the Agent's `InstanceMap` which is `Arc<Mutex<HashMap<String, InstanceInfo>>>` would be better designed as `Arc<RwLock<HashMap<String, InstanceInfo>>>` as many times only reads are needed.

Change usage of `Mutex` to `RwLock` where preferable.

harrison-tin:
…() instead of lock()

Signed-off-by: Harrison Tin <harrisontin@microsoft.com>

This PR changes the use of tokio::sync::Mutex to tokio::sync::RwLock (used by configMap and instanceMap). RwLock uses read()/write() depending on the action of the thread instead of lock() from Mutex.
It closes #376 

**If applicable**:


kate-goldenring (approved):
LGTM! Thanks for this!

PR #545, opened 2022-12-21
Add new K8S versions to tests

jbpaux:
Kubernetes released new upstream version (1.26) and all distros were updated to take it into account.
This PR also updates the previous minor versions where applicable.

I've updated also a few things:
- In 1.26 containerd must be updated as [CRI v1alpha2 was removed](https://kubernetes.io/blog/2022/12/09/kubernetes-v1-26-release/#cri-v1alpha2-removed).
- Logic for node taints was changed to take into account all next versions of Kubernetes (was quite static in current implementation)

**If applicable**:

### DIFF
--- a/.github/workflows/run-test-cases.yml
+++ b/.github/workflows/run-test-cases.yml
@@ -111,26 +111,34 @@ jobs:
             version: 1.23/stable
           - runtime: MicroK8s-1.24
             version: 1.24/stable
+          - runtime: MicroK8s-1.25
+            version: 1.25/stable
+          - runtime: MicroK8s-1.26
+            version: 1.26/stable
           - runtime: K3s-1.21
-            version: v1.21.5+k3s1
+            version: v1.21.14+k3s1
           - runtime: K3s-1.22
-            version: v1.22.6+k3s1
+            version: v1.22.16+k3s1
           - runtime: K3s-1.23
-            version: v1.23.13+k3s1
+            version: v1.23.14+k3s1
           - runtime: K3s-1.24
-            version: v1.24.6+k3s1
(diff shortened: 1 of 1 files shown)


.github/workflows/run-test-cases.yml
```diff
@@ -266,10 +276,10 @@ jobs:
       - if: (startsWith(github.event_name, 'pull_request')) && (startsWith(matrix.kube.runtime, 'Kubernetes'))
         name: Import local agent and controller to Kubernetes
         run: |
-          sudo docker load --input agent.tar
-          sudo docker load --input controller.tar
-          sudo docker load --input webhook-configuration.tar
-          sudo docker image ls
+          #sudo docker load --input agent.tar
```
kate-goldenring, line 279: Instead of commenting out, i think these commands are fine to remove now that we are importing containers with `ctr` for containerd instead

kate-goldenring:
@jbpaux can we remove Kubernetes version `1.21` and `1.22` from the test matrix as they have now reached end of life: https://endoflife.date/kubernetes. We are aiming to remove EOL versions each time we add the latest. @adithyaj, would you agree with this?

.github/workflows/run-test-cases.yml
jbpaux: Yes I hesitated but you're right it's useless now, fixed.

jbpaux:
@kate-goldenring I've removed old Kubernetes versions in latest commit. Should be good.

kate-goldenring (approved):

kate-goldenring:
Thanks @jbpaux for keeping this up to date!

PR #547, opened 2022-12-28
Ensure Akri's Helm templates use the most restrictive settings suggested by Snyk report

harrison-tin:
This PR explicitly sets more SecurityContext rather than using default in the helm templates, using Akri's Snyk report for guidance. The templates changed are:
1. agent: added more securityContext
2. controller: added more securityContext
3. anomaly-detection-app: added more securityContext and resource limits
4. video-streaming-app: added more securityContext and resource limits

**If applicable**:


deployment/samples/akri-anomaly-detection-app.yaml
```diff
@@ -16,6 +16,19 @@ spec:
       - name: akri-anomaly-detection-app
         image: ghcr.io/project-akri/akri/anomaly-detection-app:latest-dev
         imagePullPolicy: Always
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsNonRoot: true
+          readOnlyRootFilesystem: true
+          capabilities:
+            drop: ["ALL"]
+        resources:
```
kate-goldenring, line 25: Have we tested that these levels of resources are sufficient for the anomaly app? I think it would be better to leave resource limits out for our samples as they are not meant to be used in production.

deployment/helm/values.yaml
```diff
@@ -91,7 +98,12 @@ agent:
     tag:
     # pullPolicy is the Akri Agent pull policy
     pullPolicy: ""
-  securityContext: {}
+  # ensures container doesn't run with unnecessary priviledges
+  securityContext:
+    allowPrivilegeEscalation: false
```
kate-goldenring, line 103: This likely will cause the agent to fail as it needs [to run as privileged](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#device-plugin-deployment) to create device plugins with the kubelet

kate-goldenring (changes requested):
I think we need to open up the Agent as it needs privileged access. This requires running this workflow locally and inspecting kubelet logs for errors `journalctl -fu kubelet`

deployment/helm/values.yaml
```diff
@@ -91,7 +98,12 @@ agent:
...
+    readOnlyRootFilesystem: true
```
jbpaux, line 104: Tests are failing because tests are writing a file in the agent pod:
https://github.com/project-akri/akri/blob/47b42587620346e23258e19ed9c5c61ed6904487/test/run-end-to-end.py#L106

kate-goldenring: Good point! Maybe we add that file as a volume mount for the Pod during the test

deployment/samples/akri-anomaly-detection-app.yaml
harrison-tin: thanks, removed the resource limits for anomaly detection app

deployment/helm/values.yaml
harrison-tin: removed the limitation for agent

kate-goldenring: Can we explicitly state privileged access rather than removing and it defaulting to that. This security context from the [nvidia GPU device plugin should work](https://github.com/NVIDIA/k8s-device-plugin/blob/main/nvidia-device-plugin.yml#L46)

harrison-tin: thanks for the suggestion and sorry for the late reply. Updated the security context of agent to follow the nvidia gpu device plugin

deployment/helm/values.yaml
kate-goldenring: I am not sure i understand how this was resolved. Looks like `readOnlyRootFilesystem: true` still exists which is ideal.

harrison-tin: I think the above highlight is outdated? [Here](https://github.com/project-akri/akri/blob/b4ffc1c39c52b84cadb3dcce0ed3ce2f9296bdb8/deployment/helm/values.yaml#:~:text=%3A%20%22%22-,securityContext%3A,drop%3A%20%5B%22ALL%22%5D,-host%3A) shows the updated security context for akri agent. I removed setting readOnlyRootFilesystem to true (followed what nvidia gpu device plugin did)

kate-goldenring (approved):
Thanks for putting this together! We may want to do a round two at some point with the discovery handlers. If you pull in the latest from main, tests should pass. We may need to bump version depending on whether this goes in before https://github.com/project-akri/akri/pull/556

PR #554, opened 2023-02-09
Set pre_start_required to false in get_device_plugin_options

johnsonshih:
Signed-off-by: johnsonshih <jshih@microsoft.com>

Set pre_start_required to false in get_device_plugin_options to match the setting in register().

We set pre_start_required to false when akri instance device plugin registers itself to kubelet but kubelet still calls get_device_plugin_options to decide if to call device plugin's pre_start_container()  In get_device_plugin_options(), we return true for pre_start_required so kubelet calls pre_start_container().  We don't expect the pre_start_container() to be called so we dump an error message and return empty option.  Since we don't want the pre_start_container() to be called, we should set pre_start_required to false in get_device_plugin_options().

**If applicable**:


kate-goldenring:
@johnsonshih can you provide some context on why this change is desired?

johnsonshih:
> @johnsonshih can you provide some context on why this change is desired and the anticipated impact?

We set pre_start_required to false when akri instance device plugin registers itself to kubelet but kubelet still calls get_device_plugin_options to decide if to call device plugin's pre_start_container()  In get_device_plugin_options(), we return true for pre_start_required so kubelet calls pre_start_container().  We don't expect the pre_start_container() to be called so we dump an error message and return empty option.  Since we don't want the pre_start_container() to be called, we should set pre_start_required to false in get_device_plugin_options().

adithyaj (approved):

PR #556, opened 2023-02-17
Modify Agent to reduce frequency of Pods getting `UnexpectedAdmissionError`

### LINKED ISSUE #450: Modify Agent to reduce frequency of Pods getting `UnexpectedAdmissionError`
The Akri Agent creates device plugins for each device it discovers. The device plugin framework allows for specifying how many containers can use a device plugin at a time. The kubelet will then ensure that no Pod is scheduled to use that resource if all usage has been taken up. 

If the kubelet believes that a resource/device plugin should be available but the device plugin/ Agent responds that it is not available, the Pod will show an `UnexpectedAdmissionError`. 

Currently, the Agent errors when kubelet requests to use a device usage slot it is already using. Even though the kubelet is the source of truth, the Agent was erroring here in order to free up the usage slot so that any node could take it. In short, the Agent is actively creating `UnexpectedAdmissionErrors` on behalf of Node fairness for shared devices. 

This behavior should change to reduce `UnexpectedAdmissionErrors`. The Agent should trust krustlet as the source of truth. This will lead to less admission errors, especially in Jobs scenarios like the following:

```sh
helm upgrade akri akri-helm-charts/akri-dev  \
    $AKRI_HELM_CRICTL_CONFIGURATION  \
    --set agent.allowDebugEcho=true  \
    --set debugEcho.discovery.enabled=true  \
    --set debugEcho.configuration.brokerJob.image.repository=busybox  \
    --set debugEcho.configuration.brokerJob.command[0]="sh"  \
    --set debugEcho.configuration.brokerJob.command[1]="-c"  \
    --set debugEcho.configuration.brokerJob.command[2]="echo 'Hello Amazing World'"  \
    --set debugEcho.configuration.brokerJob.command[3]="sleep 5"  \
    --set debugEcho.configuration.brokerJob.parallelism=2  \
    --set debugEcho.configuration.brokerJob.completions=2  \
    --set debugEcho.configuration.enabled=true 
```

kate-goldenring:

Previously we were clearing device usage slots when a node tries to reallocate it. We did this for fairness reasons, but it meant denying kubelet as the source of truth and irritating `UnexpectedAdmissionError`s. Furthermore, fairness of reallocating slots should already be ensured by `SlotReconcilliation`, which be clearing unused slots more immediately, reducing the frequency of this scenario. Also, for local unshared devices, there is no reason to free the device up for other nodes.

fixes #450 

@bfjelds  and @romoh would be great to get quick eyes on this small but significant change.

**If applicable**:


kate-goldenring:
working on failed tests in https://github.com/project-akri/akri/pull/559

adithyaj:
> working on failed tests in #559

merged #559, and now I'm rerunning the failed tests

adithyaj (approved):

Issue #492, opened 2022-07-07
Enable mounting connectivity information for multiple devices/instances in a Pod


kate-goldenring:
## Background on how Akri adds connectivity information to brokers 
The Akri Agent creates a [device plugin](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) for each device it discovers, creating a node level resource for that device. Then, a Pod can request that IoT device just as it would a compute resource (CPU and memory) in the `resourceRequests` section of the PodSpec. This is explained further in Akri's [requesting Akri resources documentation](https://docs.akri.sh/user-guide/requesting-akri-resources). Akri's Controller does this on your behalf when a `brokerSpec` is specified in a `Configuration`. It will deploy that `Pod` or `Job`, adding the discovered device as a `resourceRequest`. Then, before the Pod is run on a node, the kubelet calls `Allocate` on the device plugin in the Akri Agent to confirm that it can use that resource (device). In it's response, the device plugin can specify any environment variables or mounts that should be added to the container using that device. This is where Akri sets connectivity information (and adds devnode mounts for `udev`). The environment variables are delineated in Akri's broker [documentation](https://docs.akri.sh/development/broker-development#discovery-handler-specified-environment-variables).

## Problem 
The issue is that there is one environment variable name for all of a protocol of devices. This has been fine in Akri's prevailing scenario of one device per broker; however, it causes missing connectivity information for the remaining N-1 devices requested.

For example, for ONVIF, the RTSP url for the camera is set under an environment variable named `ONVIF_DEVICE_SERVICE_URL`, so say a Pod has the following `resourceRequests`:
```yaml
  resources:
    limits:
      akri.sh/onvif-camera-id1: "1"
      akri.sh/onvif-camera-id2: "1"
    requests:
      akri.sh/onvif-camera-id1: "1"
      akri.sh/onvif-camera-id2: "1"
```
When the broker Pod books up, only one (of the two camera's) RTSP URL is set in the environment variable `ONVIF_DEVICE_SERVICE_URL`. Instead, a list of urls should be specified or there should be N environment variables for N devices, each with an expected prefix, so the broker could do the equivalent of `printenv | grep ONVIF_DEVICE_SERVICE_URL*` to know which devices it can connect to.

## Demo of the problem
[...]

PR #560, opened 2023-02-28
Suffix usage slot to annotation key name

johnsonshih:

When allocate a device plugin, Agent adds the usage slot id to annotations to keep track of slot usage.  Currently, when a container uses multiple device plugins, all device plugins will set their values to the same annotation key and only information of one slot is recorded.  This PR suffix slot id to the annotation key name to keep all slot information.

The new format of annotation is:
akri.agent.slot-"slot id": "slot id"

In Agent, we parse the annotation keys use the format to avoid accidentally pulling in annotations with similar key name but are not set by Agent.

This PR is part of the fix for issue https://github.com/project-akri/akri/issues/492

**If applicable**:


adithyaj:
Johnson, I am unable to see any of the tests running for this PR. Were you able to see the tests run when you committed the changed? Can you also mark that `cargo clippy`/`cargo fmt`/`cargo build` etc from up top?

adithyaj:
tests ran successfully 🎉 and this looks good to me; this should likely have a docs change as well when #560 & #561 resolves #492

adithyaj (approved):

johnsonshih:
> tests ran successfully 🎉 and this looks good to me; this should likely have a docs change as well when #560 & #561 resolves #492

PR#560 is transparent to user.
PR#561 requires document change.

PR #561, opened 2023-02-28
Append hash of device usage id to device property key name

johnsonshih:
When allocate a device plugin, Agent adds the device properties as the container's environment variables.  Currently, when a container uses multiple device plugins, all device plugins will set device properties to the same set of environment variables and only the last one write to the environment variable is recorded. This PR suffix device usage id to the device property key name so the environment variables won't be overwritten.

This PR is part of the fix for issue https://github.com/project-akri/akri/issues/492

**If applicable**:


kate-goldenring:
I believe this is a breaking change since it will break all existing brokers when they look up properties. We should make this a minor version change and we will need update docs.

agent/src/util/device_plugin_service.rs
```diff
@@ -300,6 +302,16 @@ impl DevicePluginService {
                     device_usage_id.clone(),
                 );
 
+                let hash_id_value = generate_digest(&device_usage_id, 3);
```
kate-goldenring, line 305: I wonder if we should `to_upper` this so the env vars are still all caps. This is purely aesthetic probably

johnsonshih: call to_uppercase() so environment variables remain all caps.

agent/src/util/device_plugin_service.rs
```diff
@@ -300,6 +302,16 @@ impl DevicePluginService {
                     device_usage_id.clone(),
                 );
 
+                let hash_id_value = generate_digest(&device_usage_id, 3).to_uppercase();
```
kate-goldenring, line 305: I have concerns about this approach of generating a random hash for a device property or environment variable and this being a different hash from the instance name hash.

For example, a broker that uses instances `akri-udev-camera-hagfg` and `akri-udev-camera-fgggg` will have two env vars injected called `UDEV_DEVNODE_GGGG` and `UDEV_DEVNODE_AAAAA` with no way to correlate which env var is associated with which instance. This may not matter but i wonder about a scenario in the future where it might and documenting this could be hard.

Why not use the hash from the instance name in `internal_allocate` rather than regenerating one

kate-goldenring (changes requested):
I am concerned about my comment around regenerating random hashes rather than associating properties via an instance's hash

agent/src/util/device_plugin_service.rs
johnsonshih: It's a matter of which one (instance or slot) we want to use for the environment variable.

if we use the hash from the instance name, then all slots from the same instance will use the same env. variable name.  Pros: same has suffix for both instance and env. var.  Cons: Pods cannot confirm how many virtual devices assigned to it.

if we use hash from slot, the each slot has it's own env. variable name (and all slots from the same instance will have the same value).  Pros: The number of env. var. matches the number of virtual devices allocated to the pod.  Cons: pod need to handle the duplication of env var value.

I decided to pick the slot hash, but is opened to any suggestion.

samples/brokers/onvif-video-broker/Akri.cs
```diff
@@ -38,6 +41,11 @@ private static string PostSoapRequest(String requestUri, String action, String s
 		private const String GET_PROFILES_SOAP = @"<soap:Envelope xmlns:soap=""http://www.w3.org/2003/05/soap-envelope"" xmlns:wsdl=""http://www.onvif.org/ver10/media/wsdl""><soap:Header/><soap:Body><wsdl:GetProfiles/></soap:Body></soap:Envelope>";
 		private const String GET_STREAMING_URI_SOAP_TEMPLATE = @"<soap:Envelope xmlns:soap=""http://www.w3.org/2003/05/soap-envelope"" xmlns:wsdl=""http://www.onvif.org/ver10/media/wsdl"" xmlns:sch=""http://www.onvif.org/ver10/schema""><soap:Header/><soap:Body><wsdl:GetStreamUri><wsdl:StreamSetup><sch:Stream>RTP-Unicast</sch:Stream><sch:Transport><sch:Protocol>RTSP</sch:Protocol></sch:Transport></wsdl:StreamSetup><wsdl:ProfileToken>{0}</wsdl:ProfileToken></wsdl:GetStreamUri></soap:Body></soap:Envelope>";
 
+		// Regular expression pattern of environment variables that hold OPC UA DiscoveryURL
+		// The pattern is ONVIF_DEVICE_SERVICE_URL_ followed by 6 digit digest.  e.g.
+		// ONVIF_DEVICE_SERVICE_URL_123456, ONVIF_DEVICE_SERVICE_URL_ABCDEF
+		private const string OnvifDeviceServiceUrlLabelPattern = "ONVIF_DEVICE_SERVICE_URL_[a-fA-F0-9]{6,6}$";
```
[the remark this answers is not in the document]
johnsonshih: update the pattern to only match uppercase chars.

agent/src/util/device_plugin_service.rs
kate-goldenring: > Cons: Pods cannot confirm how many virtual devices assigned to it.

I don't believe we would ever have a case where a pod has two virtual devices from the same instance allocated to it. Why would this be desirable? The concept of virtual devices is to enable device sharing across pods

johnsonshih: We don't prohibit a pod to use multiple virtual devices from the same instance, if user put the quantity to 2" in the Configuration.  It becomes possible.  {{PLACEHOLDER}}" : "2"
Understand this is not what users will usually do but since we don't prohibit it, I think use slot id hash is better than silently eliminate the duplication.

Another alternative to address this is to change the suffix schema from a device slot hash to instace hash_slotid.  ex.
instance name: akri-udev-camera-hagfg
env variable name reported by virtual devices from the same instance: akri-udev-camera-hagfg_1, akri-udev-camera-hagfg_2, ...
This will keep the correlation between var. name and instance and also preserve the mapping of env. var and virtual devices.  Thought?

kate-goldenring: even if a user used two virtual devices (`{{PLACEHOLDER}}" : "2"`), it would map to the same devices and therefore same properties/env var, so an overwrite of the environment variable would be find. the information about that singular real device would be accessible

johnsonshih: I had change the code to use instance hash id, so all virtual devices from the instance for debug echo device name: akri-debug-echo-foo-8120fe will export env var named DEBUG_ECHO_DESCRIPTION_8120FE.  Please check the change and let me know if any concern.

kate-goldenring (approved):
Thanks for this great addition @johnsonshih!

PR #564, opened 2023-03-02
udev: Implement a way to group devices of a same tree

### LINKED ISSUE #490: Enable udev discovery handler to discover multiple node devices
## Background
Akri's `udev` discovery handler assumes that each `udev` device consists of one devnode, so it reports each devnode captured by the provided udev rules as a separte device to Akri. 

However, many devices span multiple devnodes. For example, some USB cameras consist of several nodes, such as `/dev/video0` and `/dev/video1` where one contains metadata and the other does actual footage capturing. In this case, one camera should be reported rather than 2. The work around in this scenario, was to specificy only discovering cameras with `:capture:` capabilities ref https://github.com/project-akri/akri-docs/pull/36, but not all devices have a "dominant" node like this.

This issue was also inspired by the following question [from this slack thread](https://kubernetes.slack.com/archives/C01D9L7QE8Z/p1655309513284689): _I have a client who has devices they'd like to broker with Akri but their devices apparently function with multiple udev mounts. I have not dug too deeply into this use case yet, but they say this prevents the use of Akri. I guess my question is, can I use Akri with multiple udev mounts per broker?

## Potential solution
Additional discoveryDetails should be passed to the `udev` Discovery Handler that enable specifying a set or a common property shared among the devnodes associated with a device. 
Investigation needs to be done on whether common properties (such as vendor id) exist.

Pertinent code
- udev Discovery Handler [`discoveryDetails`](https://github.com/project-akri/akri/blob/2b42eb28a41aaab966eb2823a693464153a127ee/discovery-handlers/udev/src/discovery_handler.rs#L25) should be expanded to include ability to specify a "set" of nodes. This is then specified by the user in an Akri Configuration.
- udev Discovery Handler would set multiple devnode mounts [here](https://github.com/project-akri/akri/blob/2b42eb28a41aaab966eb2823a693464153a127ee/discovery-handlers/udev/src/discovery_handler.rs#L94)

diconico07:
This PR adds the `groupRecursive` option to the udev device discovery handler, this configuration switch allows to group devices that share a parent/child relation.

This can be used to group into one akri instance a complex device that is split into multiple nodes. It is up to the user to define the boundaries of the device as there are no standard way to do so.

It works by comparing sysfs path for a device prefixing another one (i.e `/devices/pci0` is a parent of `/devices/pci0/0.1/0.1.1/0.1.1:1`, this example is obviously fictive)

It adds all devnodes to the mounts and add a UDEV_DEVNODE_n property for each one of them.

The UDEV_DEVPATH just need the topmost one as all other are subdirectories of it.

This closes #490 

I will cross-link with a documentation PR as soon as I finish writing it.

**If applicable**:
- [x] this PR has an associated PR with documentation in [akri-docs](https://github.com/project-akri/akri-docs) project-akri/akri-docs#64


discovery-handlers/udev/src/discovery_impl.rs
```diff
@@ -543,6 +543,24 @@ fn device_or_parents_have_tag(device: &impl DeviceExt, value_regex: &Regex) -> b
     }
 }
 
+/// Retrieve Parent or Children of a device using their sysfs path.
+pub fn get_device_relatives<'a>(
```
kate-goldenring, line 547: @diconico07 can you write a few unit tests for this?

discovery-handlers/udev/src/discovery_handler.rs
```diff
@@ -68,37 +71,64 @@ impl DiscoveryHandler for DiscoveryHandlerImpl {
...
-                            mounts.push(Mount {
-                                container_path: devnode.clone(),
-                                host_path: devnode,
-                                read_only: true,
-                            });
+                        for (i, (_, node)) in paths.into_iter().enumerate() {
+                            let property_suffix = match discovery_handler_config.group_recursive {
+                                true => format!("_{}", i),
+                                false => "".to_string(),
+                            };
```
kate-goldenring, line 112: Believe this can be simplified like so
```suggestion
            let property_suffix = discovery_handler_config.group_recursive.then(|| format!("_{}", i)).unwrap_or_default();
```

discovery-handlers/udev/src/discovery_handler.rs
```diff
@@ -68,37 +71,64 @@ impl DiscoveryHandler for DiscoveryHandlerImpl {
...
                     let enumerator = udev_enumerator::create_enumerator();
                     let paths = do_parse_and_find(enumerator, rule).unwrap();
-                    paths.into_iter().for_each(|path| {
-                        devpaths.insert(path);
-                    });
+                    for path in paths.into_iter() {
+                        if !discovery_handler_config.group_recursive {
+                            devpaths.insert(path.0.clone(), vec![path]);
+                        } else {
+                            match get_device_relatives(&path.0, devpaths.keys()) {
```
kate-goldenring, line 82: nit: i would put this else block in a separate function so we can unit test it.

deployment/helm/values.yaml
```diff
@@ -706,6 +706,8 @@ udev:
     # created as a result of applying this udev configuration
     brokerProperties: {}
     discoveryDetails:
+      # groupRecursive defines whether to group discovered parent/children under the same instance
+      groupRecursive: false
```
kate-goldenring, line 710: I would set this to true in `helm_lint_values.yaml` to test the chart

kate-goldenring (commented):
Just a few nits but overall this looks great. Thanks! A couple unit tests here would be helpful. We also want to be mindful of version here. I believe we have another pr in queue that bumps to 0.10.0.

discovery-handlers/udev/src/discovery_impl.rs
diconico07: I wrote a test case, tell me if you feel I missed something :smile:

discovery-handlers/udev/src/discovery_handler.rs
diconico07: will do this next week, as I'll do the unit tests as well (won't have enough time before that), if you want to merge before that, I'll do it in a separate PR.

kate-goldenring (approved):

Issue #145, opened 2020-11-23
Udev devices should be mounted through DeviceSpec


jiria:
**Describe the bug**
Udev discovered devices are currently mounted through Mount, which does not allow for fine grained control of access permissions. We should switch to DeviceSpec and expose the desired permission control through Configuration CRD.

```

/// Mount specifies a host volume to mount into a container.
--
  | /// where device library or tools are installed on host and   container
  | #[derive(Clone,   PartialEq, ::prost::Message)]
  | pub struct Mount {
  | /// Path of the mount within the   container.
  | #[prost(string, tag = "1")]
  | pub container_path: std::string::String,
  | /// Path of the mount on the   host.
  | #[prost(string, tag = "2")]
  | pub host_path: std::string::String,
  | /// If set, the mount is   read-only.
  | #[prost(bool, tag = "3")]
  | pub read_only: bool,
  | }
  | /// DeviceSpec specifies a host device to mount into a   container.
  | #[derive(Clone,   PartialEq, ::prost::Message)]
  | pub struct DeviceSpec {
  | /// Path of the device within the   container.
  | #[prost(string, tag = "1")]
  | pub container_path: std::string::String,
  | /// Path of the device on the   host.
  | #[prost(string, tag = "2")]
  | pub host_path: std::string::String,
  | /// Cgroups permissions of the   device, candidates are one or more of
  | /// * r - allows container to   read from the specified device.
  | /// * w - allows container to   write to the specified device.
  | /// * m - allows container to   create device files that do not yet exist.
  | #[prost(string, tag = "3")]
  | pub permissions: std::string::String,
  | }


```

diconico07:
After some testing after a thread on the slack channel, using `Mounts` instead of `DeviceSpecs` also prevents the said devices to be added in the containers cgroups `devices.allow` list, thus denying access to the devices to non privileged containers.

PR #576, opened 2023-04-11
udev: Use device specs instead of mounts

diconico07:
Device plugin mechanism differentiate Mounts, and DeviceSpecs, Mounts should be used to make libraries or tools available to the pod, and DeviceSpecs to make device nodes available to the pod.

This commit switch from using Mounts to using DeviceSpec for device nodes. This currently makes Mounts unused.

This is related to #145 and fixes a bug related to non-privileged pods reported on Slack

**If applicable**:


kate-goldenring (approved):
Thanks for this change!

PR #548, opened 2023-01-04
Update nodejs actions to node16

jbpaux:
**What this PR does / why we need it**:
This PR updates nodejs actions used by the different workflows to node16 as node12 is deprecated.
It updates both embedded actions in the projects and dependent actions (except `actions-rs` actions, see below).
It closes #537 

**Special notes for your reviewer**:
`actions-rs` actions [(`audit-check`](https://github.com/actions-rs/audit-check) and [`toolchain`)](https://github.com/actions-rs/toolchain) don't have node16 compatible versions available and the maintainer is gone. I would advise to switch to other actions. I found one for the toolchain: the [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) one but none for the audit-check. We can maybe tackle that in another PR.

**If applicable**:
- [ ] this PR has an associated PR with documentation in [akri-docs](https://github.com/project-akri/akri-docs)
- [ ] this PR contains unit tests
- [ ] added code adheres to standard Rust formatting (`cargo fmt`)
- [ ] code builds properly (`cargo build`)
- [ ] code is free of common mistakes (`cargo clippy`)
- [ ] all Akri tests succeed (`cargo test`)
- [ ] inline documentation builds (`cargo doc`)
- [x] all commits pass the [DCO bot check](https://probot.github.io/apps/dco/) by being signed off -- see the failing DCO check for instructions on how to retroactively sign commits

PR #581, opened 2023-04-12
Upgrade Rust CI actions to maintained ones

### LINKED ISSUE #579: Migrate away from using `actions-rs`
We are updating to latest node versions in https://github.com/project-akri/akri/pull/548; however, `actions-rs` actions [(audit-check](https://github.com/actions-rs/audit-check) and [toolchain)](https://github.com/actions-rs/toolchain) don't have node16 compatible versions available and the maintainer is gone. @jbpaux advises to switch to other actions-- for example the toolchain: the [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) one but none for the audit-check.

jbpaux:
Upgrade the Rust actions not upgraded in PR #548. Switched to other actions as the current ones are not maintained anymore.
Fix #579

**If applicable**:

### DIFF
--- a/.github/workflows/build-rust-code.yml
+++ b/.github/workflows/build-rust-code.yml
@@ -72,10 +72,9 @@ jobs:
         persist-credentials: false
 
     - name: Rust install
-      uses: actions-rs/toolchain@v1
+      uses: dtolnay/rust-toolchain@master
       with:
         toolchain: 1.68.1
-        override: true
         components: clippy, rustfmt
 
     - name: Build production rust for ${{ matrix.arch.label }}
--- a/.github/workflows/check-rust.yml
+++ b/.github/workflows/check-rust.yml
@@ -31,7 +31,7 @@ jobs:
         persist-credentials: false
 
     - name: Rust install
-      uses: dtolnay/rust-toolchain@stable
+      uses: dtolnay/rust-toolchain@master
       with:
         toolchain: 1.68.1
         components: clippy, rustfmt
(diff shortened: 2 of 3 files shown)


yujinkim-msft:
@jbpaux thank you for this PR! could you please take a look at the failed check error (versioning/build)?

jbpaux:
It's just because there was no version specified in the PR. That's why I added the kennel but I don't have permission to relaunch the test

jbpaux:
If you want to have a look @kate-goldenring :)

.github/workflows/build-rust-code.yml
```diff
@@ -72,10 +72,9 @@ jobs:
         persist-credentials: false
 
     - name: Rust install
-      uses: actions-rs/toolchain@v1
+      uses: dtolnay/rust-toolchain@master
```
kate-goldenring, line 75: should we point to stable here too rather than master?

jbpaux: I guess it will depend if you want to use a specific version of the go tool chain or not.
According to the documentation about the tool chain parameter :
> Rustup toolchain specifier e.g. stable, nightly, 1.42.0, nightly-2022-01-01. Important: the default is to match the @rev as described above. When passing an explicit toolchain as an input instead of @rev, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action.

kate-goldenring: @jbpaux thanks. Sounds like we can have them both be master then and should switch `check-rust.yml` to `@master`.

jbpaux: exact. I don't know why I used `@stable` for `check-rust.yml`. I'll update.

kate-goldenring (approved):

PR #588, opened 2023-04-19
helm: Generate a self-signed cert for webhook if none given

### LINKED ISSUE #569: Auto-generate certificates required for webhook installation
Akri's validation webhook installation requires manually handling certificates and CA.
It would be a nice addition to the webhook install process to allow to automatically generate these if not provided.

This is already done by many CRD providers, like for example the gateway API that uses `Job` resources and the `registry.k8s.io/ingress-nginx/kube-webhook-certgen` image to handle this.
I guess we can do the same, we may want to make it opt-out so that a user wanting to provide their own certificate can still do it.

This would also allow to install the admission webhook by default as well.

diconico07:
Use the ingress-nginx/kube-webhook-certgen image to generate a self-signed certificate valid for 100 years for admission webhook when the caBundle is left unconfigured in the chart values.

closes #569 

**If applicable**:


deployment/helm/templates/webhook-cert-autogen.yaml
```diff
@@ -0,0 +1,181 @@
...
+  ttlSecondsAfterFinished: 0
+  template:
+    metadata:
+      name: {{ .Values.webhookConfiguration.name }}-patch
+      labels:
+        {{- include "akri.labels" . | nindent 8 }}
+        app.kubernetes.io/component: admission-webhook
+    spec:
+      containers:
+        - name: create
```
jbpaux, line 161: shouldn't be `patch` ?

deployment/helm/templates/webhook-cert-autogen.yaml
```diff
@@ -0,0 +1,181 @@
...
+          image: "registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.1.1"
```
jbpaux, line 162: maybe the image location should be put in the values (for disconnected environments or where there is no connectivity to this registry)

deployment/helm/templates/webhook-cert-autogen.yaml
```diff
@@ -0,0 +1,181 @@
...
+  template:
+    metadata:
+      name: {{ .Values.webhookConfiguration.name }}-create
+      labels:
+        {{- include "akri.labels" . | nindent 8 }}
+        app.kubernetes.io/component: admission-webhook
+    spec:
+      containers:
+        - name: create
+          image: "registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.1.1"
```
jbpaux, line 121: maybe the image location should be put in the values (for disconnected environments or where there is no connectivity to this registry)

kate-goldenring (approved):
LGTM! Thanks @diconico07!

PR #593, opened 2023-04-27
Upgrade tonic and prost

diconico07:
**What this PR does / why we need it**:
This PR initial goal is to upgrade tonic and prost so we can drop the build dependency to rustfmt, thus making it easier to re-build akri in an air-gaped system.

However, to attain this goal it was necessary to do two other unrelated changes (more details about each one after):
 - Upgrade h2
 - Upgrade to edition 2021 of rust

### Upgrade h2 
The only really needed part for tonic and prost is removing the pin to a patched h2 `0.3.3`, however for consistency I also upgraded actix (used in webhook part).
This solves several RUSTSEC issues (some of them already marked as resolved for akri, but a cargo audit still complained):
- Fixes #274 
- Fixes #275 
- Fixes #423
- Fixes #589 
- Fixes #587 

### Upgrade to edition 2021 of rust 
Prost now generates code using 2021 edition of rust, so we need to upgrade this in order to upgrade prost.

**Special notes for your reviewer**:
This might be a hard to read PR, I split it in several commits that should be a bit easier to parse (at least they only upgrade one thing)

**If applicable**:
- [ ] this PR has an associated PR with documentation in [akri-docs](https://github.com/project-akri/akri-docs)
- [ ] this PR contains unit tests
- [x] added code adheres to standard Rust formatting (`cargo fmt`)
- [x] code builds properly (`cargo build`)
- [x] code is free of common mistakes (`cargo clippy`)
- [x] all Akri tests succeed (`cargo test`)
- [x] inline documentation builds (`cargo doc`)
- [x] all commits pass the [DCO bot check](https://probot.github.io/apps/dco/) by being signed off -- see the failing DCO check for instructions on how to retroactively sign commits

PR #596, opened 2023-05-02
build: Fix intermediate dotnet openCV image

### LINKED ISSUE #580: OpenCV Container is Failing to Build
We haven't rebuilt the openCV container in a couple months. The container is [failing to build](https://github.com/project-akri/akri/actions/workflows/build-opencv-base-container.yml) and we likely need to update the Dockerfile which is a two part process, as changes to our intermediate containers must be made in a separate pr: https://github.com/project-akri/akri/blob/main/build/containers/intermediate/Dockerfile.opencvsharp-build#L8. Furthermore, this container takes notoriously long to build so the dev cycle is long on this one.

diconico07:

The image was built using the wrong version of opencvsharp, fix this.

Also switch to shallow clone to reduce build time (a bit).

Should fix #580 

**If applicable**:

### DIFF
--- a/build/intermediate-containers.mk
+++ b/build/intermediate-containers.mk
@@ -1,7 +1,7 @@
 
 BUILD_RUST_CROSSBUILD_VERSION = 0.0.7
 
-BUILD_OPENCV_BASE_VERSION = 0.0.9
+BUILD_OPENCV_BASE_VERSION = 0.0.10
 
 CROSS_VERSION = 0.1.16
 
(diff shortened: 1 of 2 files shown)


jbpaux:
It should be helpful : https://github.com/dotnet/dotnet-docker/issues/1537
I think in longterm we should switch to buildx

diconico07:
Thanks @jbpaux, this should do the trick now ...
:+1:  to switch to buildx (not only for intermediate containers), maybe we can do an issue for that.

And for this one in particular, I don't see much added value in maintaining this image, maybe we can find an alternative to opencv here that doesn't need hours of compilation on our side.

build/intermediate-containers.mk
```diff
@@ -18,15 +18,15 @@ opencv-base: opencv-base-build opencv-base-docker-per-arch
 opencv-base-build: opencv-base-build-amd64 opencv-base-build-arm32 opencv-base-build-arm64
 opencv-base-build-amd64:
 ifeq (1, ${BUILD_AMD64})
-	docker build $(CACHE_OPTION) -f $(INTERMEDIATE_DOCKERFILE_DIR)/Dockerfile.opencvsharp-build . -t $(PREFIX)/opencvsharp-build:$(BUILD_OPENCV_BASE_VERSION)-$(AMD64_SUFFIX) --build-arg PLATFORM_TAG=3.1-buster-slim
+	docker build $(CACHE_OPTION) -f $(INTERMEDIATE_DOCKERFILE_DIR)/Dockerfile.opencvsharp-build . -t $(PREFIX)/opencvsharp-build:$(BUILD_OPENCV_BASE_VERSION)-$(AMD64_SUFFIX) --build-arg PLATFORM=
```
kate-goldenring, line 21: Looks like platform is missing here

```suggestion
	docker build $(CACHE_OPTION) -f $(INTERMEDIATE_DOCKERFILE_DIR)/Dockerfile.opencvsharp-build . -t $(PREFIX)/opencvsharp-build:$(BUILD_OPENCV_BASE_VERSION)-$(AMD64_SUFFIX) --build-arg PLATFORM=-$(AMD64_SUFFIX)
```

kate-goldenring (approved):
Just one nit on a missing platform and then LGTM!

kate-goldenring:
> And for this one in particular, I don't see much added value in maintaining this image, maybe we can find an alternative to opencv here that doesn't need hours of compilation on our side.

+1 to finding a way not to maintain this opencv image

kate-goldenring:
@diconico07 once this is in, we can then merge #594 and then #593, right?

build/intermediate-containers.mk
diconico07: I should add a comment here (and maybe remove the `build-arg` argument as well) this is intended, aspnet amd64 images are not suffixed and the other uses of the `PLATFORM` variable expect empty for amd64.

PR #599, opened 2023-05-02
Add linuxOnly node selector for discovery handlers

### LINKED ISSUE #598: Akri discovery handler deployment on mixed OS clusters
When we deploy Akri on mixed-OS clusters (Linux AND Windows nodes), Akri deployment fails since it can only be deployed on Linux nodes. 

For the agent and controller, there are nodeSelectors in the YAML which we can specify for Linux so that deployment does not error out, but for the discovery handlers, there currently isn't a place we can specify that, so we need to add the **nodeSelector** for the **discovery handlers** in order to prevent deployment from erroring.

jbpaux:
Fix #598

**If applicable**:

### DIFF
--- a/deployment/helm/templates/custom-discovery-handler.yaml
+++ b/deployment/helm/templates/custom-discovery-handler.yaml
@@ -46,9 +46,14 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- if .Values.custom.discovery.nodeSelectors }}
+      {{- if or .Values.custom.discovery.linuxOnly .Values.custom.discovery.nodeSelectors }}
       nodeSelector:
-      {{- toYaml .Values.custom.discovery.nodeSelectors | nindent 8 }}
+        {{- if .Values.agent.linuxOnly }}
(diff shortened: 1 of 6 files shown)


diconico07 (approved):
Looks good to me, I wonder if we can build windows images for some components so we can switch the default to false for these

diconico07:
I think we can hardcode this everywhere then (to stay consistent), this might save us a few confused users wondering why there is such a config knob if changing its value break things.

If we ever want to support other OSes (never tried to build a windows container, so I trust you on that point), then having it as a value can come handy, but we are far from here and we can add it to values when needed.

kate-goldenring:
@jbpaux I agree that hardcoding is the way. In our docs, we explicitly state that we are linux only so I think this is a good way to enforce that. Beyond udev, i believe we'd have to make changes to the way we do uds sockets for Windows as well if we wanted to support it down the line

kate-goldenring (approved):

jbpaux:
It's ready to merge, I've forced all components to use linux node selector. I let the previous commit in case sometime in the future we may want to rollback

PR #602, opened 2023-05-10
agent: Fix possible crash when modifying Configuration

diconico07:
When modifying a Configuration that made the discovery handler drop the connection (e.g an invalid udev rule for udev handler), the finished_discovery_receiver gets None as the connection was dropped by the other end.
This lead to a crash of the agent as the Option was unwrapped here.

Now ignoring the None case to avoid crash.

**If applicable**:


kate-goldenring (approved):
Thanks @diconico07!

PR #618, opened 2023-06-12
Create internal_list_and_watch so it can be mocked

johnsonshih:
Make change to device plugin service to add internal_list_and_watch so we can provide mocked kube_interface to test the list_and_watch.
With original code, we cannot mock kube_interface and had to hard-coded test code to build unhealthy devices using #[cfg(not(test))]/#[cfg(test)].  The new internal_list_and_watch function provide an entry point to inject mocked kube_interface

**If applicable**:


diconico07 (approved):
Looks good to me, I however think we should still do a patch bump even though this is just a refactor (to me the "same version" should be reserved to changes that are only tests or intermediate containers, as all other changes basically have an impact to the runtime environment)

PR #619, opened 2023-06-14
Support discoveryProperties in Configuration

johnsonshih:

Discovery handlers need credential data for authenticated discovery. This PR allows users to specify credential data through discoveryProperties, Agent read the properties and pass them to discovery handlers.

This PR implements the proposal: https://github.com/project-akri/akri-docs/pull/61

**If applicable**:


agent/src/util/discovery_operator.rs
```diff
@@ -1586,4 +1870,795 @@ pub mod tests {
...
+
+        // get_discovery_property_value_from_secret should return error if mandatory secret name is empty
+        let result =
+            get_discovery_property_value_from_secret(&mock_kube_client, namespace_name, &selector)
+                .await;
+        assert!(result.is_err());
+    }
+
+    #[tokio::test]
+    async fn test_get_discovery_properties_value_from_config_map_no_secret_name_optional() {
```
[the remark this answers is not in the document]
johnsonshih: correct typo

agent/src/util/discovery_operator.rs
```diff
@@ -450,6 +477,234 @@ impl DiscoveryOperator {
...
+        namespace: &str,
+    ) -> anyhow::Result<Option<ConfigMap>> {
+        let resource_client: Api<ConfigMap> =
+            Api::namespaced(self.kube_interface.get_kube_client(), namespace);
+        let lp = ListParams::default();
+        for s in resource_client.list(&lp).await? {
+            if name == s.metadata.name.as_ref().unwrap() {
+                return Ok(Some(s));
+            }
+        }
```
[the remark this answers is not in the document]
johnsonshih: use get_opt

agent/src/util/discovery_operator.rs
```diff
@@ -450,6 +477,234 @@ impl DiscoveryOperator {
...
+    async fn get_secret(&self, name: &str, namespace: &str) -> anyhow::Result<Option<Secret>> {
+        let resource_client: Api<Secret> =
+            Api::namespaced(self.kube_interface.get_kube_client(), namespace);
+        let lp = ListParams::default();
+        for s in resource_client.list(&lp).await? {
+            if name == s.metadata.name.as_ref().unwrap() {
+                return Ok(Some(s));
+            }
+        }
+        Ok(None)
```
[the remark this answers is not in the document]
johnsonshih: use get_opt

johnsonshih:
> I also noticed you do the parameters gathering per discoveryEndpoints, it might be more efficient to do it only once at discoveryOperator's creation ?

I thought about that but later changed to ready the secret information at the time establishing the endpoint connection.  Reason:
1) provide as much as up-to-date secret data, in case the secret data changed, this gives agent a chance to refresh the secret data. 
2) DiscoveryOperator is a long run entity, don't want to keep the secret data in memory for a long time.

agent/src/util/discovery_operator.rs
```diff
@@ -450,6 +477,234 @@ impl DiscoveryOperator {
...
+            get_discovery_property_value_from_secret(
+                kube_client,
+                self.config.metadata.namespace.as_ref().unwrap(),
+                secret_key_selector,
+            )
+            .await
+        } else {
+            let error = Error::new(ErrorKind::InvalidInput, "no supported value_from found");
+            Err(error.into())
+        }
```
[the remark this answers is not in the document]
johnsonshih: will add validation in webhook

webhooks/validating/configuration/src/main.rs
```diff
@@ -789,6 +860,256 @@ mod tests {
...
+        let resp = run_validate_configuration_discovery_properties(discovery_properties);
+        assert!(resp.allowed);
+    }
+
+    #[test]
+    fn test_validate_configuration_discovery_properties_plain_text_empty_name() {
+        let discovery_properties = r#"
+        "discoveryProperties": [
+            {
+                "name": ""
```
bfjelds, line 933: curious, why this would be success case?

webhooks/validating/configuration/src/main.rs
```diff
@@ -789,6 +860,256 @@ mod tests {
...
+                    "fieldRef": {
+                        "fieldPath": "ffff"
+                    }
+                }
+            }
+        ],"#;
+
+        // valueFrom discovery properties, unknown ref should fail
+        run_validate_configuration_discovery_properties(discovery_properties);
+    }
```
bfjelds, line 990: should there be a `assert!(!resp.allowed);` here?

webhooks/validating/configuration/src/main.rs
```diff
@@ -789,6 +860,256 @@ mod tests {
...
+        let discovery_properties = r#"
+        "discoveryProperties": [
+            {
+                "value":"vvvv"
+            }
+        ],
+        "#;
+
+        // plain text discovery properties without name should fail, missing field 'name'
+        run_validate_configuration_discovery_properties(discovery_properties);
```
bfjelds, line 954: should there be a `assert!(!resp.allowed);` here?

webhooks/validating/configuration/src/main.rs
```diff
@@ -789,6 +860,256 @@ mod tests {
...
+        "discoveryProperties": [
+            {
+                "name": "nnnn",
+                "valueFrom": {
+                }
+            }
+        ],"#;
+
+        // valueFrom discovery properties, not specified should fail, missing content of 'valueFrom'
+        run_validate_configuration_discovery_properties(discovery_properties);
```
bfjelds, line 970: should there be a `assert!(!resp.allowed);` here?

webhooks/validating/configuration/src/main.rs
```diff
@@ -789,6 +860,256 @@ mod tests {
...
+                    "configMapKeyRef": {
+                        "name": "nnnn1",
+                        "key": "kkk",
+                        "optional": false
+                    }
+                }
+            }
+        ],"#;
+
+        // both value and valueFrom are specified should success
```
bfjelds, line 1051: curious why setting value & valueFrom is valid?

shared/src/akri/configuration.rs
```diff
@@ -18,6 +20,30 @@ use std::collections::HashMap;
...
+
+/// DiscoveryProperty represents a property for discovery devices
+#[derive(Serialize, Deserialize, Clone, Debug, JsonSchema, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct DiscoveryProperty {
+    /// Name of the discovery property
+    pub name: String,
+
+    /// value of the discovery property
+    pub value: Option<String>,
```
bfjelds, line 41: `#[serde(default, skip_serializing_if = "Option::is_none")]` ?

bfjelds: for value and value_from?

deployment/helm/crds/akri-configuration-crd.yaml
```diff
@@ -22,6 +22,12 @@ spec:
                       type: string
                     discoveryDetails:
                       type: string
+                    discoveryProperties:
+                      nullable: true
+                      type: array
+                      items: # {{DiscoveryProperty}}
+                        x-kubernetes-preserve-unknown-fields: true
```
bfjelds, line 29: why not define the DiscoveryProperty schema here?

agent/src/util/discovery_operator.rs
```diff
@@ -450,6 +475,223 @@ impl DiscoveryOperator {
...
+                vec: Some(v.0.clone()),
+            }));
+        }
+    }
+
+    // secret key/value not found
+    if optional {
+        Ok(None)
+    } else {
+        Err(Error::new(ErrorKind::InvalidInput, "secret data not found").into())
```
bfjelds, line 644: nit: might be helpful to have a tiny bit more context in error messages.  something like "discovery property's referenced secret ..."  (same for ConfigMap code below ... also, might use k8s capitalization rather than rust ... i.e. ConfigMap rather than config_map)

agent/src/util/discovery_operator.rs
```diff
@@ -1586,4 +1857,775 @@ pub mod tests {
...
+            DiscoveryProperty {
+                name: property_name_1.clone(),
+                value: None,
+                value_from: None,
+            },
+            DiscoveryProperty {
+                name: property_name_2.clone(),
+                value: None,
+                value_from: None,
+            },
```
bfjelds, line 1907: it isn't clear to me why name="", value=None, value_from=None is valid?

webhooks/validating/configuration/src/main.rs
johnsonshih: the json is invalid, so #[should_panic(expected = "Could not parse as Akri Configuration")]

webhooks/validating/configuration/src/main.rs
johnsonshih: the json is invalid, so #[should_panic(expected = "Could not parse as Akri Configuration")]

webhooks/validating/configuration/src/main.rs
johnsonshih: the json is invalid, so #[should_panic(expected = "Could not parse as Akri Configuration")]

webhooks/validating/configuration/src/main.rs
johnsonshih: we don't restrict that, if both value and valueFrom are specified, value wins

webhooks/validating/configuration/src/main.rs
johnsonshih: we don't restrict empty name, as long as it's not None, we accept it (although it probably won't be used when DH query the secret data).

agent/src/util/discovery_operator.rs
johnsonshih: we don't restrict empty name, as long as it's not None, we accept it.  For value/valueFrom, they can be None by schema definition

shared/src/akri/configuration.rs
johnsonshih: add serde attributes for value and value_from

webhooks/validating/configuration/src/main.rs
bfjelds: ah!  thanks ... i didn't pay attention to the attributes.

webhooks/validating/configuration/src/main.rs
bfjelds: seems strange to me to allow an empty name

agent/src/util/discovery_operator.rs
johnsonshih: add more context to the error messages

deployment/helm/crds/akri-configuration-crd.yaml
johnsonshih: yes, why not.  I'll update the crd

webhooks/validating/configuration/src/main.rs
johnsonshih: I had updated the schema so it's validated by by x-kubernetes-validations via OpenAPI Schema.  If both value and valueFrom are specified, the validator reject it.

webhooks/validating/configuration/src/main.rs
bfjelds: i think an empty name should be an error, but won't block the PR based on that.

bfjelds (approved):

webhooks/validating/configuration/src/main.rs
johnsonshih: if we treat empty name as error, should we treat name with all spaces as error, too?  IMO, from syntax perspective, an empty string is a valid string and is valid as a key in key-value pair.  We can allow it.
BTW, a null string for name is for sure not allowed, it's specified in the schema

bfjelds: i dont think an empty string and a string with spaces are the same.

bfjelds (approved):

webhooks/validating/configuration/src/main.rs
diconico07: I'd be in favor of specifying name in the same way it is defined in core `EnvVar`, that is as a `C_IDENTIFIER` (i.e a string conforming to this regex: `[A-Za-z_][A-Za-z0-9_]*` thus disallowing empty names and ensuring these discovery properties can be passed as environment variable if need arise.

I don't know if there is a way to translate that in the CRD or if we can only rely on the webhook to reject configurations that are in this case.

agent/src/util/discovery_operator.rs
diconico07: Doesn't the `oneOf` section of the CRD requires that either `value` or `valueFrom` is specified ?

diconico07 (approved):
I won't block the PR for not triggering errors on empty names, but please create an issue for doing that later if you don't do it in this PR.

webhooks/validating/configuration/src/main.rs
johnsonshih: yes, we can put pattern in the crd and not relies on webhook to validate the content.  I'll investigate if we really want to make that hard restriction to the key name.

johnsonshih: filed Issue for the investigation effort.
https://github.com/project-akri/akri/issues/634

agent/src/util/discovery_operator.rs
johnsonshih: yes, switch to use schema, the oneOf property makes the requirement that either value or valueFrom required.

PR #621, opened 2023-06-14
Split instance device plugin from device plugin service

johnsonshih:
This PR split the instance specific attributes and behaviors from the DevicePluginService into a separate class InstanceDevicePlugin.  This is to prepare for the new Configuration device plugin type (call ConfigurationDevicePlugin) that shares the common data offered by DevicePluginService but can have different behavior for device plugin at Instance and Configuration levels.

**If applicable**:


agent/src/util/device_plugin_service.rs
```diff
@@ -73,8 +78,6 @@ pub type InstanceMap = Arc<RwLock<HashMap<String, InstanceInfo>>>;
 pub struct DevicePluginService {
```
kate-goldenring, line 78: I am not sure what the final implementation will look like but it feels strange to still keep both `instance_name` and `config_name` in the shared `DevicePluginService` struct. Should we have one `name` field and move the other to the `DevicePluginBehavior` struct? We can also leave as is for now

kate-goldenring (approved):
One comment about how it could be a little cleaner but you could ignore it. LGTM!

Issue #597, opened 2023-05-02
Onvif discovery Handler incorrectly filter ip address


johnsonshih:
**Describe the bug**
set filter ip address to exclude 10.0.0.1 will end up filtering all ip addresses start with 10.0.0.1, including 10.0.0.1xx.

**To Reproduce**
1. Create cluster
2. Install Akri with the Helm command and deploy the Akri Configuration yaml modified from test/yaml/akri-onvif-video-configuration.yaml (see below).  In the configuration, only exclude cameras with ip address 10.0.0.1, 10.0.0.2.
3. My Onvif camera ip is 10.0.0.107, after deploying the Configuration, my camera is not detected.

apiVersion: akri.sh/v0
kind: Configuration
metadata:
  name: akri-onvif-video
spec:
  discoveryHandler:
    name: onvif
    discoveryDetails: |+
        ipAddresses:
          action: Exclude
          items:
          - 10.0.0.1
          - 10.0.0.2
        macAddresses:
          action: Exclude
          items: []
        scopes:
          action: Exclude
          items: []
        discoveryTimeoutSeconds: 2
  brokerSpec:
    brokerPodSpec:
      containers:
      - name: akri-onvif-video-broker
        image: "ghcr.io/project-akri/akri/onvif-video-broker:latest-dev"
        imagePullPolicy: Always
        resources:
          limits:
            "{{PLACEHOLDER}}" : "1"
  instanceServiceSpec:
    ports:
    - name: grpc
      port: 80
      targetPort: 8083
  configurationServiceSpec:
    ports:
    - name: grpc
      port: 80
      targetPort: 8083
  brokerProperties: {}

**Expected behavior**
Onvif camera ip address 10.0.0.107 discovered.
Onvif discovery handler should only filter out 10.0.0.1 and 10.0.0.2, not others.

johnsonshih:
The fix causes wrong behavior when filtering scopes.

johnsonshih:
verify the issue is fixed.

PR #622, opened 2023-06-15
correct filters

johnsonshih:
This PR fixes the issue. https://github.com/project-akri/akri/issues/597
The issue was fixed by https://github.com/project-akri/akri/pull/604, but that change causes regression when filter scopes.   This PR contains a correct fix for onvif ip and mac address filters.

**If applicable**:


discovery-handlers/onvif/src/discovery_impl.rs
```diff
@@ -273,6 +284,90 @@ pub mod util {
...
....org/ver20/media/wsdl\" xmlns:ter=\"http://www.onvif.org/ver10/error\" xmlns:tns1=\"http://www.onvif.org/ver10/topics\" xmlns:tnsn=\"http://www.eventextension.com/2011/event/topics\"><SOAP-ENV:Header><wsa:MessageID>urn:uuid:2bc6f06c-5566-7788-99ac-0012414fb745</wsa:MessageID><wsa:RelatesTo>uuid:7b1d26aa-b02e-4ad2-8aab-4c928298ee0c</wsa:RelatesTo><wsa:To SOAP-ENV:mustUnderstand=\"true\">http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous</wsa:To><wsa:Action SOAP-ENV:mustUnderstand=\"true\">http://schemas.xmlsoap.org/ws/2005/04/discovery/ProbeMatches</wsa:Action></SOAP-ENV:Header><SOAP-ENV:Body><d:ProbeMatches><d:ProbeMatch><wsa:EndpointReference><wsa:Address>urn:uuid:10919da4-5566-7788-99aa-0012414fb745</wsa:Address></wsa:EndpointReference><d:Types>dn:NetworkVideoTransmitter</d:Types><d:Scopes>onvif://www.onvif.org/type/video_encoder onvif://www.onvif.org/type/audio_encoder onvif://www.onvif.org/hardware/IPC-model onvif://www.onvif.org/location/country/china onvif://www.onvif.org/name/NVT onvif://www.onvif.org/Profile/Streaming </d:Scopes><d:XAddrs>{}</d:XAddrs><d:MetadataVersion>10</d:MetadataVersion></d:ProbeMatch></d:ProbeMatches></SOAP-ENV:Body></SOAP-ENV:Envelope>",
```
diconico07, line 298: I really am not fond of that big not easily parsed one line string that get repeated multiple time.
Maybe put that in a small function that will return to you the formatted string. 
And you can, I think make it more easily readable by prettifying it.
A raw string literal will also help to avoid escaping all the `"`

diconico07 (changes requested):

discovery-handlers/onvif/src/discovery_handler.rs
```diff
@@ -168,6 +168,7 @@ async fn apply_filters(
     if util::execute_filter(
         discovery_handler_config.ip_addresses.as_ref(),
         &ip_address_as_vec,
+        |scope, pattern| scope == pattern,
```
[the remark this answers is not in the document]
johnsonshih: sure, will do

discovery-handlers/onvif/src/discovery_impl.rs
johnsonshih: sure, will do.

discovery-handlers/onvif/src/discovery_handler.rs
```diff
@@ -583,4 +586,64 @@ mod tests {
             .await
             .is_none());
     }
+
+    #[tokio::test]
+    async fn test_apply_filters_include_mac_exist_different_cases() {
```
kate-goldenring, line 591: nit (and for other test)
```suggestion
    async fn test_apply_filters_include_mac_exist_different_letter_cases() {
```

kate-goldenring (commented):
@johnsonshih this seems like a fine change to ignore casing but i am not sure how this fixes https://github.com/project-akri/akri/issues/597 given how the issue as described is only number IP address filters. Can you update the PR description or create a new issue that discusses this new bug with letter casing?

discovery-handlers/onvif/src/discovery_handler.rs
johnsonshih: rename function

johnsonshih:
> @johnsonshih this seems like a fine change to ignore casing but i am not sure how this fixes #597 given how the issue as described is only number IP address filters. Can you update the PR description or create a new issue that discusses this new bug with letter casing?

I had updated the PR description to explain why/what this change is.  For filtering IP address using case-insensitive comparison,   it doesn't have any impact when filter IPv4 addresses, but in case we add IPv6 support for onvif DH, we don't need to change the filter.
I'm opened to rollback to use exact match as currently onvif DH only supports IPv4 address.

diconico07 (approved):

PR #623, opened 2023-06-16
Revamp e2e test suite

diconico07:
Move to pytest based test-suite, and ensure tests can be run more easily locally.
The main reasons to move to pytest are:
 - Allows to run a subset of the tests rather than full suite
 - Ensure better separation of tests vs helpers
 - Easily share code for setup/teardown between suites

This commit also change dependency management to use poetry, this allows to setup the test environment locally in an easy and predictable way, it also ensure more reproducibility for our test-suite as the dependencies are pinned.

On a more functional end, switch to slim agent to allow test suites to chose their discovery handlers. Also stop using helm for configuration deployment to decorrelate it from akri's installation, preventing possible noise in tests.

Also makes use of watch instead of polling as much as possible to reduce waiting time (goes from about 3 minutes for e2e run to about 80 seconds).

Hopefully this will make it easier to write more e2e tests.

**If applicable**:


diconico07:
Currently this is still WIP, the webhook suite is not migrated, the old suite is not deleted and this will also need a proper documentation PR, so I send this as a draft to get some comments about this, so please still take a look and tell me.

test/e2e/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[tool.poetry]
+name = "akri-e2e"
+version = "0.10.10"
+description = "Akri end to end test framework"
+authors = ["Your Name <you@example.com>"]
```
kate-goldenring, line 5: maybe just remove the authors line

```suggestion
```

.github/workflows/run-test-cases.yml
```diff
@@ -296,60 +307,23 @@ jobs:
...
         run: |
-          git fetch origin main
-          git show origin/main:version.txt > /tmp/version_to_test.txt
-          echo '--set agent.image.pullPolicy=Never,agent.image.tag=pr-amd64,controller.image.pullPolicy=Never,controller.image.tag=pr-amd64,webhookConfiguration.image.pullPolicy=Never,webhookConfiguration.image.tag=pr-amd64' > /tmp/extra_helm_args.txt
-      # For non-PR (i.e. push, release, manual), version.txt is corresponds
-      # to an existing Helm chart.
-      - if: (!(startsWith(github.event_name, 'pull_request')))
-        name: Use current version for push
-        run: cat version.txt > /tmp/version_to_test.txt
+          case "${{ github.event_name }}" in "push");; "release") extra_param="--release";; *) extra_param="--use-local";; esac
```
kate-goldenring, line 316: We may need to ensure that on releases, this only runs once `run-helm.yaml` completes, otherwise this may pull the previous releases charts

test/e2e/conftest.py
```diff
@@ -0,0 +1,120 @@
...
+    name: str
+    kubeconfig: Path
+    kubectl: str
+
+
+@pytest.fixture(scope="session")
+def distribution_config(pytestconfig):
+    distribution = pytestconfig.getoption("--distribution", None)
+    if distribution == "k3s":
+        yield Distribution("k3s", Path.home() / ".kube/config", "kubectl")
```
kate-goldenring, line 33: Should we update the tests and workflow to match the distribution defaults to better enable running the tests locally without having to copy over the kubeconfig?
```suggestion
        yield Distribution("k3s", Path.home() / "/etc/rancher/k3s/k3s.yaml", "kubectl")
```

kate-goldenring (commented):
Thank you for putting this together. This is a very clean setup. Just took a brief look -- will look back at it.

test/e2e/conftest.py
diconico07: I'm not sure of the added benefit for local runs, as I think most people will either have a long running instance they use for tests (so probably already configured) or use tools like k3d or rancher desktop that already manage the configuration.

In fact I first wanted to just get rid of that element here (and let the python kubernetes client handle it the same way the `helm` or `kubectl` commands do), but wasn't really sure it was the right thing to do.

test/e2e/pyproject.toml
diconico07: The authors field is mandatory, maybe I should just put my name here, I think we should get a policy for managing these authors fields (if I remember correctly all `Cargo.toml` mention you as author).

But definitely this placeholder is just me forgetting to fill it :innocent:

.github/workflows/run-test-cases.yml
diconico07: The `run-helm` is not the only one we need to wait for, there are also all the `build-.*-container`

I used the same way as before for that (just not at the same place), a.k.a `sleep 2700`, it now is in an action step, while it was in the python script before.

(quite unrelated but made me notice I didn't change the paths to look for to trigger e2e test run, so will do that)

johnsonshih (approved):

kate-goldenring (approved):
This is so nice. Thank you for putting this together, Nicolas!

PR #626, opened 2023-06-28
helm: Unify labels and add recommended labels

diconico07:
We currently don't add the recommended labels on our helm deployed resources, we even don't have consistency among the different components of akri (e.g controller has an "app" one and agent and discovery handlers have a "name" one).

This commit aims to solve both issues by using the recommended labels everywhere they are relevant (namely the `app.kubernetes.io` ones, like `part-of`, `name` and `component`)

We may want to take special care in testing the upgrade path, as there might be some breakage here (I did not test that path yet)

**If applicable**:


kate-goldenring (approved):
Thank you @diconico07!

PR #627, opened 2023-07-03
Add Configuration device plugin

johnsonshih:
This PR is to implement the proposal of exposing resource at Configuration level. https://github.com/project-akri/akri-docs/blob/main/proposals/configuration-level-resources.md#configuration-level-resources
With Configuration level resource, users can select resource to use at configuration level without knowing the instance id beforehand.
The implementation is described in the document PR:https://github.com/project-akri/akri-docs/pull/76

Summary of changes in this PR
- CL resource and IL resource share the capacity pool. i.e. (# of allocated CL virtual devices + # of allocated IL virtual devices) <= capacity.
- The name of CL device plugin is the Akri Configuration name and follows the same convention of IL device plugin, i.e., replace ['.', '/'] with "-".
- The CL device plugin uses dynamic name for virtual device ids. The virtual device id reported by CL device plugin looks like "0", "1", "2", ....  The actual device usage slot being used is selected when ConfigurationDevicePlugin::allocate() is called.
- ConfigurationDevicePlugin represent the behavior of CL device plugin.  InstanceDevicePlugin represents the behavior of IL device plugin.
- DevicePluginService contains a list_and_watch_message_senders to notify refreshing list_and_watch, used by the DevicePluginService internally, a copy of list_and_watch_message_sender is stored in the associated InstanceInfo, used by external entity to refresh the InstanceDevicePlugin's list_and_watch.
- ConfigurationDevicePlugin contains a list_and_watch_message_sender to notify refreshing list_and_watch, used by the ConfigurationDevicePlugin internally, a copy of list_and_watch_message_sender is store in the InstanceConfig, used by external entity to refresh the ConfigurationDevicePlugin's list_and_watch
- When IL DPS allocate a virtual device, it notifies CL DPS to refresh list_and_watch, and vice versa, CL DPS notify IL DPS to refresh list_and_watch when it allocates a virtual device.
- 
**If applicable**:


kate-goldenring:
For history of discussions around this work, see previous PR https://github.com/project-akri/akri/pull/565

agent/src/util/device_plugin_service.rs
```diff
@@ -247,10 +295,27 @@ impl InstanceDevicePlugin {
...
+            )
+            .await;
+
+            let virtual_devices = device_usage_state
+                .into_iter()
+                .map(|(id, state)| v1beta1::Device {
+                    id,
+                    health: match state {
+                        DeviceUsageStatus::Free | DeviceUsageStatus::ReservedByInstance => {
+                            HEALTHY.to_string()
```
bfjelds, line 313: might be helpful to add some comments here, why Free and ReservedByInstance are HEALTHY, but ReservedByConfiguration is considered UNHEALTHY

bfjelds: i think this might be difficult to maintain without some of the documentation (https://github.com/project-akri/akri-docs/blob/725ec5e202062787bda2257ee2260dc12442970a/proposals/configuration-level-resources.md) being in the code.

bfjelds:
this is so awesome to see!

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
-            "Could not find device usage slot",
-        ))
+        Err(_) => (0..*capacity)
+            .map(|x| {
+                (
+                    format!("{}-{}", instance_name, x),
+                    DeviceUsageStatus::Unknown,
+                )
+            })
+            .collect(),
```
bfjelds, line 841: would it be easier to surface possible bugs if we bubbled up an error here instead of DeviceUsageStatus::Unknown?  Is there a valid state where Unknown is expected?

agent/src/util/device_plugin_service.rs
```diff
@@ -1637,7 +2233,848 @@ mod device_plugin_service_tests {
...
+            this_node,
+            instance_namespace,
+            &capacity,
+            Arc::new(mock),
+        )
+        .await;
+        assert!(result.is_empty());
+    }
+
+    // Tests 0 virtual device id is returned if instance not found from kube find_instance
```
bfjelds, line 2338: there are a few tests to validate get_available_virtual_devices that return 0.  it seems like there should be a test that returns an expected non-zero number of results.

bfjelds: nevermind ... i see it: test_get_available_virtual_devices_all_free

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+            instance_namespace,
+            capacity,
+            kube_interface.clone(),
+        )
+        .await;
+        device_usage_states.insert(instance_name.to_string(), device_usage_state);
+    }
+    let mut vdev_ids_to_report = HashSet::new();
+    let mut free_instances = HashSet::new();
+    let mut reserved_instances = HashSet::new();
```
bfjelds, line 866: why does `reserved_instances` exist?

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
-///     * (ACTION) return true
-/// 2. device_usage\[id\] == self.nodeName ... this means THIS node previously used id, but the DevicePluginManager knows that this is no longer true
-///     * (ACTION) return false
-/// 3. device_usage\[id\] == (some other node) ... this means that we believe this device is in use by another node and should be marked unhealthy
-///     * (ACTION) return error
-/// 4. No corresponding id found ... this is an unknown error condition (BAD)
-///     * (ACTION) return error
-fn slot_available_to_reserve(
-    device_usage_id: &str,
+pub async fn get_instance_device_usage_state(
```
bfjelds, line 812: nit: since this returns a vec, it seems like it should be plural: `get_instance_device_usage_states`

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
-                trace!(
-                    "try_update_instance_device_usage - could not find Instance {}",
-                    instance_name
-                );
-                return Err(Status::new(
-                    Code::Unknown,
+) -> HashSet<String> {
+    let mut device_usage_states = HashMap::new();
+    for instance_name in instance_config.instances.keys() {
+        let device_usage_state = get_instance_device_usage_state(
```
bfjelds, line 854: nit: `instance_device_usage_states`

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+            capacity,
+            kube_interface.clone(),
+        )
+        .await;
+        device_usage_states.insert(instance_name.to_string(), device_usage_state);
+    }
+    let mut vdev_ids_to_report = HashSet::new();
+    let mut free_instances = HashSet::new();
+    let mut reserved_instances = HashSet::new();
+    for (instance_name, device_usage_state) in device_usage_states {
```
bfjelds, line 867: nit: `instance_device_usage_states`

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+                DeviceUsageStatus::ReservedByConfiguration(vdev_id) => {
+                    reserved_instances.insert(instance_name.clone());
+                    vdev_ids_to_report.insert(vdev_id);
+                }
+                _ => (),
+            };
+        }
+    }
+
+    // Decide virtual device id, report already reserved virtual devices + 1 free device usage (if available) per instance
```
bfjelds, line 882: why are `already reserved virtual devices` returned from a function named get_available_virtual_devices?  are they available?  should this function be renamed to something like get_virtual_devices_not_used_by_other_nodes?  or maybe that is just what "available" means in this context.  either way, more comments would help.

bfjelds: maybe available=healthy?

can this function include this and return healthy devices to simplify ConfigurationDevicePlugin.list_and_watch?

```
            let current_virtual_devices = available_virtual_devices
                .into_iter()
                .map(|id| (id, HEALTHY.to_string()))
                .collect::<HashMap<String, String>>();
```

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+    }
+
+    let mut usage_ids_to_use = HashMap::<String, (String, String)>::new();
+
+    // Get all available virtual devices, group by device usage status,
+    // a virtual device is available if it's Free or ReservedByConfiguration with requested vdev_ids
+    let mut free_device_usage_states = HashMap::new();
+    let mut reserved_by_configuration_usage_states = HashMap::new();
+    for (instance_name, device_usage_state) in device_usage_states {
+        for (id, state) in device_usage_state {
```
bfjelds, line 924: nit: change `id` to `device_name`?

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+                            .insert(id, vdev_id);
+                    }
+                }
+
+                _ => (),
+            };
+        }
+    }
+
+    // Find (Instance, usage slot) for vdev ids, (Instance, usage slot) in free_device_usage_states and
```
bfjelds, line 946: i find it a bit confusing to see id, device, vdev, and slot ... i'm not certain what is equivalent (it looks like what is refered to as `id` above is `slot` below) and what is different.  it would be helpful (if there are synonyms) to use just one.

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+                }
+                if resource.is_some() {
+                    break;
+                }
+            }
+
+            if resource.is_none() {
+                if let Some((inst, slots)) = &free_device_usage_states
+                    .iter()
+                    .max_by(|a, b| a.1.len().cmp(&b.1.len()))
```
bfjelds, line 971: curious: what does this max_by line do?

bfjelds: is it finding the instance with the most free slots?

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+                {
+                    if let Some(slot) = slots.iter().next() {
+                        resource = Some((inst.to_string(), slot.to_string()));
+                    }
+                }
+            }
+
+            match resource {
+                Some((inst, slot)) => {
```
bfjelds, line 980: nit: use `instance_name` rather than `inst`

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+                    free_device_usage_states.remove(&inst);
+                    usage_ids_to_use.insert(vdev_id.clone(), (inst, slot));
+                    false
+                }
+                None => true,
+            }
+        })
+        .collect::<Vec<String>>();
+
+    // Return error if any pending requested vdev id
```
bfjelds, line 991: i find 'pending' to be a little confusing here.  i think, in this context, pending means that the device plugin failed to allocate because there weren't enough free slots?  if so, maybe pending can be renamed to unallocated or unallocatable or something like that?

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+        let device_usage = DeviceUsage::from_str(device_usage_string.unwrap()).map_err(|_| {
+            Status::new(
+                Code::Unknown,
+                format!(
+                    "Fails to parse {} to DeviceUsage ",
+                    device_usage_string.unwrap()
+                ),
+            )
+        })?;
+        match get_device_usage_state(&device_usage, node_name) {
```
bfjelds, line 1078: might be nice to pull this `match ...` into a function and add some tests to validate what happens when Free is encountered vs when Configuration or Instance (for this node and for other nodes) vs when ReservedForOthers.

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
-            // Instance slot already reserved for this node
-            return Ok(());
-        }
+            DeviceUsageStatus::ReservedByConfiguration(_) => {
+                if matches!(device_usage_kind, DeviceUsageKind::Configuration(_)) {
+                    return Ok(());
+                } else {
+                    return Err(Status::new(
+                        Code::Unknown,
+                        "Requested device already in use",
```
bfjelds, line 1110: nit: add some context ... maybe "Requested device of type {device_usage_kind} already in use by ConfigurationLevel"

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+                    ));
+                }
+            }
+            DeviceUsageStatus::ReservedByInstance => {
+                if matches!(device_usage_kind, DeviceUsageKind::Instance) {
+                    return Ok(());
+                } else {
+                    return Err(Status::new(
+                        Code::Unknown,
+                        "Requested device already in use",
```
bfjelds, line 1120: nit: add some context ... maybe "Requested device of type {device_usage_kind} already in use by InstanceLevel"

agent/src/util/device_plugin_service.rs
```diff
@@ -403,100 +809,337 @@ impl InstanceDevicePlugin {
...
+                    ));
+                }
+            }
+            DeviceUsageStatus::ReservedByOtherNode => {
+                trace!("try_update_instance_device_usage - request for device slot {} previously claimed by a diff node {} than this one {} ... indicates the device on THIS node must be marked unhealthy, invoking ListAndWatch ... returning failure, next scheduling should succeed!",
+                    device_usage_id, device_usage.get_usage_name(), node_name);
+                return Err(Status::new(
+                    Code::Unknown,
+                    "Requested device already in use",
```
bfjelds, line 1129: nit: add some context ... maybe "Requested device already in use by another node"

bfjelds:
Regarding `report already reserved virtual devices + 1 free device usage (if available) per instance`, it seems to be a design decision to only allow CL to claim resources from differing nodes (so requesting 2 CL resources would only work if there were 2 nodes with 1 healthy slot ... 1 node with 2 healthy slots could not be scheduled to).

I assume the purpose of that is to ensure high availability.  And certainly, that provides it from a node perspective.

I'm not sure I'd make that a hard requirement though ... or at least allow the user to pick if they want it to be a requirement or a preference.  For example, if there was a configuration that specifieid NodeHA=preference (rather than requirement) and there was a single node with 2 healthy slots, the user could schedule their workload.  This might provide HA from a device API perspective (maybe the camera has glitchy software) or from a network perspective (maybe the network connection to the camera isn't great).

bfjelds:
overall, this looks super awesome and am very excited about you bringing this feature in!!!

agent/src/util/device_plugin_service.rs
johnsonshih: that's correct

agent/src/util/device_plugin_service.rs
johnsonshih: that's correct.  Before adding CL resource support, a slot can either be 
free,
allocated by Instance dp on this node or 
allocated by other nodes.

for Instance dp, a slot is considerred healthy if it's free or allocated by Instance dp on this node

After CL resource support, there are two types of device plugin (Instance and Configuration dp) running on a node. A slot can be 
free, 
allocated by Instance dp on this node
allocated by Configuration dp on this node
allocated by other nodes

for Instance dp, a slot is considered healthy if it's free or allocated by Instance dp on this node
for Configuration dp, a slot is considered healthy if it's free or allocated by Configuration dp on this node

agent/src/util/device_plugin_service.rs
johnsonshih: Unknown is a valid state, mostly due to the akrii is deleted and the Agent hasn't update instance map

agent/src/util/device_plugin_service.rs
johnsonshih: no, remove

agent/src/util/device_plugin_service.rs
bfjelds: i think a comment for list_and_watch with that info would be quite useful

agent/src/util/device_plugin_service.rs
bfjelds: thanks!  sorry for repeating the Unknown questions in this PR and the other one.  are both PRs going to be merged or is this the one PR that will be merged?

bfjelds: (just trying to figure out where best to focus energy)

agent/src/util/device_plugin_service.rs
johnsonshih: rename id to usage_id

agent/src/util/device_plugin_service.rs
johnsonshih: use consistent name

agent/src/util/device_plugin_service.rs
johnsonshih: rename inst to instance_name

agent/src/util/device_plugin_service.rs
johnsonshih: rename pending_device_ids to unallocated_device_ids

agent/src/util/device_plugin_service.rs
johnsonshih: we have test cases to cover that

agent/src/util/device_plugin_service.rs
johnsonshih: update error messages

agent/src/util/device_plugin_service.rs
johnsonshih: update messages

agent/src/util/device_plugin_service.rs
johnsonshih: update error messages

agent/src/util/device_plugin_service.rs
johnsonshih: rename variable

agent/src/util/device_plugin_service.rs
johnsonshih: rename variable

agent/src/util/device_plugin_service.rs
johnsonshih: rename to get_instance_device_usage_states

agent/src/util/device_plugin_service.rs
johnsonshih: I prefer to keep get_available_virtual_devices() neutral and have list_and_watch handle the returned result so the function get_available_virtual_devices() can be reused

bfjelds: makes sense, the reason i suggested it is that the function is not reused.

agent/src/util/device_plugin_service.rs
johnsonshih: add comment

agent/src/util/device_plugin_service.rs
johnsonshih: since the change scope of this PR is too big, I extracted changes for device usage to a separate PR.  Let's focus on the device usage PR first and review this one after the device usage PR committed.
https://github.com/project-akri/akri/pull/628

agent/src/util/device_plugin_service.rs
johnsonshih: Single responsibility principle :)

bfjelds: lol ... indeed.  unless the function is renamed to get_healthy_devices

johnsonshih:
> Regarding `report already reserved virtual devices + 1 free device usage (if available) per instance`, it seems to be a design decision to only allow CL to claim resources from differing nodes (so requesting 2 CL resources would only work if there were 2 nodes with 1 healthy slot ... 1 node with 2 healthy slots could not be scheduled to).
> 
> I assume the purpose of that is to ensure high availability. And certainly, that provides it from a node perspective.
> 
> I'm not sure I'd make that a hard requirement though ... or at least allow the user to pick if they want it to be a requirement or a preference. For example, if there was a configuration that specifieid NodeHA=preference (rather than requirement) and there was a single node with 2 healthy slots, the user could schedule their workload. This might provide HA from a device API perspective (maybe the camera has glitchy software) or from a network perspective (maybe the network connection to the camera isn't great).

The algorithm of CL resource allocation ensures allocating slots from different devices for a container on a given node.  For example, 2 devices are discovered and the capacity is 3.  On nodeA, a container can allocate 1 or 2 slots and you can have multiple containers scheduled to nodeA (e.g. 6 container, each requests 1 slot, or 3 containers each requests 2 slots, or combination of 1-slot and 2-slot containers).  On nodeB, the same algorithm is used and both nodes share the total 6 slots.

The purpose of `report already reserved virtual devices + 1 free device usage (if available) per instance` is to reduce the change that kubelet allocate slots more than available devices.  For the example above, 2 devices are discovered and the capacity is 3, we can report 2 * 3 slots or 2 slots first and increase the available slots later.

In the case of reporting 2 * 3 slots, if a container requests 4 CL resources, kubelet issue allocate request for the container and Agent will fail the request due only 2 devices are available.  This end up in a infinite loop that kubelet keeps retry to allocate for 4 CL resources until another 2 devices discovered.

If we report 2 slots and increase the available slots later, then the container request 4 CL resources won't be scheduled until the reported available slots exceeds 4. This will reduce the chance that kubelet entering the infinite loop allocating resources for the container.

agent/src/util/device_plugin_service.rs
```diff
@@ -58,12 +63,27 @@ pub struct InstanceInfo {
...
+    /// Instance's hash id
+    pub instance_id: String,
+    /// Device that the instance represents.
+    /// Contains information about environment variables and volumes that should be mounted
+    /// into requesting Pods.
+    pub device: Device,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct InstanceConfig {
```
bfjelds, line 75: what is the thinking behind InstanceConfig as a name?  configuration and instance have pretty specific meanings in akri and this combines them in a way that isn't super-intuitive to me.  

maybe something like DevicePluginContext?

agent/src/util/device_plugin_service.rs
```diff
@@ -58,12 +63,27 @@ pub struct InstanceInfo {
...
+    /// Sender to tell Configuration device plugin `list_and_watch` to either prematurely continue looping or end
+    pub usage_update_message_sender: Option<broadcast::Sender<ListAndWatchMessageKind>>,
+    /// Map of all Instances from the same Configuration
+    pub instances: HashMap<String, InstanceInfo>,
 }
 
-pub type InstanceMap = Arc<RwLock<HashMap<String, InstanceInfo>>>;
+pub type InstanceMap = Arc<RwLock<InstanceConfig>>;
```
bfjelds, line 82: InstanceMap should probably be renamed, it is more than a map at this point.

agent/src/util/device_plugin_service.rs
johnsonshih: DevicePluginContext is a good name, rename InstanceConfig to DevicePluginContext

agent/src/util/device_plugin_service.rs
johnsonshih: rename.

bfjelds:
> The algorithm of CL resource allocation ensures allocating slots from different devices for a container on a given node. 

i'm not sure i grasp this, but hope that a container requesting 2 slots will be schedulable if there is 1 node discovering 1 device and capacity 2.  i can understand prioritizing high availability across nodes if possible, but i don't really agree with that being a requirement.

agent/src/util/device_plugin_service.rs
```diff
@@ -455,962 +534,2299 @@ impl InstanceDevicePlugin {
...
-        }
 
-        let current_device_usage = NodeUsage::from_str(current_device_usage_string.unwrap())
-            .map_err(|_| {
-                Status::new(
-                    Code::Unknown,
-                    format!(
-                        "Fails to parse {} to DeviceUsage ",
-                        current_device_usage_string.unwrap()
+            let virtual_devices_to_allocate = get_virtual_device_resources(
```
bfjelds, line 679: can 679-700 be combined:

```rust
            let resources_to_allocate = match get_virtual_device_resources(
                request.devices_i_ds.clone(),
                &dps.instance_map.read().await.clone(),
                &dps.node_name,
                &dps.config_namespace,
                &dps.config.capacity,
                kube_interface.clone(),
            )
            .await {
                Ok(resources) => {
                    info!(
                        "ConfigurationDevicePlugin::allocate - resource to allocate = {:?}",
                        resources
                    );
                    resources
                },
                Err(e) => {
                    dps.list_and_watch_message_sender
                        .send(ListAndWatchMessageKind::Continue)
                        .unwrap();
                    return Err(e);
                }
            };
```

bfjelds: (or something like it)

agent/src/util/device_plugin_service.rs
```diff
@@ -455,962 +534,2299 @@ impl InstanceDevicePlugin {
...
-                    "Requested device already in use",
-                ));
-            }
-            DeviceUsageStatus::Unknown => {
+
+                let node_usage = NodeUsage::create(
+                    &DeviceUsageKind::Configuration(vdev_id.clone()),
+                    &dps.node_name,
+                )
+                .unwrap();
```
bfjelds, line 757: i don't remember if there was some policy on when to return an error and when to unwrap, do we want to return an error here instead of unwrap call?

bfjelds: mostly, i just want to make sure that unwarp is called in appropriate and intentional places vs returning an error.

bfjelds:
> The purpose of report already reserved virtual devices + 1 free device usage (if available) per instance is to reduce the change that kubelet allocate slots more than available devices. For the example above, 2 devices are discovered and the capacity is 3, we can report 2 * 3 slots or 2 slots first and increase the available slots later.
>
> In the case of reporting 2 * 3 slots, if a container requests 4 CL resources, kubelet issue allocate request for the container and Agent will fail the request due only 2 devices are available. This end up in a infinite loop that kubelet keeps retry to allocate for 4 CL resources until another 2 devices discovered.
>
> If we report 2 slots and increase the available slots later, then the container request 4 CL resources won't be scheduled until the reported available slots exceeds 4. This will reduce the chance that kubelet entering the infinite loop allocating resources for the container.

i can't seem to wrap my brain around this.  i'm not sure i get why we would fail if a container requests 4 slots and there are 6.  i can understand why we'd prefer to succeed using slots from 4 different devices, but i don't understand why we'd fail to schedule if we had enough slots for the 2 devices.

agent/src/util/device_plugin_service.rs
johnsonshih: combine two statements

johnsonshih:
> > The purpose of report already reserved virtual devices + 1 free device usage (if available) per instance is to reduce the change that kubelet allocate slots more than available devices. For the example above, 2 devices are discovered and the capacity is 3, we can report 2 * 3 slots or 2 slots first and increase the available slots later.
> > In the case of reporting 2 * 3 slots, if a container requests 4 CL resources, kubelet issue allocate request for the container and Agent will fail the request due only 2 devices are available. This end up in a infinite loop that kubelet keeps retry to allocate for 4 CL resources until another 2 devices discovered.
> > If we report 2 slots and increase the available slots later, then the container request 4 CL resources won't be scheduled until the reported available slots exceeds 4. This will reduce the chance that kubelet entering the infinite loop allocating resources for the container.
> 
> i can't seem to wrap my brain around this. i'm not sure i get why we would fail if a container requests 4 slots and there are 6. i can understand why we'd prefer to succeed using slots from 4 different devices, but i don't understand why we'd fail to schedule if we had enough slots for the 2 devices.

There are actually only 2 devices even we reports 6 slots.  When allocating virtual devices to a container, does Agent treat each slot as a different virtual device? or Agent maps the slot back to actual device and allocate different devices?  If Agent treat each slot as a different virtual device, when a container request 4 virtual devices, it should allow it.  If Agent maps slots back to actual devices, then it should reject the allocation since only 2 devices are available.

We had a long discussion about what behavior should be when facing this situation.  I pasted the paragraph from the CL-resource proposal (https://github.com/project-akri/akri-docs/blob/main/proposals/configuration-level-resources.md) below.  The implementation is based to the concept of allocation by "unique device".  If we think allocation by "unique slot" is a valid scenario, we can add that support.  (I do have that implemented, we can add a field "uniqueDevice" in Configuration CRD to decide which allocation policy to use.)

There are two implementation options in the case where there are not enough unique Instances to meet the requested number of Configuration-level resources. One scenario where this could happen is if a Pod requests 3 cameras when only 2 exist with a capacity of 2 each. In this case, the Configuration-level camera resource shows as having a quantity of 4, despite there being two cameras. In this case, the kubelet will try to schedule the Pod, thinking there are enough resources. The Agent could either allocate 2 spots on one camera and one on the other or deny the allocation request. The latter is the preferred approach as it is more consistent and ensures the workload is getting the number of unique devices it expects. After failing an allocate request from the kubelet, the Pod will be in an UnexpectedAdmissionError state until another camera comes online and it can be successfully scheduled.

agent/src/util/device_plugin_service.rs
johnsonshih: dps.list_and_watch_message_sender is for notifying list_and_watch thread within the same device plugin service to re-scan the device availability.  Since it is always valid as long as the device plugin service is running, the unwrap() should always succeed, if not, then it's unexpected and we fail fast.

bfjelds:
> > > The purpose of report already reserved virtual devices + 1 free device usage (if available) per instance is to reduce the change that kubelet allocate slots more than available devices. For the example above, 2 devices are discovered and the capacity is 3, we can report 2 * 3 slots or 2 slots first and increase the available slots later.
> > > In the case of reporting 2 * 3 slots, if a container requests 4 CL resources, kubelet issue allocate request for the container and Agent will fail the request due only 2 devices are available. This end up in a infinite loop that kubelet keeps retry to allocate for 4 CL resources until another 2 devices discovered.
> > > If we report 2 slots and increase the available slots later, then the container request 4 CL resources won't be scheduled until the reported available slots exceeds 4. This will reduce the chance that kubelet entering the infinite loop allocating resources for the container.
> > 
> > 
> > i can't seem to wrap my brain around this. i'm not sure i get why we would fail if a container requests 4 slots and there are 6. i can understand why we'd prefer to succeed using slots from 4 different devices, but i don't understand why we'd fail to schedule if we had enough slots for the 2 devices.
> 
> There are actually only 2 devices even we reports 6 slots. When allocating virtual devices to a container, does Agent treat each slot as a different virtual device? or Agent maps the slot back to actual device and allocate different devices? If Agent treat each slot as a different virtual device, when a container request 4 virtual devices, it should allow it. If Agent maps slots back to actual devices, then it should reject the allocation since only 2 devices are available.
> 
> We had a long discussion about what behavior should be when facing this situation. I pasted the paragraph from the CL-resource proposal (https://github.com/project-akri/akri-docs/blob/main/proposals/configuration-level-resources.md) below. The implementation is based to the concept of allocation by "unique device". If we think allocation by "unique slot" is a valid scenario, we can add that support. (I do have that implemented, we can add a field "uniqueDevice" in Configuration CRD to decide which allocation policy to use.)
> 
> There are two implementation options in the case where there are not enough unique Instances to meet the requested number of Configuration-level resources. One scenario where this could happen is if a Pod requests 3 cameras when only 2 exist with a capacity of 2 each. In this case, the Configuration-level camera resource shows as having a quantity of 4, despite there being two cameras. In this case, the kubelet will try to schedule the Pod, thinking there are enough resources. The Agent could either allocate 2 spots on one camera and one on the other or deny the allocation request. The latter is the preferred approach as it is more consistent and ensures the workload is getting the number of unique devices it expects. After failing an allocate request from the kubelet, the Pod will be in an UnexpectedAdmissionError state until another camera comes online and it can be successfully scheduled.

i guess i read `preference` in this line in the proposal: `the Agent will preference unique Instances` ... it is fine to preference unique instances, but if you don't have that, available access (capacity) should be acceptable.

bfjelds:
> i guess i read `preference` in this line in the proposal: `the Agent will preference unique Instances` ... it is fine to preference unique instances, but if you don't have that, available access (capacity) should be acceptable.

@kate-goldenring, I imagine you've done more thinking on this than I have.  My gut is that preferring unique instances is fine for CL, but that there should be a way to schedule to available slots (regardless of instance uniqueness).  What is your take?

diconico07:
> > i guess i read `preference` in this line in the proposal: `the Agent will preference unique Instances` ... it is fine to preference unique instances, but if you don't have that, available access (capacity) should be acceptable.
> 
> @kate-goldenring, I am way late to this party and I imagine you've done far more in-depth thinking on this than I have. My gut is that preferring unique instances is fine for CL, but that there should be a way to schedule to available slots (regardless of instance uniqueness). What is your take?

@bjfield for me the main use case is "I want to deploy a workload with 2 (distinct) cameras", if I have a single camera available on a node then the workload is not scheduled on that node, but if you expose all slots, if a single camera has two slots then it will end up scheduled on the node even though that's not what you wanted.

The current implementation has a lot of unsolved corner cases here, as @johnsonshih told during community call there is a "race window" when a workload gets deleted where two device plugin slots will still be "available" even though we only want one, or a case where we have two containers on the same pod each asking for the 2 cameras that wouldn't be scheduled. But I don't really see any way to solve these with the Device Plugin API (maybe we could solve these with the new [Dynamic Resource Allocation API](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/), but I don't think we want to push it right now) without sacrificing the "If I want 2 distinct cameras, I don't want to end-up with two slots of the same camera" thing.

bfjelds:
> @bjfield for me the main use case is "I want to deploy a workload with 2 (distinct) cameras", if I have a single camera available on a node then the workload is not scheduled on that node, but if you expose all slots, if a single camera has two slots then it will end up scheduled on the node even though that's not what you wanted.
> 
> The current implementation has a lot of unsolved corner cases here, as @johnsonshih told during community call there is a "race window" when a workload gets deleted where two device plugin slots will still be "available" even though we only want one, or a case where we have two containers on the same pod each asking for the 2 cameras that wouldn't be scheduled. But I don't really see any way to solve these with the Device Plugin API (maybe we could solve these with the new [Dynamic Resource Allocation API](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/), but I don't think we want to push it right now) without sacrificing the "If I want 2 distinct cameras, I don't want to end-up with two slots of the same camera" thing.

i think i've always understood this scenario to be: i want to schedule against slots without having to understand instance hashes.  nothing about distinct or non-distinct entered into my mind.

bfjelds:
> i think i've always understood this scenario to be: i want to schedule against slots without having to understand instance hashes. nothing about distinct or non-distinct entered into my mind.

but i'll defer to the other folks.  the code seems good to me, approved.

bfjelds (approved):

johnsonshih:
> > i think i've always understood this scenario to be: i want to schedule against slots without having to understand instance hashes. nothing about distinct or non-distinct entered into my mind.
> 
> but i'll defer to the other folks. the code seems good to me, approved.

thanks for reviewing this PR.

PR #628, opened 2023-07-11
add DeviceUsage and DeviceUsageKind for Instance.device_usage

johnsonshih:
This PR is part of the works for Configuration level resource support.  In order to keep the device usage state for a usage slot, we introduce new data types DeviceUsageKind and DeviceUsage to keep track of the owner of usage slot.  
The design of DeviceUsageKind and DeviceUsage is described in the doc PR: https://github.com/project-akri/akri-docs/pull/76
device usage design at:
https://github.com/johnsonshih/akri-docs/blob/user/jshih/cl-implementation/proposals/configuration-level-resources.md#maintaining-device-usage

The complete PR for CL resource support is in https://github.com/project-akri/akri/pull/627

**If applicable**:


agent/src/util/device_plugin_service.rs
```diff
@@ -335,19 +382,39 @@ impl InstanceDevicePlugin {
...
             let mut akri_device_properties = HashMap::new();
+            let mut akri_devices = HashMap::<String, Device>::new();
             for device_usage_id in request.devices_i_ds {
                 trace!(
                     "InstanceDevicePlugin::allocate - for Instance {} processing request for device usage slot id {}",
                     &dps.instance_name,
                     device_usage_id
                 );
 
+                if let Err(e) = try_update_instance_device_usage(
```
bfjelds, line 393: i can't imagine how changing where try_update_instance_device_usage is called relative to the other code in allocate would matter ... but am curious why the move?

agent/src/util/device_plugin_service.rs
```diff
@@ -403,48 +455,62 @@ impl InstanceDevicePlugin {
...
+///     * (ACTION) return previously reserved kind, DeviceUsageStatus::ReservedByConfiguration or DeviceUsageStatus::ReservedByInstance
+/// 3. device_usage.usage == (some other node) ... this means that we believe this device is in use by another node
+///     * (ACTION) return DeviceUsageStatus::ReservedByOtherNode
+fn get_device_usage_state(device_usage: &DeviceUsage, node_name: &str) -> DeviceUsageStatus {
+    let device_usage_state = match device_usage.get_kind() {
+        DeviceUsageKind::Free => DeviceUsageStatus::Free,
+        DeviceUsageKind::Configuration(vdev_id) => {
+            DeviceUsageStatus::ReservedByConfiguration(vdev_id)
+        }
+        DeviceUsageKind::Instance => DeviceUsageStatus::ReservedByInstance,
```
bfjelds, line 506: what happens if Unknown is found?

agent/src/util/device_plugin_service.rs
```diff
@@ -453,6 +519,7 @@ async fn try_update_instance_device_usage(
     node_name: &str,
     instance_name: &str,
     instance_namespace: &str,
+    device_usage_kind: DeviceUsageKind,
```
bfjelds, line 522: if a param of type DeviceUsage is passed in, would we need both device_usage_d and device_usage_kind?  or, can device_usage_kind be parsed from device_usage_id? do we need 2 args for device_usage info?

bfjelds: nevermind ... i see.

device_usage_id is used to get current slot usage ... device_usage_kind is the request/desired type.

agent/src/util/device_plugin_service.rs
```diff
@@ -403,48 +455,62 @@ impl InstanceDevicePlugin {
...
+        }
+        DeviceUsageKind::Instance => DeviceUsageStatus::ReservedByInstance,
+    };
+    if device_usage_state != DeviceUsageStatus::Free && !device_usage.is_same_usage(node_name) {
+        return DeviceUsageStatus::ReservedByOtherNode;
+    }
+    device_usage_state
+}
+
 /// This tries up to `MAX_INSTANCE_UPDATE_TRIES` to update the requested slot of the Instance with the this node's name.
```
bfjelds, line 514: update comment to more accurately describe what this function is doing.  for example, describe when errors are returned (if slot is owned by another node, slot is owned by another kind, slot is unknown) vs when/how the instance is updated.

agent/src/util/device_plugin_service.rs
bfjelds: maybe: change device_usage_kind to desired_device_usage_kind

agent/src/util/device_plugin_service.rs
```diff
@@ -476,27 +543,93 @@ async fn try_update_instance_device_usage(
         }
 
         // Update the instance to reserve this slot for this node iff it is available and not already reserved for this node.
-        if slot_available_to_reserve(device_usage_id, node_name, &instance)? {
-            instance
-                .device_usage
-                .insert(device_usage_id.to_string(), node_name.to_string());
+        let device_usage_string = instance.device_usage.get(device_usage_id);
```
bfjelds, line 546: maybe change device_usage_string into existing_device_usage_string or current_device_usage_string

agent/src/util/device_plugin_service.rs
```diff
@@ -476,27 +543,93 @@ async fn try_update_instance_device_usage(
...
-                random_delay().await;
-            } else {
-                return Ok(());
             }
-        } else {
-            // Instance slot already reserved for this node
-            return Ok(());
-        }
+            DeviceUsageStatus::ReservedByConfiguration(_) => {
+                if matches!(device_usage_kind, DeviceUsageKind::Configuration(_)) {
```
bfjelds, line 595: maybe add comment ... something like : safe to assume that this node owns the slot because get_device_usage_state only returns ReservedByConfiguration when this node owns the slot, for slots owned by other nodes (by Configuration or Instance), ReservedByOtherNode is returned.

agent/src/util/device_plugin_service.rs
```diff
@@ -476,27 +543,93 @@ async fn try_update_instance_device_usage(
...
+                    return Ok(());
+                } else {
+                    return Err(Status::new(
+                        Code::Unknown,
+                        "Requested device already in use",
+                    ));
+                }
+            }
+            DeviceUsageStatus::ReservedByInstance => {
+                if matches!(device_usage_kind, DeviceUsageKind::Instance) {
```
bfjelds, line 605: maybe add comment ... something like : safe to assume that this node owns the slot because get_device_usage_state only returns ReservedByInstance when this node owns the slot, for slots owned by other nodes (by Configuration or Instance), ReservedByOtherNode is returned.

agent/src/util/device_plugin_service.rs
```diff
@@ -476,27 +543,93 @@ async fn try_update_instance_device_usage(
...
+                return Err(Status::new(
+                    Code::Unknown,
+                    "Requested device already in use",
+                ));
+            }
+            DeviceUsageStatus::Unknown => {
+                trace!(
+                    "try_update_instance_device_usage - request for device slot {} status unknown!",
+                    device_usage_id
+                );
```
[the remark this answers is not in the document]
bfjelds: It might make catching bugs easier if Unknown were always an error case that essentially panic'd.

agent/src/util/device_plugin_service.rs
```diff
@@ -476,27 +543,93 @@ async fn try_update_instance_device_usage(
...
+        let device_usage = DeviceUsage::from_str(device_usage_string.unwrap()).map_err(|_| {
+            Status::new(
+                Code::Unknown,
+                format!(
+                    "Fails to parse {} to DeviceUsage ",
+                    device_usage_string.unwrap()
+                ),
+            )
+        })?;
+        match get_device_usage_state(&device_usage, node_name) {
```
bfjelds, line 568: if this match were in its own function it might make creating tests easier ... ignore if this match already has good test coverage :)

agent/src/util/device_plugin_service.rs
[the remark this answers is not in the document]
johnsonshih: it's a valid case that when an kube instance (akrii) is removed and the instance map in local Agent hasn't being updated.

agent/src/util/device_plugin_service.rs
johnsonshih: nothing special, since we return error when try_update_instance_device_usage fails, call it earlier then we don't need to waste time to add annotations, and other bookkeeping works.

agent/src/util/device_plugin_service.rs
```diff
@@ -403,48 +455,62 @@ impl InstanceDevicePlugin {
...
-        // No corresponding id found
-        trace!(
-            "slot_available_to_reserve - could not find {} id in device_usage",
-            device_usage_id
-        );
-        Err(Status::new(
-            Code::Unknown,
-            "Could not find device usage slot",
-        ))
+        Err(_) => (0..*capacity)
```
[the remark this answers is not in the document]
johnsonshih: this function doesn't return error. The caller has nothing to do if receiving error but marks the device unhealthy, so use UNKNOWN.  The definition of DeviceUsageStatus::Unknown

    /// Unknown, insufficient information to determine the status,
    /// mostly due to the device usage slot is not found from the instance map
    Unknown,

agent/src/util/device_plugin_service.rs
johnsonshih: This function return DeviceUsageStatus based on DeviceUsageKind and node name, it has sufficient information to determine the DeviceUsageStatus and never returns DeviceUsageStatus::Unkown

agent/src/util/device_plugin_service.rs
johnsonshih: The doc comments describe what the function is doing.  I think it covers what you asked.

/// This returns device usage status of a `device_usage_id` slot for an instance on a given node
/// # More details
/// Cases based on the device usage value
/// 1. DeviceUsageKind::Free ... this means that the device is available for use
///     * (ACTION) return DeviceUsageStatus::Free
/// 2. device_usage.usage == node_name ... this means node_name previously used device_usage
///     * (ACTION) return previously reserved kind, DeviceUsageStatus::ReservedByConfiguration or DeviceUsageStatus::ReservedByInstance
/// 3. device_usage.usage == (some other node) ... this means that we believe this device is in use by another node
///     * (ACTION) return DeviceUsageStatus::ReservedByOtherNode

agent/src/util/device_plugin_service.rs
johnsonshih: rename device_usage_kind to desired_device_usage_kind

agent/src/util/device_plugin_service.rs
johnsonshih: rename device_usage_string to current_device_usage_string

agent/src/util/device_plugin_service.rs
```diff
@@ -476,27 +543,93 @@ async fn try_update_instance_device_usage(
...
+        }
 
-            if let Err(e) = kube_interface
-                .update_instance(&instance, instance_name, instance_namespace)
-                .await
-            {
-                if x == (MAX_INSTANCE_UPDATE_TRIES - 1) {
-                    trace!("try_update_instance_device_usage - update_instance returned error [{}] after max tries ... returning error", e);
-                    return Err(Status::new(Code::Unknown, "Could not update Instance"));
+        let device_usage = DeviceUsage::from_str(device_usage_string.unwrap()).map_err(|_| {
```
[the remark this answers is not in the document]
johnsonshih: rename device_usage to current_device_usage

agent/src/util/device_plugin_service.rs
johnsonshih: yes, we have good test coverage for both try_update_instance_device_usage and get_device_usage_state

agent/src/util/device_plugin_service.rs
johnsonshih: add comments

agent/src/util/device_plugin_service.rs
johnsonshih: add comments

agent/src/util/device_plugin_service.rs
bfjelds: sorry, meant comment for try_update_instance_device_usage

agent/src/util/device_plugin_service.rs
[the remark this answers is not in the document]
bfjelds: interesting.  maybe the type should be called MissingInstance?

agent/src/util/device_plugin_service.rs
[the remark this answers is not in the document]
bfjelds: so, in this case, it seems like the error happens because of a missing instance.  this code exists to tell the device plugin to report all slots as unhealthy?  if so, maybe add a comment to that effect here.

agent/src/util/device_plugin_service.rs
bfjelds: sorry, my bad!  there is no DeviceUsageKind::Unknown ... only a DeviceUsageStatus::Unknown.

agent/src/util/device_plugin_service.rs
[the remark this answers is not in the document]
johnsonshih: it's not an error, when a device is gone, the kube instance can be deleted by one of the nodes that has discovered the device. Once the kube instance is deleted, dp on other nodes will see the instance missing and report the device slots unhealthy.  I'll add comment for this.

agent/src/util/device_plugin_service.rs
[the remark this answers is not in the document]
johnsonshih: missing instance is one of the case, another case is when the device plugin has problem parsing the device usage string.  Currently, we never return error for parsing device usage string, but in case the code changes to return parsing error, device usage state will be Unknown since insufficient information to tell the status.

agent/src/util/device_plugin_service.rs
johnsonshih: add comments for try_update_instance_device_usage

agent/src/util/device_plugin_service.rs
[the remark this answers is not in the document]
bfjelds: > another case is when the device plugin has problem parsing the device usage string
this one i'd favor panic over resliency so we can understand why this happens and eliminate it.

shared/src/akri/instance.rs
```diff
@@ -323,6 +323,108 @@ fn default_shared() -> bool {
...
+
+        pub fn get_kind(&self) -> DeviceUsageKind {
+            self.kind.clone()
+        }
+
+        pub fn get_usage_name(&self) -> String {
+            self.usage_name.clone()
+        }
+
+        pub fn is_same_usage(&self, usage_name: &str) -> bool {
```
bfjelds, line 422: should `is_same_usage` depend on the usage kind in addition to the usage name?

maybe this should be renamed to used_by_same_node?

shared/src/akri/instance.rs
```diff
@@ -323,6 +323,108 @@ fn default_shared() -> bool {
...
+                kind: kind.clone(),
+                usage_name: usage_name.into(),
+            })
+        }
+
+        pub fn get_kind(&self) -> DeviceUsageKind {
+            self.kind.clone()
+        }
+
+        pub fn get_usage_name(&self) -> String {
```
bfjelds, line 418: usage_name is actually the node name, right? should it be usage_node to be clearer?

agent/src/util/slot_reconciliation.rs
```diff
@@ -163,16 +170,31 @@ impl DevicePluginSlotReconciler {
                 .device_usage
                 .iter()
                 .filter_map(|(k, v)| {
-                    if v != node_name && node_slot_usage.contains(k) {
-                        // We need to add node_name to this slot IF
-                        //     the slot is not labeled with node_name AND
-                        //     there is a container using that slot on this node
-                        Some(k.to_string())
-                    } else {
+                    let same_node_name = match DeviceUsage::from_str(v) {
```
bfjelds, line 173: would this be more dogmatic rust if it were:

```rust
                    match DeviceUsage::from_str(v) {
                        Ok(usage) => if usage.is_same_usage(node_name) {
                            None
                        } else {
                            // We need to add node_name to this slot IF
                            //     the slot is not labeled with node_name AND
                            //     there is a container using that slot on this node
                            node_slot_usage
                                .iter()
                                .find(|u| match DeviceUsage::from_str(u) {
                                    Ok(slot_usage) => slot_usage.is_same_usage(k),
                                    Err(_) => false,
                                })
                                .map(|usage| {
                                    (
                                        k.to_string(),
                                        DeviceUsage::from_str(usage).unwrap().get_kind(),
                                    )
                                })
                        },
                        Err(_) => None,
                    };
```

agent/src/util/device_plugin_service.rs
```diff
@@ -403,56 +455,73 @@ impl InstanceDevicePlugin {
...
-///     * (ACTION) return true
-/// 2. device_usage\[id\] == self.nodeName ... this means THIS node previously used id, but the DevicePluginManager knows that this is no longer true
-///     * (ACTION) return false
-/// 3. device_usage\[id\] == (some other node) ... this means that we believe this device is in use by another node and should be marked unhealthy
-///     * (ACTION) return error
-/// 4. No corresponding id found ... this is an unknown error condition (BAD)
-///     * (ACTION) return error
-fn slot_available_to_reserve(
-    device_usage_id: &str,
+pub async fn get_instance_device_usage_states(
```
bfjelds, line 458: add function comment about get_instance_device_usage_states ... something that includes that any !Ok (including expected things like missing instance and unexpected like parsing issues) encountered will result in Unknown status.

agent/src/util/device_plugin_service.rs
```diff
@@ -403,56 +455,73 @@ impl InstanceDevicePlugin {
...
+    let mut device_usage_states = Vec::new();
+    match kube_interface
+        .find_instance(instance_name, instance_namespace)
+        .await
+    {
+        Ok(kube_akri_instance) => {
+            for (device_name, device_usage_string) in kube_akri_instance.spec.device_usage {
+                let device_usage_status = match DeviceUsage::from_str(&device_usage_string) {
+                    Ok(device_usage) => get_device_usage_state(&device_usage, node_name),
+                    Err(_) => DeviceUsageStatus::Unknown,
```
bfjelds, line 474: add logging for Err(_) when get_device_usage_state fails ... would be good to know when this happens and why it happens.

shared/src/akri/instance.rs
johnsonshih: no, is_same_usage compare the usage only.  the usage can be any attribute, not limited to node_name.

usage can be any attribute that's referred to the usage of the slot. currently, it can be
node_name: when refers to the slot usage in akrii
slot_id: when refers to the slot usage in annotation

agent/src/util/device_plugin_service.rs
johnsonshih: add comments

agent/src/util/device_plugin_service.rs
johnsonshih: add error log

agent/src/util/slot_reconciliation.rs
johnsonshih: we want to treat it as different usage if the first parsing DeviceUsage::from_str(v) returns error.  The proposed code treats it as the same usage.
anyway, I'll refactor the code.

shared/src/akri/instance.rs
bfjelds: oh.  that is not an intuitive overload of behavior to me.  i'll have to take another look.

bfjelds: i see the annotation code:

```rust
let slot_usage =
    DeviceUsage::create(&DeviceUsageKind::Instance, &device_usage_id).unwrap();
akri_annotations.insert(
    format!("{}{}", AKRI_SLOT_ANNOTATION_NAME_PREFIX, &device_usage_id),
    slot_usage.to_string(),
);
```

given that the device_usage_id is part of the annotation name, why put it into the DeviceUsage object as well?  why not put the node_name in the DeviceUsage object and be consistent with the slot usage?

bfjelds: given the annotation name and value are using the same device_usage_id, is there a time where the annotation name and value won't be alligned?

where this is false: `DeviceUsage::from_str(value).is_same_usage(slot_id)` in crictl_containers.rs:

```rust
if key.starts_with(AKRI_SLOT_ANNOTATION_NAME_PREFIX) {
    let slot_id = key
        .strip_prefix(AKRI_SLOT_ANNOTATION_NAME_PREFIX)
        .unwrap_or_default();
    match DeviceUsage::from_str(value) {
        Ok(slot_usage) => {
            if slot_usage.is_same_usage(slot_id) {
                Some(value.clone())
            } else {
                None
            }
        }
        Err(_) => None,
    }
} else {
    None
}
```

bfjelds: nevermind ... seems to match the original code where we were checking annotation name against the value.

bfjelds:
```rust
let slot_usage =
    DeviceUsage::create(&DeviceUsageKind::Instance, &device_usage_id).unwrap();
akri_annotations.insert(
    format!("{}{}", AKRI_SLOT_ANNOTATION_NAME_PREFIX, &device_usage_id),
    slot_usage.to_string(),
);
```

i'm probably missing the reason, but it looks like slot_reconciliation is just using the name in slot_usage ... the original code just passed device_usage_id, why change it?

agent/src/util/slot_reconciliation.rs
bfjelds: yup, my code is definitely wrong ... sorry!

i think my code suggestion should've been something closer to this:

```rust
if match DeviceUsage::from_str(v) {
    Ok(usage) => usage.is_same_usage(node_name),
    Err(_) => false,
} {
    None
} else {
    // We need to add node_name to this slot IF
    //     the slot is not labeled with node_name AND
    //     there is a container using that slot on this node
    node_slot_usage
        .iter()
        .find(|u| match DeviceUsage::from_str(u) {
            Ok(slot_usage) => slot_usage.is_same_usage(k),
            Err(_) => false,
        })
        .map(|usage| {
            (
                k.to_string(),
                DeviceUsage::from_str(usage).unwrap().get_kind(),
            )
        })
}
```

johnsonshih:
> 
This is for supporting Configuration level resources.

Since the Akri Instance is the source of true for Agent, we need to keep the device plugin type in the device usage, for the same reason, we need the same information for reconciler.  Originally the annotation name looks like 

"akri.agent.slot-my-resource-00095f-3": "my-resource-00095f-3"

with the DeviceUsage, the annotation can hold the device plugin type information.

"akri.agent.slot-my-resource-00095f-2": "C:0:my-resource-00095f-2"

Please check the details from the CL resource proposal
https://github.com/project-akri/akri-docs/blob/main/proposals/configuration-level-resources.md#maintaining-device-usage

bfjelds:
Sorry @johnsonshih, I meant to quote reply to your last comment, but seem to have editted it instead.  Adding your original comment as best as I can:

This is for supporting Configuration level resources.

Since the Akri Instance is the source of true for Agent, we need to keep the device plugin type in the device usage, for the same reason, we need the same information for reconciler. Originally the annotation name looks like

    "akri.agent.slot-my-resource-00095f-3": "my-resource-00095f-3"

with the DeviceUsage, the annotation can hold the device plugin type information.

    "akri.agent.slot-my-resource-00095f-2": "C:0:my-resource-00095f-2"

Please check the details from the CL resource proposal
[https://github.com/project-akri/akri-docs/blob/main/proposals/configuration-level-resources.md#maintaining-device-usage](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fproject-akri%2Fakri-docs%2Fblob%2Fmain%2Fproposals%2Fconfiguration-level-resources.md%23maintaining-device-usage&data=05%7C01%7CBrian.Fjeldstad%40microsoft.com%7C9afaf3aa77e948ab8deb08db897da265%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638254944922230391%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=D%2BQFYrGoaW3S3WU7RqtNZTcCagRszLEsXsFZkxv%2F58Y%3D&reserved=0)

johnsonshih:
> Sorry @johnsonshih, I meant to quote reply to your last comment, but seem to have editted it instead. Restored your original comment as best as I can and replying here:
> 
> > This is for supporting Configuration level resources.
> > ...
> > Please check the details from the CL resource proposal
> > https://github.com/project-akri/akri-docs/blob/main/proposals/configuration-level-resources.md#maintaining-device-usage
> 
> sorry, the doc definitely does a great job explaining what to expect, but didn't help my brain understand why kind is required in the reconciler. unfortunately, it has been a long time and i didn't remember the block of code in reconciler that adds missing usage to Instance.deviceUsage (which would need to reconstruct the DeviceUsage object).
> 
> i still feel like overloading DeviceUsage to accept node name or slot name is confusing. i would rather the annotation be changed to be something like `"<slotname>=<deviceusage.tostring>"`, then the reconciler could just split it apart once into `map<slotname, deviceusage>` and use deviceusage to repopulate the missing Instance.deviceUsage. if that were the case, then DeviceUsage's name would be unambiguous.

I'll change the code to avoid overloading DeviceUsage.

Here is the summary for what reconciler does and why it needs usage_kind:
the deviceUsage in Akri Instance keep the information slot_id: (usage_kind: node_name).  The Akri Instances can be written by multiple nodes in parallel, there is a race condition that can cause the deviceUsage information in Arki Instance be accidentally wiped or have incorrect data written to it.  The reconciler is responsible to ensure the deviceUsage status for the node that reconciler runs on are correct.

Reconciler needs slot_id, usage_kind and node_name to do the check. It gets that information by using the name of node it is running on and the annotation (usage_kind: slot_id) from pods running on the node.  Without usage_kind, reconciler cannot correct the deviceUsage to correct state.

bfjelds:
> 
> I'll change the code to avoid overloading DeviceUsage.
> 
> Here is the summary for what reconciler does and why it needs usage_kind: the deviceUsage in Akri Instance keep the information slot_id: (usage_kind: node_name). The Akri Instances can be written by multiple nodes in parallel, there is a race condition that can cause the deviceUsage information in Arki Instance be accidentally wiped or have incorrect data written to it. The reconciler is responsible to ensure the deviceUsage status for the node that reconciler runs on are correct.
> 
> Reconciler needs slot_id, usage_kind and node_name to do the check. It gets that information by using the name of node it is running on and the annotation (usage_kind: slot_id) from pods running on the node. Without usage_kind, reconciler cannot correct the deviceUsage to correct state.

one last suggestion, if SlotUsage is `write!(f, "{}:{}", <slot-id>, <nodeusage.tostring>)`, then you won't need to parse the nodeusage kind and recreate a NodeUsage object to fix the race condition (just use the string found in the annotation).

bfjelds (approved):

johnsonshih:
> > I'll change the code to avoid overloading DeviceUsage.
> > Here is the summary for what reconciler does and why it needs usage_kind: the deviceUsage in Akri Instance keep the information slot_id: (usage_kind: node_name). The Akri Instances can be written by multiple nodes in parallel, there is a race condition that can cause the deviceUsage information in Arki Instance be accidentally wiped or have incorrect data written to it. The reconciler is responsible to ensure the deviceUsage status for the node that reconciler runs on are correct.
> > Reconciler needs slot_id, usage_kind and node_name to do the check. It gets that information by using the name of node it is running on and the annotation (usage_kind: slot_id) from pods running on the node. Without usage_kind, reconciler cannot correct the deviceUsage to correct state.
> 
> one last suggestion, if SlotUsage is `write!(f, "{}:{}", <slot-id>, <nodeusage.tostring>)`, then you won't need to parse the nodeusage kind and recreate a NodeUsage object to fix the race condition (just use the string found in the annotation).
> 
> i'm approving regardless of whether you take the suggestion, just thought it might be easier to maintain.

I thought about that before but decided to keep use the slot_id to avoid causing break change.  After a second thought, I think it's worth to pay the price now to get a clean design.  I'll update the PR to use NodeUsage in the annotation value to reduce the parsing

bfjelds (approved):

Issue #250, opened 2021-02-22
Support ONVIF cameras that require authentication


kate-goldenring:
**Is your feature request related to a problem? Please describe.**
Our sample ONVIF broker currently does not support connecting to cameras that require authentication . This support should be added along with documentation on how to pass credentials.

**Describe the solution you'd like**
A solution should be proposed and made for passing unique credentials such as user/pass to brokers. One solution could be to use a ConfigMap. The ONVIF sample broker should be updated to optionally use authentication with the provided credentials. 

**Describe alternatives you've considered**
Credentials could be passed via kubernetes secrets mounted into broker Pods, but then all brokers would get the same credentials. All discovered IP cameras would need to accept the same user/pass which is unlikely and unsage.

**Additional context**
Credential passing is currently supported for OPC UA -- aided by the use of x.509v3 certificates which can have a common CA over standard user/pass; however, that is not a complete solution either. 
For [OPC UA](https://github.com/deislabs/akri/blob/main/docs/opcua-configuration.md), kubernetes Secrets containing an x.509 certificate for an OPC UA client and the associated certificate authority are [set in the broker PodSpec](https://github.com/deislabs/akri/blob/main/docs/proposals/credentials-passing.md#how-secrets-are-used-in-opc-ua) of a Configuration. However, this assumes that all discovered OPC UA servers have the same Certificate Authority as the certificate passed to the broker or have manually accepted the broker's client. It also means that all brokers have the same certificate, which is unideal. A configMap could also be leveraged by OPC UA to give brokers unique certificates. Or the OPC UA brokers could call out to a CA to get a signed certificate.

kate-goldenring:
[...]

PR #630, opened 2023-07-14
retrieve uuid from endpoint address, use it with service url as device id

johnsonshih:
refs #250 
This PR retrieve the Onvif device uuid and use it with service url to form the device id reported to Agent.  

According to [ONVIF core spec](https://www.onvif.org/specs/core/ONVIF-Core-Specification.pdf) sec 7.3.1. The combination of an wsadis:Address
and wsadis:ReferenceProperties provide a stable and globally-unique identifier.

We can use "wsadis:Address + wsadis:ReferenceProperties.Address" as the identifier of Onvif devices. The DH needs to expose the uuid to brokers so brokers have the same information to uniquely identify a device.

7.3.1 Endpoint reference
A device or an endpoint that takes the client role should use a URN:UUID [RFC4122] as the address property
of its endpoint reference.
The device or an endpoint that takes the client role shall use a stable, globally unique identifier that is constant
across network interfaces as part of its endpoint reference property. The combination of an wsadis:Address
and wsadis:ReferenceProperties provide a stable and globally-unique identifier.

**If applicable**:


bfjelds (approved):

diconico07 (approved):
Will need a note in the release notes about the fact it changes the id of ONVIF devices I think.

johnsonshih:
> Will need a note in the release notes about the fact it changes the id of ONVIF devices I think.

sure, there will be follow up change for onvif DH w.r.t. the device id, I'll update the doc in one shot

PR #631, opened 2023-07-17
Use ENTRYPOINT instead of CMD in Dockerfile

### LINKED ISSUE #239: Container best practice is to use ENTRYPOINT rather than CMD
Our containers use CMD rather than ENTRYPOINT, but it seems that best practice is to use ENTRYPOINT:

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint

We should change our dockerfiles according to best practices.

gauravgahlot:

Closes #239 

**If applicable**:


diconico07 (approved):

bfjelds:
change looks good, it would be nice to have the tests run on it though.

diconico07:
> change looks good, it would be nice to have the tests run on it though.

About tests runs, I noticed the GitHub action to bump version doesn't re-trigger all tests, but you can still look at the test results of previous commit (with obviously the version test that fails) for that, I didn't find the root cause of tests not running, so that's my workaround

Issue #635, opened 2023-07-22
Make IP/MAC address exported by Onvif Optional


johnsonshih:
**Is your feature request related to a problem? Please describe.**
Retrieving IP and MAC address of Onvif requires authentication.  Currently, Onvif DH always queries IP/MAC address hence it won't work for authentication-enabled Onvif devices (most of the Onvif devices are authentication-enabled in product environment).  We can change the Onvif to skip query IP/MAC address if the IP and Mac filters are not used.  That can enable Onvif DH to discover authentication-enabled devices.
If we query IP/MAC address optionally, the consequence is that Onvif DH won't expose environment variables for IP and MAC address if it cannot query the IP/MAC address from Onvif devices correctly.  We need to add this to Onvif DH document

**Is your feature request related to a way you would like Akri extended? Please describe.**
A clear and concise description of what cool feature Akri lacks. Ex. I'd love Akri to support discovery of Bluetooth devices.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

johnsonshih:
PR committed

PR #640, opened 2023-08-01
Return discovered devices that fails to get ip/mac address if ip/mac filters are not used

johnsonshih:
…address filters are not used

ref #250 
ref #635 

Currently, Onvif DH will not return discovered devices if it fails to get devices' ip/mac address.  This is not necessary if ip/mac address filter is not used.  This PR change Onvif DH so when ip/mac address filter is not used, Onvif DH returns discovered devices no matter it gets ip/mac address from that device (other filters still applied).

When returning discovered devices, Onvif DH does best efforts to retrieve ip/mac address and set to environment variables.  If ip/mac address filter is not used, it is possible that ip/mac address environments won't be set (if Onvif DH cannot retrieve ip/mac address from devices).

**If applicable**:


diconico07 (changes requested):
Except the refactor I mentioned and the need for some unit test, this is good.

diconico07 (approved):

PR #654, opened 2023-09-01
fix(agent): correct agent's behavior on Instance deletion

### LINKED ISSUE #650: A shared device Instance should get deleted only when all discovery handlers lose track of it
**Describe the bug**
As noticed by @kate-goldenring in project-akri/akri-docs#77,  we currently delete `Instances` for shared devices when any of the
discovery handler's instance doesn't see the device anymore.

**Expected behavior**
When agent "deletes" a shared device Instance, it should remove itself from the `nodes` field, and delete the resource only if it makes that list empty.

We must ensure the controller reschedules broker pods to nodes that still have the resource available when this happens.

diconico07:
Make the agent remove itself from the instance nodes list instead of deleting the instance if it's not the last node with access to the device.

Refactored the way we get the AGENT_NODE_NAME to only fetch it once at startup.

This fixes #650.

**If applicable**:


agent/src/util/discovery_operator.rs
```diff
@@ -1869,6 +2009,29 @@ pub mod tests {
         // 4: Assert that local devices that go offline are removed from the instance map
         //
         let mut mock = MockKubeInterface::new();
+        let instance: akri_shared::akri::instance::Instance = serde_json::from_str(
```
kate-goldenring, line 2012: nit: might be worth moving this test instance into a `get_test_instance()` function to reduce lines of code, maybe supporting a configurable nodes list

kate-goldenring (approved):
LGTM! just one nit of code reduction in tests but feel free to ignore.

PR #672, opened 2023-10-13
update anomaly detection app dependencies

johnsonshih:
Add missing dependency to anomaly detection app, the python3-sklearn was removed accidentally in previous PR, the dependency is required or the application cannot run properly.

Also remove the version stamp for numpy and let the system picks proper version to use.  Use numpy version 1.21.4 causes issues when building arm32v7 images.  The build stuck at installing dependencies and never complete.

**If applicable**:


build/containers/Dockerfile.anomaly-detection-app
```diff
@@ -9,7 +9,7 @@ COPY ./samples/apps/anomaly-detection-app .
 
 RUN echo "Creating container based on ${PLATFORM}/debian:bullseye-slim" && \
     apt-get update && \
-    apt-get install -y protobuf-compiler libprotoc-dev python3-pip && \
+    apt-get install -y protobuf-compiler libprotoc-dev python3-pip python3-sklearn && \
```
diconico07, line 12: This makes the app to use system wide `scikit_learn` as well as system wide `numpy`, for bullseye none of these two will be the version we had in requirements.txt

samples/apps/anomaly-detection-app/requirements.txt
```diff
@@ -1,6 +1,5 @@
 grpcio==1.53.0
 grpcio-tools==1.53.0
 Flask==2.2.5
-numpy==1.21.4
-protobuf==4.21.6
-scikit_learn==1.0.1
```
diconico07, line 6: scikit_learn is still a dependency for the sample app

diconico07 (commented):
If I understand correctly, this is needed to make armv7 version to work, if so, as it's just a sample app I guess we can disable armv7 build for this one.

johnsonshih:
> If I understand correctly, this is needed to make armv7 version to work, if so, as it's just a sample app I guess we can disable armv7 build for this one.

yes, but I'd like to fix the build first, we can disable armv7 build in a separate PR so in case we bring armv7 back in the future, it still works.

build/containers/Dockerfile.anomaly-detection-app
johnsonshih: that's correct, arm32v7 build stuck forever if use the version specified in requirements for numpy and scikit_learn.

samples/apps/anomaly-detection-app/requirements.txt
johnsonshih: The scikit_learn is pulled in by the Dockerfile

johnsonshih: add scikit_learn back but use default version.

johnsonshih:
> Fine for me, I still think the best way for a fix would be to just disable armv7 here. As the only way to get back an armv7 build (if we switch to `python` base image rather than debian + apt python) for this sample app would be to not depend on numpy and scikit (maybe we could use a simpler way than LOF for this, I'm not sure of the value of something that complex for a sample app).
> 
> But we can postpone that discussion and just fix the build for now

agree

PR #696, opened 2024-05-16
Updated discovery.proto. Removed optional label.

auyidi1:

**If applicable**:


kate-goldenring (changes requested):
@auyidi1 thank you bringing this up as some libraries no longer allow optional fields unless a flag is set. 

This change should also come with a rebuild of the generated code (`cd discovery-utils && make build`)  and dependent changes to the discovery handler implementations. I know this is a delayed review. Let me know if you would like me to push those changes into here or whether you are happy to take this on.

PR #697, opened 2024-06-07
Remove stalebot

diconico07:
As discussed in monthly meeting, removing stale issue marking as it is more annoying than useful currently


kate-goldenring (approved):
🎉

PR #698, opened 2024-06-27
fix: add proto build flag to allow option proto fields

kate-goldenring:
Currently builds fail due to using deprecated use of optional fields. This adds in a build flag to ignore the error. We should remove the proto `optional` specifier and update the crate

ref https://github.com/project-akri/akri/pull/696


kate-goldenring:
@adithyaj @diconico07 this should be ready to review

diconico07 (approved):

PR #703, opened 2024-10-02
chore(rust): update to use rust 1.79 and apply clippy cleanups

kate-goldenring:
- Updates to Rust 1.79
- Updates Cargo.lock files to use workspace level rust version, project version etc.
- Updates version.sh to simplify it thanks to the above
- Applies changes required by latest clippy and fmt 
- Allows a few clippy ignores


Cargo.toml
```diff
@@ -21,3 +21,11 @@ members = [
     "discovery-handler-modules/udev-discovery-handler",
 ]
 resolver = "2"
+
+[workspace.package]
```
diconico07, line 25: nit: We may want to also set the authors key here for consistency ?

diconico07 (approved):
LGTM, just a small possible improvement regarding the migration of `Cargo.toml` fields at workspace level.

Cargo.toml
kate-goldenring: I wasn't sure what to put and punted but i think this is worth unifying. I set the global author to be "The Akri Team"

PR #711, opened 2024-10-29
Fix broker properties not getting updated on Configuration change

### LINKED ISSUE #705: Akri Agent is not recreating instances upon configuration updates
**Describe the bug**
The latest changes to the agent exhibit different behavior than the previous (released) agent. If i install the debugEcho DH and discover two devices, two instances are created. Then if i update the installation to add `brokerProperties`, the instances should be recreated or at least updated to have the `brokerProperty` added now, but the new configuration state is not resolved.

**To Reproduce**
Install Akri dev with debug Echo DH and config:
```sh
helm install akri akri-helm-charts/akri-dev --set agent.allowDebugEcho=true     --set debugEcho.discovery.enabled=true     --set debugEcho.configuration.enabled=true     --set debugEcho.configuration.brokerPod.image.repository=nginx     --set debugEcho.configuration.brokerPod.image.tag=stable-alpine     --set debugEcho.configuration.shared=false
```
Result:
```console
$ k get po,akrii,akric
NAME                                              READY   STATUS    RESTARTS   AGE
pod/akri-agent-daemonset-f2kmw                    1/1     Running   0          15m
pod/akri-controller-deployment-6c7c86ffdd-vkp48   1/1     Running   0          15m
pod/akri-debug-echo-489660-pod                    1/1     Running   0          15m
pod/akri-debug-echo-93a7b0-pod                    1/1     Running   0          15m
pod/akri-debug-echo-discovery-daemonset-5qccr     1/1     Running   0          15m
pod/akri-webhook-configuration-8447999b55-2587b   1/1     Running   0          15m

NAME                                      CONFIG            SHARED   NODES                               AGE
instance.akri.sh/akri-debug-echo-489660   akri-debug-echo   false    ["kagold-thinkpad-x1-carbon-6th"]   15m
instance.akri.sh/akri-debug-echo-93a7b0   akri-debug-echo   false    ["kagold-thinkpad-x1-carbon-6th"]   15m

NAME                                    CAPACITY   AGE
configuration.akri.sh/akri-debug-echo   2          15m
```
Then upgrade installation with brokerProperty in config:
```sh
 helm upgrade --install akri akri-helm-charts/akri-dev --set agent.allowDebugEcho=true     --set debugEcho.discovery.enabled=true     --set debugEcho.configuration.enabled=true     --set debugEcho.configuration.brokerPod.image.repository=nginx     --set debugEcho.configuration.brokerPod.image.tag=stable-alpine     --set debugEcho.configuration.shared=false --set debugEcho.configuration.brokerProperties.random=value
```
The instances are not updated and broker pods not restarted:

```console
$ k get akrii -o yaml | grep random
# no output
```

diconico07:
Move the moment we move to `cdi::Kind` abstraction so that updates to broker properties can be reflected here.

On Configuration reconciliation if request already exists, update broker properties

Fix #705

**If applicable**:


agent/src/discovery_handler_manager/discovery_handler_registry.rs
```diff
@@ -169,35 +171,60 @@ pub trait DiscoveryHandlerRegistry: Sync + Send {
...
+        }
+    }
 }
 
 impl DHRequestImpl {
-    fn device_to_instance(&self, dev: &DiscoveredDevice) -> Instance {
+    fn device_to_instance(
+        &self,
+        dev: &DiscoveredDevice,
+        extra_device_properties: &HashMap<String, String>,
```
kate-goldenring, line 216: Why are you passing in properties instead of reading them from `self.extra_device_properties`?

agent/src/discovery_handler_manager/discovery_handler_registry.rs
```diff
@@ -169,35 +171,60 @@ pub trait DiscoveryHandlerRegistry: Sync + Send {
...
+    extra_device_properties: RwLock<HashMap<String, String>>,
     kube_client: Arc<dyn DiscoveryManagerKubeInterface>,
     termination_notifier: Arc<Notify>,
 }
 
+#[async_trait]
 impl DiscoveryHandlerRequest for DHRequestImpl {
-    fn get_instances(&self) -> Result<Vec<Instance>, DiscoveryError> {
+    async fn get_instances(&self) -> Result<Vec<Instance>, DiscoveryError> {
+        let properties = self.extra_device_properties.read().await;
```
kate-goldenring, line 187: Same point here: could this skipped and instead let the called function (`device_to_instance`) access the properties? Or is the concern wanting to avoid too many reads? It is RWLock, so reads should be cheaper. I could also see how your strategy is trying to avoid a racecase where some containers have properties that were written after the initial call

kate-goldenring (approved):
LGTM -- just one clarifying question

agent/src/discovery_handler_manager/discovery_handler_registry.rs
diconico07: acquiring the lock is an async operation, you cannot do this from within the `map` call (same reason for not doing it in `device_to_instance`). 
But in the end I find it better anyway, as it means we have a consistent set of properties for the entire run

kate-goldenring: `device_to_instance` could be made async but i am aligned on wanting the consistent set of properties

PR #717, opened 2024-11-05
Add Helm pre-delete hook to remove configurations before Akri chart c…

kate-goldenring:
ref https://github.com/project-akri/akri/issues/706


deployment/helm/templates/rbac.yaml
```diff
@@ -90,4 +90,39 @@ subjects:
...
+    app.kubernetes.io/component: helm-hook
+rules:
+- apiGroups: ["akri.sh"]
+  resources: ["configurations"]
+  verbs: ["deletecollection"]
+---
+apiVersion: 'rbac.authorization.k8s.io/v1'
+kind: 'ClusterRoleBinding'
+metadata:
+  name: "akri-helm-hook-biding"
```
scb01, line 116: nit: typo `biding`

deployment/helm/templates/pre-delete-job.yaml
```diff
@@ -0,0 +1,35 @@
...
+  template:
+    metadata:
+      name: "{{ .Release.Name }}"
+      labels:
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+        app.kubernetes.io/component: helm-hook
+    spec:
+      restartPolicy: Never
```
scb01, line 27: missing `spec.serviceAccountName`?

deployment/helm/templates/pre-delete-job.yaml
```diff
@@ -0,0 +1,35 @@
...
+      containers:
+      - name: delete-configurations
+        image: curlimages/curl:latest
+        command: ["/bin/sh", "-c"]
+        args: [
+          "curl -ik -X DELETE -H \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\" https://kubernetes.default.svc.cluster.local/apis/akri.sh/v0/namespaces/default/configurations"
```
scb01, line 33: Question: will this `https://kubernetes.default.svc.cluster.local/apis/akri.sh/v0/namespaces/default/configurations` work even if _akri_ is installed in a namespace other than _default_?

deployment/helm/templates/rbac.yaml
```diff
@@ -90,4 +90,39 @@ subjects:
   - kind: 'ServiceAccount'
     name: 'akri-agent-sa'
     namespace: {{ .Release.Namespace }}
+
+{{- if .Values.cleanupHook.enabled }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: akri-helm-hook-sa
```
scb01, line 98: nit: naming, here and in the clusterRole/clusterRoleBinding - perhaps better to call this `akri-helm-cleanup-hook-sa`

deployment/helm/templates/pre-delete-job.yaml
kate-goldenring: Thank you!

deployment/helm/templates/pre-delete-job.yaml
kate-goldenring: this was one of the tests i wanted to run. I confirmed your concern. We need to template the namespace in. I updated it and am able to confirm it cleans up namespace installations

diconico07: The Configurations can also be scattered over multiple namespaces. 
We should do the equivalent of `kubectl delete -A --all --wait akric` (or maybe just use kubectl here rather than crafting the curl request)

kate-goldenring: The fact that Akri handles configs outside it's namespace is weird behavior to me. If i install Akri in the `foo` ns and then apply a config to the `bar` ns, the config is seen by the `foo` agent and instances are created and broker pods deployed (both in the `bar` namespace):
```sh
foo           pod/akri-agent-daemonset-2j2tc                                 1/1     Running     0          2m46s
foo           pod/akri-agent-daemonset-fk8pm                                 1/1     Running     0          2m46s
foo           pod/akri-agent-daemonset-89mhp                                 1/1     Running     0          2m46s
foo           pod/akri-controller-deployment-695fcd4c4f-mf6js                1/1     Running     0          2m46s
foo           pod/akri-webhook-configuration-6b6947dbc-qjwpj                 1/1     Running     0          2m46s
bar           pod/k3d-test-cluster-server-0-akri-debug-echo-foo-a19705-pod   1/1     Running     0          110s
bar           pod/k3d-test-cluster-server-0-akri-debug-echo-foo-8120fe-pod   1/1     Running     0          110s
bar           pod/k3d-test-cluster-agent-1-akri-debug-echo-foo-a19705-pod    1/1     Running     0          110s
bar           pod/k3d-test-cluster-agent-1-akri-debug-echo-foo-8120fe-pod    1/1     Running     0          110s
bar           pod/k3d-test-cluster-agent-0-akri-debug-echo-foo-8120fe-pod    1/1     Running     0          110s
bar           pod/k3d-test-cluster-agent-0-akri-debug-echo-foo-a19705-pod    1/1     Running     0          110s

NAMESPACE   NAME                                          CONFIG                SHARED   NODES                                                                                 AGE
bar         instance.akri.sh/akri-debug-echo-foo-a19705   akri-debug-echo-foo   true     ["k3d-test-cluster-agent-0","k3d-test-cluster-server-0","k3d-test-cluster-agent-1"]   110s
bar         instance.akri.sh/akri-debug-echo-foo-8120fe   akri-debug-echo-foo   true     ["k3d-test-cluster-agent-0","k3d-test-cluster-server-0","k3d-test-cluster-agent-1"]   110s

NAMESPACE   NAME                                        CAPACITY   AGE
bar         configuration.akri.sh/akri-debug-echo-foo   5          110s
```
I would expect the agent to only watch resources in it's ns? That is clearly not the behavior. Is that an issue we want to look at? If not, then yes, it makes sense to delete configs across all namespaces

diconico07: The plan for splitting the Configuration resource is to have the discovery part to be cluster wide, and only the broker part be namespaced, so it will probably be clearer when done.

But, as is, I expect akri to watch for resources across all namespaces, as I usually want the broker pods to be in app specific namespaces that are not the same as the akri namespace (when doing tests, I usually install akri in its own namespace and creates my Configurations in the default one)

kate-goldenring: I may not have time to adjust this to be cluster wide before we release tomorrow. I will try to get to it, but if not, would we be interested in merging this as is?

kate-goldenring: Updated to use `kubectl`

deployment/helm/templates/pre-delete-job.yaml
```diff
@@ -0,0 +1,36 @@
...
+        app.kubernetes.io/component: helm-hook
+    spec:
+      restartPolicy: Never
+      serviceAccountName: akri-helm-cleanup-hook-sa
+      containers:
+      - name: delete-configurations
+        image: bitnami/kubectl:latest
+        command: ["/bin/sh", "-c"]
+        args: [
+          "kubectl delete configurations.akri.sh --all --all-namespaces"
```
diconico07, line 34: I think we should get it to wait for the deletion to really occur, otherwise we may end up marking it for deletion and helm then deleting the agents daemonset before the delete being processed.

diconico07 (commented):
Apart from the possible race between helm and the agent that may happen here, LGTM.

deployment/helm/templates/pre-delete-job.yaml
kate-goldenring: Good call. Just updated

diconico07 (approved):

kate-goldenring:
`version.sh` runs `cargo update` which caused a dep of the onvif discovery handler to bump in a way the updated the ONVIF XML output should be expected by the `test_create_onvif_discovery_message` test. Pushed an update to the test to fix it

Issue #726, opened 2024-11-26
Enable configuring device mount permissions in udev Discovery Handler


kate-goldenring:
USB devices are mounted by Akri using the [device plugin interface](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/). [Here](https://github.com/project-akri/akri/blob/edc2803f5591a00bd2ff9f6fc4a1011dd1b84a86/agent/src/plugin_manager/device_plugin_instance_controller.rs#L464) is where that is done. By default, read/write/modify permissions are [set by the udev discovery handler](https://github.com/project-akri/akri/blob/edc2803f5591a00bd2ff9f6fc4a1011dd1b84a86/discovery-handlers/udev/src/discovery_handler.rs#L108). These permissions should be configurable for the discovery handler for a Configuration. We'd need to add a new field to the `discoveryDetails` of the discovery handler here: https://github.com/kate-goldenring/akri/blob/733fae50c1dae6ec8351c92a5e6ba9695ba1f7e1/discovery-handlers/udev/src/discovery_handler.rs#L28-L29

PR #737, opened 2025-01-17
added configuring of device mount permissions in udev discovery handler #726

bindsi:
This PR adds configuration options for the device mount permissions in the udev discovery handler.

Corresponding doc PR https://github.com/project-akri/akri-docs/pull/116

**If applicable**:
- [x] this PR has an associated PR with documentation in [akri-docs](https://github.com/project-akri/akri-docs)


bindsi:
@yujinkim-msft can you add maybe @kate-goldenring as review as she requested/suggested configuring of device mount permissions...I cannot do it so far...maybe we can discuss that I could become "maintainer" too.

- would be great to merge this quickly as I also resolved some of the errors in the validation workflows caused by an old rust toolchain version, unfortunately using the latest 1.84 doesn´t work as it would require some code changes due to breaking changes and yaserde cannot be used as it...so was updating to 1.81 which is the min-version required by some of our dependencies.

kate-goldenring (commented):
Just one small improvement. Please also add default overriding configuration to the Helm lint files for [pods](https://github.com/project-akri/akri/blob/8e5b3ab08f6cb109387c03a05e577cd701d09db9/test/helm-lint-values.yaml#L21) and [jobs](https://github.com/project-akri/akri/blob/8e5b3ab08f6cb109387c03a05e577cd701d09db9/test/helm-lint-values-jobs.yaml#L22). Thanks for adding this!

discovery-handlers/udev/src/discovery_handler.rs
```diff
@@ -30,6 +30,14 @@ pub struct UdevDiscoveryDetails {
 
     #[serde(default)]
     pub group_recursive: bool,
+
+    #[serde(default = "default_permissions")]
+    pub permissions: String,
+}
+
```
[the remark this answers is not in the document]
bindsi: Done

discovery-handlers/udev/src/discovery_handler.rs
```diff
@@ -187,9 +217,23 @@ mod tests {
...
+    #[test]
+    #[should_panic]
+    fn test_deserialize_discovery_details_permissions_invalid() {
+        let yaml = r#"
+          udevRules:
+          - 'KERNEL=="video[0-9]*"'
+          permissions: xyz
+        "#;
+        let config: UdevDiscoveryDetails = deserialize_discovery_details(yaml).unwrap();
+        assert_eq!(&config.permissions, "xyz")
```
kate-goldenring, line 237: This assertion should never be reached since the error will be thrown during deserialization. A better practice than `#[should_panic]` would be to test for the specific error:
```suggestion
        match deserialize_discovery_details::<UdevDiscoveryDetails>(yaml) {
            Ok(_) => panic!("Expected error parsing invalid permissions"),
            Err(e) => assert!(e.to_string().contains("a valid permission combination")),
        }
```

kate-goldenring (approved):
One final nit suggestion on the test but LGTM

discovery-handlers/udev/src/discovery_handler.rs
bindsi: of course...to much .Net recently ;-)

yujinkim-msft (approved):

Issue #732, opened 2025-01-03
nodeSelector not working for webhookConfiguration


martinspaniol:
**Describe the bug**
I install akri with this command:
```shell
helm install akri akri-helm-charts/akri \
  --set useLatestContainers=true \
  --set-string agent.nodeSelectors.worker="true"  \
  --set-string webhookConfiguration.nodeSelectors.worker="true"
```
> Note: I need to use --set-string here to make "true" (a boolean value) work as a string

**Result**
```shell
Error: INSTALLATION FAILED: YAML parse error on akri/templates/webhook-configuration.yaml: error converting YAML to JSON: yaml: line 121: did not find expected key
```

**Possible reason**
I looked into the [webhook-configuration.yaml](https://github.com/project-akri/akri/blob/327442f0562dfcd345b6a692d0d7cba451b580aa/deployment/helm/templates/webhook-configuration.yaml). The `nodeSelector` is in [line 110](https://github.com/project-akri/akri/blob/327442f0562dfcd345b6a692d0d7cba451b580aa/deployment/helm/templates/webhook-configuration.yaml#L110): 
```yaml
          nodeSelector:
            {{- if .Values.webhookConfiguration.nodeSelectors }}
              {{- toYaml .Values.webhookConfiguration.nodeSelectors | nindent 8 }}
            {{- end }}
            # ...
            {{- end }}
```
The value has an indent of 8 and I think that's the reason. Shouldn't it be 12?

I also checked with the `template` command. Executing this:
```shell
helm template akri akri-helm-charts/akri   \
  --set useLatestContainers=true \
  --set-string agent.nodeSelectors.worker="true" \
  --set-string webhookConfiguration.nodeSelectors.worker="true" \
  --debug
```

results in:
```yaml
# Source: akri/templates/webhook-configuration.yaml

apiVersion: v1
kind: List
metadata:
  name: akri-webhook-configuration
  labels:
    helm.sh/chart: akri-0.13.8
    app.kubernetes.io/part-of: akri
    app.kubernetes.io/instance: akri
    app.kubernetes.io/version: "0.13.8"
    app.kubernetes.io/managed-by: Helm
items:
  - apiVersion: v1
    kind: ServiceAccount
    metadata:
      name: akri-webhook-configuration
      namespace: default
      labels:
        helm.sh/chart: akri-0.13.8
        app.kubernetes.io/part-of: akri
        app.kubernetes.io/instance: akri
        app.kubernetes.io/version: "0.13.8"
[...]

PR #749, opened 2025-04-01
Fix: corrected webhook nodeSelector indentation in Helm chart

revanthkoganti:
This PR fixes an indentation issue in the nodeSelector field within the webhookConfiguration section of the Helm chart.
Improper indentation in this field was causing Helm template rendering to fail or resulting in misconfigured deployments.

closes issue #732


bindsi (approved):
thanks

Issue #747, opened 2025-03-14
custom-configuration.yaml uses incorrect reference for imagePullPolicy for brokerPods


aaronezack:
**Description**
```custom-configuration.yaml``` does not correctly reference the imagePullPolicy that is set in the ```values.yaml```.
[custom-configuration.yaml](https://github.com/project-akri/akri/blob/main/deployment/helm/templates/custom-configuration.yaml)

**Current Code**
Line 19: ```imagePullPolicy: {{ . }}```

**Expected behavior**
Require this line to correctly reference the pullPolicy set in a custom configuration ```values.yaml```. It should be ```imagePullPolicy: {{ .Values.custom.configuration.brokerPod.image.pullPolicy }}```

PR #752, opened 2025-04-04
fix imagePullPolicy reference for brokerPods

revanthkoganti:

Fixes the imagePullPolicy in custom-configuration.yaml to correctly use the value from values.yaml. This makes sure the broker pod uses the pull policy set by the user.

#747


bindsi (approved):
thanks

PR #760, opened 2025-05-07
feat: Cleanup metadata

ananos:
In multi-node setups, instances are not deleted correctly. The Agent fails to submit the updated spec due to metadata fields being present. Also, finalizers are not cleared properly, so Instances remain stale even if the devices are not discovered.

**If applicable**:

### DIFF
--- a/agent/src/util/discovery_configuration_controller.rs
+++ b/agent/src/util/discovery_configuration_controller.rs
@@ -18,6 +18,7 @@ use crate::discovery_handler_manager::{
     discovery_handler_registry::DiscoveryHandlerRegistry, DiscoveryError,
 };
 
+use kube::api::ObjectMeta;
 use kube::{Resource, ResourceExt};
 use kube_runtime::{
     controller::Action,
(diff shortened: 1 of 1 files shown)


agent/src/util/discovery_configuration_controller.rs
```diff
@@ -204,7 +204,16 @@ async fn delete_instance(
             return Ok(());
         }
         let mut new_instance = instance.clone();
+        // FIXME: temporary workaround for multinode setups.
```
kate-goldenring, line 207:
```suggestion
        let mut new_instance = Instance {
            metadata: ObjectMeta::default(),
            spec: instance.spec.clone()
        };
        new_instance.spec.nodes = vec![];
```
Only updating the spec is another option here.

kate-goldenring (approved):
@ananos thank you for identifying this issue and fixing it. If the latest changes work/fix your issue then LGTM!

ananos:
they do! thanks so much for the detailed walkthrough!

gauravgahlot (approved):

PR #761, opened 2025-05-12
fix: Keep name & namespace fields

ananos:
In commit f57587e we cleanup the metadata structure to avoid issues related to finalizers and managed fields. However, it seems that the k8s api does not like empty name / namespace entries. This patch adds those two entries from the original instance structure.

Hot fix for PR #760 
**If applicable**:
- [X] added code adheres to standard Rust formatting (`cargo fmt`)
- [X] all commits pass the [DCO bot check](https://probot.github.io/apps/dco/) by being signed off -- see the failing DCO check for instructions on how to retroactively sign commits


kate-goldenring (approved):
This is a good patch to the existing bug around name and namespace needing to be set so we should get this in. However, we should move to using Patch::apply for object updates as discussed in this thread: https://kubernetes.slack.com/archives/C01D9L7QE8Z/p1747150005750109?thread_ts=1746553367.459469&cid=C01D9L7QE8Z

gauravgahlot (approved):

yujinkim-msft:
@ananos thank you for the PR! could you please add the version update so we can merge this PR?

PR #800, opened 2026-04-24
Include Akri CRDs in kubectl get all

AkashKumar7902:
**What this PR does / why we need it**:

Adds the `all` category to the Configuration and Instance CRDs so `kubectl get all` includes Akri resources.

Closes #268.

**Special notes for your reviewer**:

Validated with:
- `helm lint deployment/helm`
- `helm template akri deployment/helm --include-crds`
- `git diff --check`

**If applicable**:
- [ ] this PR has an associated PR with documentation in [akri-docs](https://github.com/project-akri/akri-docs)
- [ ] this PR contains unit tests
- [ ] added code adheres to standard Rust formatting (`cargo fmt`)
- [ ] code builds properly (`cargo build`)
- [ ] code is free of common mistakes (`cargo clippy`)
- [x] all Akri tests succeed (`cargo test`)
- [ ] inline documentation builds (`cargo doc`)
- [x] all commits pass the [DCO bot check](https://probot.github.io/apps/dco/) by being signed off -- see the failing DCO check for instructions on how to retroactively sign commits

PR #815, opened 2026-05-27
(chore): update kube to v1.0.0

gauravgahlot:

- Updates `kube` from `v0.91.0` to `v1.0.0`
- Updates `k8s-openapi` from `v0.22.0` to `v0.25.0` 
- Updates webhook test fixture as per new pod spec

**If applicable**:


controller/src/util/pod_watcher.rs
```diff
@@ -201,20 +201,21 @@ impl BrokerPodWatcher {
...
                 info!("handle_pod - Deleted: {:?}", &pod.metadata.name);
                 self.handle_deleted_pod_if_needed(&pod, kube_interface)
                     .await?;
             }
-            Event::Restarted(pods) => {
+            Event::Init => {
                 if *first_event {
-                    info!("handle_pod - pod watcher [re]started. Pods are : {pods:?}");
+                    info!("handle_pod - pod watcher [re]started");
                 } else {
```
bindsi, line 212: Minor observability note: the previous log included the full pod list (`Pods are : {pods:?}`). With the new `Init` / `InitApply` split that exact log can't be reproduced as-is, and it was arguably noisy. Flagging only so the loss is a deliberate choice rather than accidental — no action needed unless operators relied on it for diagnostics.

controller/src/util/instance_action.rs
```diff
@@ -150,6 +150,7 @@ async fn handle_instance(
                 ));
             }
         }
+        _ => {}
     }
```
bindsi, line 154: Suggestion: replace the `_ => {}` catch-all with an explicit list of the variants we intend to ignore:

```rust
Event::InitApply(_) | Event::InitDone => {}
```

`kube::runtime::watcher::Event` is `#[non_exhaustive]`, so the catch-all will silently swallow any new variants added in future `kube` releases. The current behavior is preserved (the old `Restarted(vec)` bundled the same items and we already discarded them), but being explicit keeps the compiler honest on the next bump and documents intent. Same applies in `node_watcher.rs` and `pod_watcher.rs`.

controller/src/util/node_watcher.rs
```diff
@@ -134,6 +134,7 @@ impl NodeWatcher {
                     ));
                 }
             }
+            _ => {}
```
bindsi: Same point as in `instance_action.rs` — prefer `Event::InitApply(_) | Event::InitDone => {}` over `_ => {}` so future `kube::runtime::watcher::Event` variants surface as a compile error rather than being silently dropped.

shared/Cargo.toml
```diff
@@ -12,8 +12,8 @@ rust-version.workspace = true
 anyhow = "1.0.38"
 async-trait = "0.1.0"
 either = "1.13"
-k8s-openapi = { version = "0.22.0", default-features = false, features = ["schemars", "v1_25"] }
-kube = { version = "0.91.0",  features = ["derive" ] }
+k8s-openapi = { version = "0.25.0", default-features = false, features = ["schemars", "v1_33"] }
```
bindsi, line 15: Worth verifying: the `k8s-openapi` feature bumps to `v1_33` from `v1_25` across `shared`, `agent`, `controller`, and the webhook. That's a large jump in the embedded Kubernetes type definitions, and the `Configuration` / `Instance` CRDs in `deployment/helm/crds/` are unchanged in this PR.

Could you run `cargo run --bin gen_crds` (or whatever the project's regeneration entrypoint is — `shared/src/gen_crds.rs` looked like it) and either commit the resulting diff or confirm there is none? `ResourceRequirements.claims` is one example of a field that appeared in this window; there may be others affecting the generated schema. Catching schema drift here avoids a silent skew between what the binaries serialize and what users have installed.

controller/src/util/pod_watcher.rs
```diff
@@ -201,20 +201,21 @@ impl BrokerPodWatcher {
...
                     return Err(anyhow::anyhow!(
                         "Pod watcher restarted - throwing error to restart controller"
                     ));
                 }
             }
+            _ => {}
         };
```
bindsi, line 219: Same point as in `instance_action.rs` and `node_watcher.rs` — prefer an explicit `Event::InitApply(_) | Event::InitDone => {}` over the `_` catch-all so new `watcher::Event` variants in future `kube` releases produce a compile error instead of being silently ignored.

webhooks/validating/configuration/src/main.rs
```diff
@@ -550,11 +555,7 @@ mod tests {
                 "name": "name"
             }
         ],
-        "resources": {
-            "limits": {
-                "{{PLACEHOLDER}}": "1"
-            }
-        },
+        "akriUnknownPodField": "this_field_does_not_exist",
```
bindsi: The strategy here is sound — the `check` function walks the original JSON and requires every value to survive the typed round-trip, so a field unknown to `PodSpec` gets dropped on deserialization and the mismatch trips rejection. The inline comment captures the rationale well.

Two small nits:

1. The marker key `akriUnknownPodField` reads like a real Akri concept. Something more self-describing such as `xUnknownSyntheticField` or `__akriWebhookTestUnknownField` would reduce the risk of future confusion.
2. The contract being tested ("PodSpec drops unknown fields on deserialize") is now load-bearing for this test. If upstream `PodSpec` ever grew a field with this name, or `deny_unknown_fields` semantics changed, the test would start passing for the wrong reason. A tiny direct assertion proving "unknown field round-trips to absent" alongside this fixture would harden it against drift. Non-blocking.

bindsi (commented):
Dependency upgrade looks mechanically clean. Inline comments cover:

- Three `_ => {}` catch-alls in the watcher event matches — suggest making the `InitApply` / `InitDone` arms explicit since `watcher::Event` is `#[non_exhaustive]`.
- `k8s-openapi` feature jump from `v1_25` to `v1_33` — worth regenerating the CRDs to confirm no schema drift slipped through.
- Webhook fixture rewrite — strategy is correct; small naming + hardening suggestions.
- One observability note on the dropped pod list from the `[re]started` log.

Nit on the PR description: it says "Updates `kube` from `v0.92.0` to `v1.0.0`", but `main` is on `0.91.0` and all the `Cargo.toml` diffs are `0.91.0` → `1.0.0`. Worth fixing before merge so release notes stay accurate.

None of these are blocking — happy to approve after the CRD regeneration check.

bindsi (approved):
Thanks for taking this on, @gauravgahlot — the upgrade is mechanically clean and the watcher Event migration plus the webhook fixture rewrite are both well-handled. Approving.

A few non-blocking notes from the inline comments worth a quick look before merge:

- The three `_ => {}` arms in the watcher matches preserve current behavior, but listing `Event::InitApply(_) | Event::InitDone` explicitly would keep the compiler honest the next time `kube` adds a variant (the enum is `#[non_exhaustive]`).
- Quick sanity check on the `k8s-openapi` feature jump from `v1_25` to `v1_33`: would you mind running `cargo run --bin gen_crds` and confirming the CRDs in `deployment/helm/crds/` don't need to be regenerated (or committing the diff if they do)?
- Minor: the PR description says "from `v0.92.0`" but `main` is on `0.91.0` — worth fixing for the release notes.

Happy to leave the catch-all and fixture-naming nits to your discretion. Nice work!

controller/src/util/instance_action.rs
gauravgahlot: You are right, it's always better to be explicit.

webhooks/validating/configuration/src/main.rs
gauravgahlot: Fair point. Added a test.

shared/Cargo.toml
gauravgahlot: Diffed against the committed `Instance` CRD — no schema content differences, only YAML cosmetic style. The `Instance` schema doesn't depend on k8s-openapi (only primitive types in `InstanceSpec`), and the `Configuration` CRD uses `x-kubernetes-preserve-unknown-fields: true` for its PodSpec/JobSpec/ServiceSpec blocks, so neither CRD picks up changes from the v1_25 → v1_33 bump. Leaving the CRDs as-is.

Separately, I noticed the categories: [all] from #800 is encoded only in the YAML, not in the `#[kube(...)]` derive — happy to open a follow-up adding `category = "all" ` to the macro so future regenerations don't drop it.

PR #833, opened 2026-06-26
docs: add developer guide

gauravgahlot:

Adds a developer guide for locally building and testing Akri components.

**If applicable**:

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -110,10 +110,15 @@ One note on the Agent: by default it's "slim" and embeds no discovery handlers, 
 why you deploy handlers as separate DaemonSets (you'll see `<handler>.discovery.enabled`
 below). If you'd rather embed the handlers in the Agent itself, that's the separate
 *agent-full* build — run `AGENT_FEATURES="agent-full onvif-feat opcua-feat udev-feat" make
-akri-agent-full`, which produces its own `PREFIX/agent-full:LABEL` image (so you'd point
-`agent.image.repository` at `…/agent-full`). Watch out: plain `make akri-agent` always builds
-the slim Agent regardless of `AGENT_FEATURES` — see `build/akri-containers.mk:9` for the
-single-command alternative.
+akri-agent-full`, which produces its own `PREFIX/agent-full:LABEL` image. To *deploy* that
+image, flip the chart's dedicated toggle rather than overriding the ordinary repository:
+`--set agent.full=true` makes the template select `agent.image.fullRepository` instead of
+`agent.image.repository`, and also gives the DaemonSet the `hostNetwork` that embedded ONVIF
+discovery needs — so point `agent.image.fullRepository` (not `agent.image.repository`) at
+your `…/agent-full` image. Watch out: plain `make akri-agent` always builds the slim Agent
+regardless of `AGENT_FEATURES` — the generic `akri-%` rule in `build/akri-containers.mk` never
+passes the feature flags, so the full image comes only from the dedicated `akri-agent-full`
+target.
 
 ## A cluster to run on
 
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -1,0 +1,409 @@
+# Developer Guide: Building and Testing Akri Locally
+
+So you've made a change to Akri and you want to see it actually work. This guide walks you
+from a fresh checkout to watching Akri discover a device and stand up a broker for it on a
+real cluster — all on your own machine.
+
+The thing to understand up front is that testing Akri is a layered activity, and most of
+the time you don't need the whole stack. A change to discovery logic or a controller
+reconcile path is usually proven fastest with `cargo` alone. You only reach for container
+images and a cluster when you need to watch the Agent, Controller, and Discovery Handlers
+behave together at runtime. The guide follows that same order — cheapest and fastest first,
+heaviest last — so you can stop at whatever layer answers your question.
+
+```mermaid
+flowchart LR
+    A(Edit code) --> B(cargo build and test)
+    B -->|logic changed| A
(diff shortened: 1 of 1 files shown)


bindsi (approved):
That´s nice...really good

bindsi (commented):
Thanks for writing this up — it's a genuinely good guide. The layered "cheapest loop first" structure, the K3s-vs-kind `/dev` caveat, and the `[patch.crates-io]` git-vs-path container-build trap are all accurate and useful. I verified the concrete claims against the codebase; most are correct, but two are wrong and would cost a reader real time.

## 🔴 Wrong Rust version (`docs/developer-guide.md`, "What you'll need")

> the workspace's `rust-version` in the root `Cargo.toml` (currently **1.88**) … `rustup default 1.88`

The root `Cargo.toml` actually sets `rust-version = "1.81"`. Since the whole point of this sentence is to point readers at the authoritative value, the wrong number (and the wrong `rustup default 1.88` command) is the most damaging kind of error here. Suggest changing `1.88` → `1.81`, or better, avoid a number that will drift: *"check the `rust-version` field in the root `Cargo.toml` and `rustup default` to that or newer."*

## 🟠 agent-full deployment points at the wrong Helm key ("Building the component images")

> produces its own `PREFIX/agent-full:LABEL` image (so you'd point `agent.image.repository` at `…/agent-full`).

The chart doesn't select the full image that way. `deployment/helm/templates/agent.yaml` uses:

```gotemplate
{{- $repository := ternary .Values.agent.image.fullRepository .Values.agent.image.repository .Values.agent.full -}}
```

There's a dedicated toggle `agent.full` (default `false`) and a separate key `agent.image.fullRepository` (`values.yaml`). The idiomatic invocation is `--set agent.full=true --set agent.image.fullRepository=akri/agent-full`. Overriding `agent.image.repository` happens to work only while `agent.full=false`, and bypasses the mechanism the chart provides — confusing for anyone who later finds the `full`/`fullRepository` keys.

## 🟡 "Always slim" footnote cites a contradictory makefile line

> plain `make akri-agent` always builds the slim Agent regardless of `AGENT_FEATURES` — see `build/akri-containers.mk:9`

The claim is correct (the generic `akri-%` rule never references `AGENT_FEATURES`), but the cited line 9 documents `… AGENT_FEATURES="…" make akri-agent` as the way to build a *full* agent — i.e. the opposite of what the footnote says it supports. Either the makefile's header comment is stale or the behavior is subtler. Suggest citing the generic build rule (`akri-%`, ~line 16) instead, and double-checking that contradiction.

## ✅ Verified accurate (no action)

`allowDebugEcho` being a no-op in the slim + standalone-handler setup (confirmed via `templates/agent.yaml` → `agent/src/discovery_handler_manager/embedded_handler.rs`), `webhookConfiguration.enabled: true` default, broker tag defaulting to `latest`, the `-handler`-stripping image-naming rule, the debug-echo `akri-debug-echo` / capacity 2 / `foo0`+`foo1` defaults, and `BUILD_RELEASE_FLAG=` skipping `--release` all match the code.

(Heads-up: the DCO check is currently failing — the commit needs a sign-off before merge.)

bindsi (changes requested):
Requesting changes for the factual issues raised in my earlier review — these are quick fixes:

1. 🔴 **Wrong Rust version** — `docs/developer-guide.md` says `rust-version` is "currently **1.88**" and suggests `rustup default 1.88`, but the root `Cargo.toml` sets `rust-version = "1.81"`. Please correct the number (or reword to point at the field without hard-coding a value that will drift).

2. 🟠 **agent-full Helm key** — the guide says to point `agent.image.repository` at `…/agent-full`, but the chart selects the full image via `ternary .Values.agent.image.fullRepository .Values.agent.image.repository .Values.agent.full`. The idiomatic path is `--set agent.full=true --set agent.image.fullRepository=…`.

3. 🟡 **makefile citation** — the "plain `make akri-agent` always builds slim" footnote cites `build/akri-containers.mk:9`, which actually documents building a *full* agent; please cite the generic `akri-%` rule instead.

Once the Rust version is fixed I'm happy to approve. Thanks again — this is a strong guide. (Also note the DCO check is still failing and needs a sign-off.)

gauravgahlot:
Thanks for the review @bindsi 

**1. Rust version:** root `Cargo.toml` is `rust-version = "1.88"` (set in #777, Dec 2025) on both `main` and this branch; `1.81` isn't in the tree.

  **2. agent-full key:** fixed — now `--set agent.full=true` + `agent.image.fullRepository` (which also enables `hostNetwork` for embedded ONVIF), not `agent.image.repository`.

   **3. makefile citation:** fixed in the guide — the footnote now cites the `akri-%` rule / `akri-agent-full` target instead of line 9. (Line 9's comment is itself stale, but that's a separate makefile fix, not this PR.)


I always sign my commits. DCO was already green.

bindsi (approved):
Re-reviewed against the current base — this looks good now, approving. ✅

Two of my earlier points were addressed cleanly:
- **agent-full deployment** now correctly uses `--set agent.full=true` + `agent.image.fullRepository`, and the added note about `agent.full` also enabling `hostNetwork` for embedded ONVIF discovery matches the chart (`deployment/helm/templates/agent.yaml`). 👍
- **The "always slim" footnote** now cites the generic `akri-%` rule instead of the contradictory line. 👍

And a correction on my part: my earlier comment claimed the **`rust-version` should be 1.81** — that was wrong, I'd checked a stale branch. The root `Cargo.toml` on `main` is `rust-version = "1.88"`, so the guide is correct as written. Apologies for the noise.

Nice guide overall — the layered structure and the patched-`h2` container-build trap are especially useful. (One non-blocking note: the DCO check is still red and needs a sign-off before merge.)

PR #844, opened 2026-08-31
fix(agent): accept the kubelet's re-allocation of a claimed configuration-level device

### LINKED ISSUE #843: Configuration-level resource: `Allocate` rejects the kubelet's re-offer of the same device for a later container of the same Pod ("Unable to claim slot")
## Environment

- Akri agent: `ghcr.io/project-akri/akri/agent:v0.13.8`
- Reproduced identically with an agent built from `main` (`v0.13.26-dev`); the code path is unchanged on `main` as of 2026-08-24 (`edad1b6`, version 0.13.27)
- Kubernetes: K3s `v1.34.6+k3s1`, single node
- Discovery handler: `udev` (one USB device matched, `capacity: 1`)

## Summary

When a Pod contains more than one container (e.g. an init container plus the main container) and each of them requests the same **Configuration-level** resource, the Pod is rejected at admission with:

```text
Pod was rejected: Allocate failed due to rpc error: code = Unknown
desc = Unable to claim slot, which is unexpected
```

The same two-container Pod succeeds when the containers request the **Instance-level** resource instead, and a single-container Pod requesting the Configuration-level resource also succeeds — so the failure is specific to the Configuration-level device plugin's slot-claim path under the kubelet's per-container allocation.

This is not #722: that report shares the error string but its cause was the agent failing to reach the kubelet socket, with a single container and an Instance-level resource. Here the agent is healthy and every single-container Pod allocates fine.

## Why a Pod would do this

Kubernetes computes the effective request of ordinary (sequential) init containers as the maximum, not the sum, and the kubelet's device manager re-offers devices allocated to completed init containers to subsequent containers (`devicesToReuse`). "
One Pod, several sequential stages, each stage a different image, all bound to the same physical device" is therefore a legitimate pattern— for hardware-in-the-loop testing it maps to probe → test → report stages holding one DUT for the whole Pod lifetime.

Akri itself already treats this as a supported shape: `brokerPodSpec` supports `initContainers` with the resource `PLACEHOLDER` (#645, fixed by #649). That path lands on the Instance-level plugin, which tolerates the re-offer; the Configuration-level plugin does not, so the two plugin paths are out of parity.

## Minimal reproduction

Configuration (udev; matches exactly one USB device):

```yaml
apiVersion: akri.sh/v0
kind: Configuration
metadata:
  name: poc-dut-pool
  namespace: akri-system
spec:
  brokerProperties: {}
  capacity: 1
  discoveryHandler:
    name: udev
    discoveryDetails: |
      groupRecursive: false
      udevRules:
        - 'SUBSYSTEM=="usb", ATTR{idVendor}=="05c6", ATTR{idProduct}=="9135", ATTR{serial}=="<serial>"'
```

Discovered instance: `poc-dut-pool-3bc9b6`; node allocatable shows both
`akri.sh/poc-dut-pool: 1` and `akri.sh/poc-dut-pool-3bc9b6: 1`.

**Pod A — single container, Configuration-level resource → Succeeded:**

```yaml
apiVersion: v1
kind: Pod
metadata: {name: ctl-one}
spec:
  restartPolicy: Never
  containers:
    - name: only
      image: busybox:1.36
      command: ["sh","-c","env | grep ^UDEV_"]
      resources:
        requests: {akri.sh/poc-dut-pool: "1"}
        limits: {akri.sh/poc-dut-pool: "1"}
```

**Pod B — init + main, Configuration-level resource → rejected:**

```yaml
apiVersion: v1
kind: Pod
metadata: {name: ctl-two}
spec:
  restartPolicy: Never
  initContainers:
    - name: first
      image: busybox:1.36
      command: ["sh","-c","echo init-claimed"]
      resources:
        requests: {akri.sh/poc-dut-pool: "1"}
        limits: {akri.sh/poc-dut-pool: "1"}
  containers:
    - name: second
      image: busybox:1.36
      command: ["sh","-c","echo main-claimed"]
      resources:
        requests: {akri.sh/poc-dut-pool: "1"}
        limits: {akri.sh/poc-dut-pool: "1"}
```

Result: `Failed / UnexpectedAdmissionError: Pod was rejected: Allocate failed due to rpc error: code = Unknown desc = Unable to claim slot, which is unexpected`

**Pod C — init + main, Instance-level resource (`akri.sh/poc-dut-pool-3bc9b6`) → Succeeded.**

The three Pods are identical apart from the resource name and container count, so the variable is isolated to "multiple containers requesting the same Configuration-level resource".

Two variations rule out capacity as the cause: `capacity: 2` (verified applied on the Instance) still rejects Pod B, and two init containers with a non-claiming main container still reject — the kubelet re-offers the *same* virtual device id in every case.

## Agent log excerpt (TRACE)

```text
kubelet called allocate Request { ... AllocateRequest { container_requests:
  [ContainerAllocateRequest { devices_i_ds: ["poc-dut-pool-0"] }] } ... }
allocate - kubelet called allocate for Configuration poc-dut-pool
...
(second per-container allocate for the same virtual device id fails; Pod is rejected; the instance device flips to Unhealthy)
...
device_plugin_slot_reclaimer: freeing slot: poc-dut-pool-0 (≈26 s later; the instance then returns to Healthy)
```

## Analysis

The kubelet allocates devices per container and deliberately re-offers the same device id (`poc-dut-pool-0`) for the later containers of the Pod. 
The first `Allocate` claims the virtual-device slot (the `deviceUsage` entry `C:<virtual_device_id>:<node>` described in the resource-sharing docs). 
The second `Allocate` for the same virtual device id finds the slot already occupied — by the Pod's own previous claim — and fails with "Unable to claim slot, which is unexpected" instead of treating the re-claim as idempotent.
The Instance-level plugin path tolerates the same sequence.

### Source-level root cause (`main` @ `edad1b6`; identical in v0.13.8)

`agent/src/plugin_manager/device_plugin_instance_controller.rs`:

- The shared `claim_slot()` helper **does** have an idempotency branch — `d if *d == wanted_state => id`, commented "The kubelet asks for the same slot, it knows best"([L198-L199](https://github.com/project-akri/akri/blob/edad1b6ea25664760df04c1955f41e83f1df80e8/agent/src/plugin_manager/device_plugin_instance_controller.rs#L198-L199)).
  The Instance-level plugin's `allocate`([L383-L406](https://github.com/project-akri/akri/blob/edad1b6ea25664760df04c1955f41e83f1df80e8/agent/src/plugin_manager/device_plugin_instance_controller.rs#L383-L406)) reaches it with `DeviceUsage::Node(node)`, which is why the Instance-level path tolerates per-container re-allocation.
- The Configuration-level `allocate` ([L682-L701](https://github.com/project-akri/akri/blob/edad1b6ea25664760df04c1955f41e83f1df80e8/agent/src/plugin_manager/device_plugin_instance_controller.rs#L682-L701)) never reaches that branch: its guard accepts only
  `ConfigurationSlot::DeviceFree` (L682). For the Pod's second container the slot is already `DeviceUsed` — by the same virtual device id on the same node — so the `else` arm returns `"Unable to claim slot"` unconditionally (L701), with no check that the current holder is the same vdev/node (a kubelet re-offer rather than a genuine conflict).
- Nothing in this path depends on how many instances the Configuration matches: a serial-pinned single-device Configuration takes the same code path as a pool, which matches the reproduction above.

## Expected behavior

The Configuration-level `allocate` should treat a re-offer of a virtual device that is already `DeviceUsed` by the same vdev on the same node as idempotent — return the same physical device again — mirroring what `claim_slot()` already does one layer down. Cross-Pod behavior must stay unchanged: a slot used by a different holder is still rejected, and `capacity` accounting is untouched.

## Proposed fix (implemented and validated; PR offered)

We carry a small patch to that one file, applied cleanly to current `main` (`git apply --check` against `edad1b6`):

1. Route the `ConfigurationSlot::DeviceUsed { device, slot_id }` case to `claim_slot(Some(slot_id), DeviceUsage::Configuration { vdev, node })`, i.e. into the existing idempotent path, instead of returning the error.
2. Serialize Configuration-level allocations with a mutex so a repeated request cannot observe a stale `DeviceFree` entry before the first claim is published.
3. Publish the selected slot into the shared slot map immediately after the claim rather than waiting for the Instance watcher to converge.

The bundled unit test (`test_config_plugin_allocate`) allocates the same virtual device twice and asserts both calls succeed against the same slot.

Validated on a cluster (K3s v1.34.6, real USB device, stock discovery handler): discovery unaffected; single-container Pod unchanged; init+main Pod now admits and both containers receive the same device; a second Pod requesting the same resource while the first holds it stays `Pending` (`Insufficient akri.sh/poc-dut-pool`) for the entire hold and is admitted after the holder is deleted; slot reclaim after release unchanged; rollback to the stock agent verified.

If maintainers agree with the direction, we can open a DCO-signed PR against `main`. Open question for review: whether the mutex + immediate publish (items 2–3) are the preferred way to close the timing window, or whether a narrower change to the watcher is preferred.

## Questions

1. Is multi-container (init-container) reuse of a Configuration-level resource intended to work? If yes, this looks like a missing idempotency check in the Configuration device plugin's claim path.
2. If it is an intended limitation, could it be documented, and could the error message distinguish this case from genuine slot conflicts?

## Workarounds found

- Let only the first container claim the resource and give later containers claim-free device access (works, but gives up the per-container device injection for those containers).
- Collapse to a single container per Pod (loses the per-stage image separation that init containers provide).

## Related

- #722 — same error string, different cause (kubelet socket connectivity; single container; Instance-level resource).
- #645 / #649 — `initContainers` with the resource `PLACEHOLDER` in `brokerPodSpec` (the Instance-level path).
- #7 — slot-reclaim polling; the ≈26 s reclaim delay in the log above is that path.

fang80913:
## What

When several containers of one Pod request the same configuration-level resource, the kubelet re-offers the virtual device it already assigned to that Pod. The configuration device plugin only handled the `DeviceFree` state and rejected the re-offer with `Unable to claim slot`, so the Pod died at admission before any container started. Full analysis, minimal reproduction and cluster evidence: #843.

This PR:

1. serializes configuration-level allocations with a per-plugin lock (closes the stale-`DeviceFree` race);
2. routes the `DeviceUsed` state to the instance plugin's claim path, which already accepts an idempotent re-claim of the same slot by the same workload ("kubelet asks for the same slot, it knows best");
3. publishes the chosen mapping before returning to the kubelet.

Cross-Pod exclusion is unchanged: a device claimed by a different workload is still rejected by the existing same-workload check.

## Validation

- `test_config_plugin_allocate` extended to cover the kubelet repeating the allocation for the same assigned virtual device.
- `cargo build -p agent` / `cargo test -p agent` on `main` @ edad1b6 in `rust:1.88` with the CI dependency set: 46 passed, 0 failed.
- Cluster validation (v0.13.8 base, agent image swap on a live K3s cluster with a real USB DUT): single-container control, init + main containers of one Pod receiving the same device, a competing Pod held `Pending` while the resource was claimed and admitted after release, slot reclaim after the holder exited, rollback to the stock agent. Details in #843.

## Open question

The same-workload trust decision currently lives in the instance layer (this PR reuses it). If maintainers prefer the configuration plugin to validate the workload identity itself before delegating, happy to rework in that direction.

Fixes #843


fang80913:
Rebased onto `main` after #842 landed (v0.14.0). The conflict was only in the version bookkeeping files — the agent change itself still applies cleanly and is unchanged.

- Base is now `a2ee683`; `version.txt` bumped 0.14.0 → **0.14.1** via `./version.sh -u -p` (`Cargo.toml`, `Cargo.lock`, `deployment/helm/Chart.yaml` updated with it).
- Re-verified on the rebased tree: `cargo fmt --all -- --check` clean, `cargo test -p agent` 46 passed / 0 failed.

If you would rather this PR not carry a version bump right after the 0.14.0 release, I am happy to drop it — just add the `same version` label and I will push the revert.

fang80913:
Hi maintainers — a small status note on this PR.

It is rebased onto the fresh v0.14.0 release, but the workflow runs are sitting at "Action required", so CI has not run on the current head (`67925a3`). Could someone approve the runs when convenient? The pre-rebase head had the full suite green — 46 unit tests and the 9 cluster matrices — and nothing but the rebase has changed since.

Also happy to drop the version bump if you would rather handle versioning at release time; just say the word and I will push that change.

Context for anyone picking this up: the analysis is in #843. The change routes the Configuration-level plugin's dropped "device already held by this same workload" case into the Instance-level `claim_slot()` branch that already accepts it, so no new policy is introduced and cross-Pod exclusion is unchanged.

